PR #1, opened 2018-08-24
Added beginnings of an nRF52840 hal

jscarrott:
This is just all the old hal code I had for the 52840, edited just to get the GPIO stuff working. I don't mind if you would rather get rid of this and start again.

### DIFF
--- a/nrf52840-hal/Cargo.toml
+++ b/nrf52840-hal/Cargo.toml
@@ -1,0 +1,35 @@
+[package]
+name = "nrf52840-embedded-hal"
+version = "0.1.0"
+authors = ["John Scarrott <johnps@outlook.com>"]
+
+[dependencies]
+bare-metal = "0.2.0"
+cortex-m = "0.5.2"
+cortex-m-rt = "0.5.1"
+nb = "0.1.1"
+nrf52840 = "0.2.0"
+
+[dependencies.cast]
+default-features = false
+version = "0.2.2"
+
+[dependencies.embedded-hal]
+features = ["unproven"]
+version = "0.2.1"
+
+[dependencies.void]
+default-features = false
+version = "1.0.2"
+
+[features]
+default = ["rt"]
+rt = ["nrf52840/rt"]
+[profile.dev]
+debug = true
+lto = false
+
+[profile.release]
+debug = true
+lto = true
+opt-level = "s"
--- a/nrf52840-hal/memory.x
+++ b/nrf52840-hal/memory.x
@@ -1,0 +1,23 @@
+MEMORY
+{
+  /* NOTE K = KiBi = 1024 bytes */
+  /* TODO Adjust these memory regions to match your device memory layout */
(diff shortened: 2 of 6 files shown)


nrf52840-hal/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+name = "nrf52840-embedded-hal"
```
wez: Should we keep this a bit shorter? `nrf52840-hal`?

nrf52840-hal/memory.x
```diff
@@ -0,0 +1,23 @@
+MEMORY
+{
+  /* NOTE K = KiBi = 1024 bytes */
+  /* TODO Adjust these memory regions to match your device memory layout */
+  FLASH : ORIGIN = 0x00000000, LENGTH = 950K
```
wez: These seem a little smaller than I was expecting; are these adjusted for the softdevice?
https://github.com/particle-iot/nrf5_sdk/blob/6177fb3797962aaa091e04fa363bfb93a55eb7ef/modules/nrfx/mdk/nrf52840_xxaa.ld#L8

the memory.x in the hal should probably not factor in the softdevice and that should probably be something that is handled in the "application" crate.

nrf52840-hal/src/gpio.rs
```diff
@@ -0,0 +1,365 @@
+//! General Purpose Input / Output
```
wez: this file looks similar to `nrf52-hal/src/gpio.rs`... is it close enough that you could copy that over this one and maybe tweak the `use nrf52;` lines to `use nrf52840`?

nrf52840-hal/memory.x
jscarrott: No, actually a good catch. When I first tried to get my dev board working I had a real mare of a time. Thinking back I think I reduced the sizes to try and prevent this weird issue I was having, I can't for the life of me remember my rationale. I'll put it back and remove the TODO.

nrf52840-hal/src/gpio.rs
jscarrott: That's because it was originally a copy of an earlier version. I've updated it and made a couple of tweaks for the 52840. Couldn't get the GPIOs on P1 to work with the macro though due to the shared module p0, and I'm not sure of the best way to fix it.

nrf52840-hal/src/gpio.rs
```diff
@@ -0,0 +1,339 @@
...
+    P0_31: (p0_31, 31, Input<Floating>),
+
+]);
+
+// ===========================================================================
+// The macro doesn't seem to work for P1  as it has a different struct but uses
+// the same module. Not sure the best way to deal with P1 only implementing half the pins
+// either
+// ===========================================================================
+//gpio!(P1, p0, P1_Pin [
```
wez: Once https://github.com/jamesmunns/nrf52-hal/issues/8 is a bit further along, I think this will get a bit easier

PR #2, opened 2018-08-25
Add a crate for the adafruit nrf52 board

wez:
At this stage it is convenient to have some board support crates
in this same repo; it makes it easier to test and understand API
changes, and simpler for CI.

I don't know how many variants of the nrf52 boards are out there
and in common use, so there may well be a point at which we decide
to split off the boards directories.

All this crate does is define some nicer names for the pins and
includes a blinky example that I've flashed to my device.

### DIFF
--- a/.gdbinit
+++ b/.gdbinit
@@ -1,18 +1,28 @@
-target remote :3333
+# print demangled symbols by default
+set print asm-demangle on
 
-monitor arm semihosting enable
+# JLink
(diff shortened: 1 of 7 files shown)


wez:
I'm going to land this to streamline a couple of other changes that I'd like to make.  Hopefully this isn't contentious and I'd be happy to discuss this further; we can always change things later.

PR #4, opened 2018-08-27
Added timer, clock and delay hal functions from 52832

jscarrott:


jscarrott:
@wez The CI build see a little flakey for some reason. The first build failed with linker errors, but the second one passed with no changes....

wez:
Looks good, thanks!
re: flakey CI, it's weird; normally I get emails from the CI when the build fails and I didn't get any for your PR, but I did get an email about a **successful** build for https://github.com/nrf-rs/nrf52/commit/8285df84c12a236dbe672e881f2297fbfc82f9ab around the time that you submitted this.  Feels like a travis notification blip?

jscarrott:
No, I don't think it was a reporting blip as the build log looked consistent with what you would expect to see when getting linker errors. I can't find the build log anymore which is a bit annoying, but if it is a real issue I'm sure we will see it again.

PR #5, opened 2018-08-28
add nrf52-hal-common crate

wez:
This commit takes what I believe to be the lowest maintenance approach
to sharing a common implementation between two device crates.

The common crate has a set of features that cause the underlying
svd2rust generated device crate to be imported as `target_device`.
The implementation in that crate then targets `target_device` rather
than directly using either `nrf52` or `nrf523840`.

This avoids littering the module with conditional code and/or macros.

The consuming hal crates re-export the items from the common crate
and augment with any other additional functionality.

One downside to this approach is that the `nrf-hal-common` crate cannot
be successfully compiled when multiple targets are built.  This is
fine in practice but does mean that building in a workspace will
fail when that workspace enables multiple devices.

For that reason this commit adjusts the travis config to build the
crates individually.

Refs: https://github.com/jamesmunns/nrf52-hal/issues/9


nrf52-hal-common/src/spim.rs
```diff
@@ -143,7 +141,7 @@ impl<T> Spim<T> where T: SpimExt {
             //
             // The MAXCNT field is 8 bits wide and accepts the full range of
             // values.
-            unsafe { w.maxcnt().bits(tx_buffer.len() as u8) }
+            unsafe { w.maxcnt().bits(tx_buffer.len() as _) }
```
wez: `maxcnt` is apparently wider on the 52840, so this casts to whatever the size may be

hannobraun:
Sorry, I'm a bit swamped right now. I'll take a look as soon as I take a chance, probably later this week.

hannobraun:
@wez I haven't forgotten this. Still busy though. I expect to have time to review next week. Very sorry to keep this hanging for so long!

nrf52-hal-common/Cargo.toml
```diff
@@ -0,0 +1,40 @@
+[package]
+authors = [
+    "James Munns <james.munns@gmail.com>",
+    "Hanno Braun <hanno@braun-robotics.com"
```
hannobraun: You could add yourself to `authors`.

nrf52-hal-common/Cargo.toml
```diff
@@ -0,0 +1,40 @@
...
+]
+categories = ["embedded", "hardware-support", "no-std"]
+description = "HAL for the nRF52 family of microcontrollers"
```
hannobraun: If I understand correctly, this crate is intended to be a dependency for HAL implementations, not as something to be used by HAL users directly. Unless this is incorrect, this description should be updated to reflect this.

nrf52-hal-common/src/lib.rs
```diff
@@ -0,0 +1,36 @@
+#![no_std]
+
+extern crate cast;
+extern crate cortex_m;
+extern crate embedded_hal as hal;
+extern crate nb;
+extern crate void;
+
+#[cfg(feature = "52832")]
+pub extern crate nrf52 as target_device;
```
hannobraun: Since this is used fairly often, I'm wondering if just `target` would be a better name. Just a though, I don't feel strongly about this.

nrf52-hal-common/src/spim.rs
hannobraun: The comment should be updated then. The "8 bits wide" bit is misleading.

nrf52-hal-common/Cargo.toml
hannobraun: I realized that I worded my criticism too strongly. It makes perfect sense to depend on the common crate whenever you can get away with it (for example in libraries). Still, the description should be updated to state that this is a common crate for the whole nRF52 series, and that there are more specific crates to be found.

wez: Sounds reasonable to me; amended in the latest commit

nrf52-hal-common/src/lib.rs
wez: Yep, sounds good; done!

nrf52-hal-common/src/spim.rs
wez: I've expanded the comment in the latest iteration.  I'll file an issue for handling data longer than 8 bits allow.

hannobraun:
Thanks again, @wez! I don't know if anyone else wants to take a look, but from my perspective, this is good to merge.

hannobraun:
I'm just going to merge this now. This has been open long enough, and I need to make some changes for my downstream project that are currently being complicated by all the open pull requests :)

PR #21, opened 2018-09-14
Unconditionally enable nrf52-hal-common features

hannobraun:
The features that define which target is chosen is always required,
independently of whether the `rt` feature is enabled.


hannobraun:
Thanks for the review, @wez!

PR #23, opened 2018-09-14
Add support for GPIO P1 on nrf52840

wez:
Due to the way that the svd codegen works, the P1 types are emitted
into the p0 module in the peripheral crate, so we need to add a
parameter to the macro to distinguish between the source module and
the exported module; I'm choosing to emit the P1 pins in a p1 module
rather than also jamming them into the p0 module in the hal.

### DIFF
--- a/nrf52-hal-common/src/gpio.rs
+++ b/nrf52-hal-common/src/gpio.rs
@@ -47,7 +47,7 @@ pub struct PushPull;
 
 macro_rules! gpio {
     (
-        $PX:ident, $px:ident, $Pg:ident [
+        $PX:ident, $pxsvd:ident, $px:ident, $Pg:ident [
             $($PXi:ident: ($pxi:ident, $i:expr, $MODE:ty),)+
         ]
     ) => {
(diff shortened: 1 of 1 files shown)


nrf52-hal-common/src/gpio.rs
```diff
@@ -316,3 +316,25 @@ gpio!(P0, p0, P0_Pin [
     P0_30: (p0_30, 30, Input<Floating>),
     P0_31: (p0_31, 31, Input<Floating>),
 ]);
+
+// The p1 types are present in the p0 module generated from the
+// svd, but we want to export them in a p1 module from this crate.
+#[cfg(feature = "52840")]
```
hannobraun: Alternative idea: Instead of `cfg`ing here, make the macro public and use it in `nrf52840-hal` to generate the module there. That way `nrf52-hal-common` only contains common code, and everything's neat and tidy.

Not sure if that would actually be better (I'd have to see it in action to judge). Just thinking out loud.

wez: Nod.  I know your feelings on `cfg` pollution and also desire to minimize it.  At the moment it would require some more work to restructure both this crate and the `nrf52840-hal` crate to support using the macro in that way, so I'm inclined to consider that as a follow-up, especially since both you and @jscarrott are looking at gpio at the moment.

hannobraun: Makes sense. I think my idea is worth looking into, but that doesn't need to happen now (or ever, actually, as long as the `cfg` pollution doesn't get much worse).

PR #25, opened 2018-09-20
Clear counter value when starting timer

hannobraun:
Previously, calling the `start` method while the timer was already
running had no effect, and the `cycles` argument was ignored. Now it
effectively restarts the timer.


hannobraun:
@nrf-rs/nrf52 Anyone up for reviewing/merging this? If not, I'll probably do it myself soon, since it's such a simple change.

jscarrott:
Looks fine. Sorry would have looked at this earlier but work has been very busy recently.

PR #26, opened 2018-09-28
Add interface to RNG peripheral

hannobraun:
I've only tested this on an nRF52832, but since it compiles, I assume it'll work on nRF52840, too.

### DIFF
--- a/nrf52-hal-common/src/lib.rs
+++ b/nrf52-hal-common/src/lib.rs
@@ -16,6 +16,7 @@ pub mod delay;
 pub mod spim;
 pub mod gpio;
 pub mod clocks;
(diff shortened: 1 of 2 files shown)


hannobraun:
@nrf-rs/nrf52 Still waiting for review :)

Note: When I originally submitted this, I was under the impression that there were no RNG traits in embedded-hal yet, and therefore didn't implement one. Turns out I was wrong.

I'll fix this later. There are some upcoming RNG changes in embedded-hal anyway, so it makes sense to wait a bit, until the dust has settled.

hannobraun:
Thanks for the review, @jamesmunns!

PR #27, opened 2018-10-07
Add TX-only uarte0

jamesmunns:
Some boards, such as the DWM1001-DEV, have this wired up to the on-board debugger, and is useful for outputting debug information.

This is based heavily on the SPIM driver

### DIFF
--- a/nrf52-hal-common/src/uarte.rs
+++ b/nrf52-hal-common/src/uarte.rs
@@ -1,0 +1,176 @@
+//! HAL interface to the UARTE peripheral
+//!
+//! See product specification:
+//!
+//! - nrf52832: Section 35
+//! - nrf52840: Section 6.34
+
+use core::ops::Deref;
+
+use target::{
+    uarte0,
+    UARTE0,
+};
+
+use prelude::*;
+use gpio::{
+    p0::P0_Pin,
+    Output,
+    PushPull,
+};
+
+// Re-export SVD variants to allow user to directly set values
+pub use target::uarte0::{
+    baudrate::BAUDRATEW as Baudrate,
+    config::PARITYW as Parity,
+};
+
+pub trait UarteExt: Deref<Target = uarte0::RegisterBlock> + Sized {
+    fn constrain(self, pins: Pins, parity: Parity, baudrate: Baudrate) -> Uarte<Self>;
+}
+
+impl UarteExt for UARTE0 {
+    fn constrain(self, pins: Pins, parity: Parity, baudrate: Baudrate) -> Uarte<Self> {
+        Uarte::new(self, pins, parity, baudrate)
+    }
+}
+
+
+/// Interface to a UARTE instance
+///
+/// This is a very basic interface that comes with the following limitations:
+/// - The UARTE instances share the same address space with instances of UART.
+///   You need to make sure that conflicting instances
+///   are disabled before using `Uarte`. See product specification:
+///     - nrf52832: Section 15.2
+///     - nrf52840: Section 6.1.2
+pub struct Uarte<T>(T);
+
+impl<T> Uarte<T> where T: UarteExt {
+    pub fn new(uarte: T, mut pins: Pins, parity: Parity, baudrate: Baudrate) -> Self {
+        // Select pins
+        pins.rxd.set_high();
+        uarte.psel.rxd.write(|w| {
+            let w = unsafe { w.pin().bits(pins.rxd.pin) };
+            w.connect().connected()
+        });
+        pins.txd.set_high();
+        uarte.psel.txd.write(|w| {
+            let w = unsafe { w.pin().bits(pins.txd.pin) };
+            w.connect().connected()
+        });
+
+        // Optional pins
+        uarte.psel.cts.write(|w| {
+            if let Some(ref pin) = pins.cts {
+                let w = unsafe { w.pin().bits(pin.pin) };
+                w.connect().connected()
+            } else {
+                w.connect().disconnected()
+            }
+        });
+
+        uarte.psel.rts.write(|w| {
+            if let Some(ref pin) = pins.rts {
+                let w = unsafe { w.pin().bits(pin.pin) };
+                w.connect().connected()
+            } else {
+                w.connect().disconnected()
+            }
+        });
+
+        // Enable UARTE instance
+        uarte.enable.write(|w|
+            w.enable().enabled()
+        );
+
+        // Configure
+        let hardware_flow_control = pins.rts.is_some() && pins.cts.is_some();
+        uarte.config.write(|w|
+            w.hwfc().bit(hardware_flow_control)
+             .parity().variant(parity)
+        );
+
+        // Configure frequency
+        uarte.baudrate.write(|w|
+            w.baudrate().variant(baudrate)
+        );
+
+        Uarte(uarte)
+    }
+
+    /// Write via UARTE
+    ///
+    /// This method uses transmits all bytes in `tx_buffer`
+    ///
+    /// The buffer must have a length of at most 255 bytes
+    pub fn write(&mut self,
+        tx_buffer  : &[u8],
+    )
+        -> Result<(), Error>
+    {
+        // This is overly restrictive. See (similar SPIM issue):
+        // https://github.com/nrf-rs/nrf52/issues/17
+        if tx_buffer.len() > u8::max_value() as usize {
+            return Err(Error::TxBufferTooLong);
+        }
+
+        // Set up the DMA write
+        self.0.txd.ptr.write(|w|
+            // We're giving the register a pointer to the stack. Since we're
+            // waiting for the UARTE transaction to end before this stack pointer
+            // becomes invalid, there's nothing wrong here.
+            //
+            // The PTR field is a full 32 bits wide and accepts the full range
+            // of values.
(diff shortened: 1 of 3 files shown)


wez:
(I'm traveling through the end of this next week so will have unpredictable bandwidth for reviewing)

nrf52-hal-common/src/uarte.rs
```diff
@@ -0,0 +1,153 @@
+//! HAL interface to the UARTE peripheral
+//!
+//! See product specification, chapter 33.
```
hannobraun: Wrong chapter number. It's chapter 35 in the nRF52832 product specification, chapter 6.34 for nRF52840.

nrf52-hal-common/src/uarte.rs
```diff
@@ -0,0 +1,153 @@
...
+        Uarte::new(self, pins, parity, baudrate)
+    }
+}
+
+/// Interface to a UARTE instance
+///
+/// This is a very basic interface that comes with the following limitations:
+/// - The UARTE instances share the same address space with instances of UART.
+///   You need to make sure that conflicting instances
+///   are disabled before using `Uarte`. See product specification, section 15.2.
```
hannobraun: That reference to the product specification is specific to nRF52832. It's section 6.1.2 for nRF52840.

nrf52-hal-common/src/uarte.rs
```diff
@@ -0,0 +1,153 @@
...
+            // of values.
+            unsafe { w.ptr().bits(tx_buffer.as_ptr() as u32) });
+        self.0.txd.maxcnt.write(|w|
+            // We're giving it the length of the buffer, so no danger of
+            // accessing invalid memory. We have verified that the length of the
+            // buffer fits in an `u8`, so the cast to `u8` is also fine.
+            //
+            // The MAXCNT field is 8 bits wide and accepts the full range of
+            // values.
+            unsafe { w.maxcnt().bits(tx_buffer.len() as _) });
```
hannobraun: I think this is fine as it is, but we should make a mental note (or better yet, open an issue) about creating a DMA API, so all the different peripherals don't have to duplicate this code.

nrf52-hal-common/src/uarte.rs
```diff
@@ -0,0 +1,153 @@
...
+        // Start UART Receive transaction
+        self.0.tasks_starttx.write(|w|
+            // `1` is a valid value to write to task registers.
+            unsafe { w.bits(1) });
+
+        // Wait for transmission to end
+        while self.0.events_endtx.read().bits() == 0 {}
+
+        // Reset the event, otherwise it will always read `1` from now on.
+        self.0.events_endtx.write(|w| w);
```
hannobraun: Same as the previous comment: We should create a generic event API, so all peripheral code doesn't have to remember to reset event flags and such.

The regularity and well-thought-out-ed-ness of the Nordic peripheral APIs lends itself really well to a mid-level API layer, between the low-level svd2rust API and the high-level HAL API.

nrf52-hal-common/src/uarte.rs
```diff
@@ -0,0 +1,153 @@
...
+
+        if self.0.txd.amount.read().bits() != tx_buffer.len() as u32 {
+            return Err(Error::Transmit);
+        }
+
+        Ok(())
+    }
+
+    /// Return the raw interface to the underlying UART peripheral
```
hannobraun: Nit: Should be `UARTE` instead of `UART`.

nrf52-hal-common/src/uarte.rs
```diff
@@ -72,23 +67,18 @@ impl<T> Uarte<T> where T: UarteExt {
...
-                .hwfc().bit(
-                    pins.rts.is_some() && pins.cts.is_some()
-                )
-                .parity().variant(parity)
-        );
+        uarte.config.write(|w| {
+            w.hwfc()
+                .bit(pins.rts.is_some() && pins.cts.is_some())
+                .parity()
+                .variant(parity)
```
hannobraun: Call me unsophisticated, but I never understood what rustfmt is good for. And here's an example where it makes the code much less readable. `variant` clearly belongs behind `parity`, and `bit` belongs behind `hwfc`.

I'm not going to make a fuss about this, if you disagree, but in my opinion this PR would be much improved, if you dropped that last commit.

hannobraun: Oops, I'm referring to the rustfmt commit (3 of 4), not the last commit. That one's good :)

jamesmunns: I'm in general a fan of Rustfmt, however in this case (and really, when peripheral crates in general) it tends to make poor tradeoffs.

I'll go ahead and rebase that commit out, I specifically made it it's own commit for that reason :)

nrf52-hal-common/src/uarte.rs
jamesmunns: I was thinking about that, the interesting part will be handling different cases around which/how many events to set for each peripheral. For example, SPIM has one for tx+rx, while UARTE has one for each of tx and rx, as they are independent. I'm sure we can find something though.

nrf52-hal-common/src/uarte.rs
jamesmunns: Ah, nice. I don't have a copy of the 840 OPS (haven't worked on that chip before), so I will make sure to pull that down and update it. I was hoping they would be laid out the same :)

nrf52-hal-common/src/uarte.rs
jamesmunns: Hm, I might have a stale version of the datasheet. It is Chapter 33 for me: https://send.firefox.com/download/46a875af6e/#CP1iWUg5VDe1gF2AawRvKg (OPS v0.6.3, which sounds suspiciously non-final).

nrf52-hal-common/src/uarte.rs
jamesmunns: It's probably good to think about. Right now I am working on a more actively DMA driven interface, which allows the user to trade the `Uarte` struct for basically a `(Sender<&[u8]>, Receiver<&[u8]>)` tuple, backed by a set of `heapless` `RingBuffer`s.

This would effectively double-buffer the incoming and outgoing messages, and the user can always `free()` the tuple pair to get a `UARTE0` or `Uarte` object back.

Once it actually works, I'll chime back in. IMO, we can have multiple different DMA abstractions (e.g. as a blocking tx/rx, as an async stream, etc.) that can coexist in the code, and the `UARTE0` periph is just used as an ownership token forcing only one to be active at a time.

nrf52-hal-common/src/uarte.rs
hannobraun: I have 1.4, which I just confirmed is the current one. I win :)

nrf52-hal-common/src/uarte.rs
hannobraun: I was thinking about something less ambitious: Just a `write` method that takes a slice and handles the `ptr` and `maxcnt` registers (plus all the associated casting and `unsafe`) under the hood. Basically, just a friendlier interface that the HAL implementations can use internally.

Thinking about higher-level abstractions along the lines you presented here is beneficial too, of course. I don't have much of an opinion about how that should look like yet.

nrf52-hal-common/src/uarte.rs
hannobraun: Here, too, I was thinking about something less ambitious. Something like that:
``` rust
impl Event {
    pub fn peek(&self) -> bool {
        // Return bit from register
    }

    pub fn consume(&mut self) -> bool {
        // Return bit from register and reset it
    }
}
```
And have a macro that generates that for every `EVENTS_` register, to cut down on duplicated code.

nrf52-hal-common/src/uarte.rs
jamesmunns: Pushed my WIP here as `uarte0-async` in case you feel like looking (totally not ready for serious review, doesn't work at all yet): https://github.com/nrf-rs/nrf52-hal/blob/uarte0-async/nrf52-hal-common/src/uarte_stream.rs

The big things to do still:
* Implement the interrupt handler (basically check which event triggered, reload and restart the buffer if there is something left to do)
* Write interfaces for the `Handle` structs, basically shove stuff in the buffer and trigger an event if not already in progress, probably by setting a flag and calling the inner handler directly
* Figure out if there is a less insane way of setting the handler-dispatch, allowing for different inner functions to service the interrupt, depending on mode.
* Figure out the interface for configuring buffer length/passing in buffers (mostly around lifetimes). I might do something along the lines of [this](https://github.com/jamesmunns/mcf-config-demo)
* Figure out what to do about early-flushing of the receive buffer. Right now, there isn't really a way to flush out the items from the DMA without polling. One of the forums mentioned tying a GPIO (internally?) to the RX line, and using a timer to check for idle states longer than some window, or we could use a 1-shot time to check for the amount of data received (and if unchanged for some amount of time), but I don't have a great elegant answer here.

Sorry to go off track of this review :)

jamesmunns:
@hannobraun I have backed out the formatting, and cleaned up the remaining nits. Let's save the DMA API for another PR.

Ready to merge if you are happy.

jamesmunns:
Squashed (forgot I could do it through the UI), merging.

PR #31, opened 2018-10-23
Implement basic TWIM (I2C) API

hannobraun:


nrf52-hal-common/src/twim.rs
```diff
@@ -0,0 +1,258 @@
...
+impl_twim_ext!(
+    TWIM0,
+    TWIM1,
+);
+
+
+/// Interface to a TWIM instance
+///
+/// This is a very basic interface that comes with the following limitation:
+/// The TWIM instances share the same address space with instances of SPIM,
```
jamesmunns: It might be worth clarifying that the address space is shared with SPIMn, TWISn, etc.

For example, we can use SPIM0 with TWIM1, but not SPIM1 with TWIM1.

In the future, it might make sense to have these peripheral groups in a wrapped group in the HAL Peripherals block, so you can only take one at a time, e.g. something like:

```rust
struct PeripheralGroup3 {
    SPIM0: ...,
    SPIS0: ...,
    TWIM0: ...,
    ...,
}

impl PeripheralGroup3 {
pub fn into_spim_0(self) -> SPIM0 {}
// ...
}

hannobraun: I agree. Also see https://github.com/nrf-rs/nrf52-hal/issues/7, as that related to solving this issue and similar ones.

I'll update the documentation. I need to update the commit anyway, as I just noticed I wrote "SPIM" in the message where I should have written "TWIM".

hannobraun:
@jamesmunns Force-pushed a new commit with the changes we talked about.

PR #36, opened 2018-12-24
RTFM Minimal Example

jamesmunns:
I've been messing around with RTFM a bit, I thought I would add a minimal skeleton in case others found it useful.

EDIT: This PR supports both the 52832 and 52840.


examples/rtfm-demo/src/main.rs
```diff
@@ -0,0 +1,33 @@
...
+
+#[cfg(feature = "52832")]
+use nrf52832_pac;
+
+#[cfg(feature = "52840")]
+use nrf52840_pac;
+
+
+#[cfg_attr(feature="52832", app(device = nrf52832_pac))]
+#[cfg_attr(feature="52840", app(device = nrf52840_pac))]
```
wez: nice; I was sightly concerned that dealing with this might get super ugly, but it looks pretty decent.

examples/rtfm-demo/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+name = "rtfm-demo"
+version = "0.1.0"
+authors = ["James Munns <james.munns@ferrous-systems.com>"]
+edition = "2018"
+
+[dependencies]
+cortex-m-rtfm = "0.4"
```
wez: I've been using `"~0.4"` rather than an exact version for these in other personal projects... hasn't caused me any problems yet(!)   Do we have a preference for one over the other?

hannobraun: `0.4` and `~0.4` are identical. It's `=0.4` that's different (exact version requirement).

We should have a clear preference for `0.4`/`~0.4`, in my opinion. Exact version requirements will sooner or later lead to trouble in someone's build, due to mismatched version requirements.

PR #38, opened 2018-12-28
Version 0.6.1 Release

jamesmunns:
Issuing a point release to include the support for pullups for GPIOs.

NB: I have changed the dependencies in the `nrf528xx-hal` crates to use an **exact** dependency on the `nrf52-hal-common` crate. I do not believe there would be a case where you would ever want the `-hal` and `-hal-common` to differ, so this matches the "works as expected" case IMO.

### DIFF
--- a/nrf52-hal-common/Cargo.toml
+++ b/nrf52-hal-common/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "nrf52-hal-common"
-version = "0.6.0"
+version = "0.6.1"
 description = "Common HAL for the nRF52 family of microcontrollers.  More specific HAL crates also exist."
 
 repository = "https://github.com/nrf-rs/nrf52-hal"
--- a/nrf52832-hal/Cargo.toml
+++ b/nrf52832-hal/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "nrf52832-hal"
-version = "0.6.0"
+version = "0.6.1"
 description = "HAL for nRF52832 microcontrollers"
 
 repository = "https://github.com/nrf-rs/nrf52-hal"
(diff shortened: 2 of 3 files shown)


hannobraun:
@jamesmunns 
> NB: I have changed the dependencies in the `nrf528xx-hal` crates to use an **exact** dependency on the `nrf52-hal-common` crate. I do not believe there would be a case where you would ever want the `-hal` and `-hal-common` to differ, so this matches the "works as expected" case IMO.

I'm not sure about this.
1. What problem does this solve? Semver should be enough here.
2. I am absolutely sure that sooner or later someone (maybe me?) will forget to update this, and then this will cause havoc in someone's Cargo build.

Unless you can demonstrate a clear advantage, I'm opposed to this change.

jamesmunns:
@hannobraun actually, I think you are right. I forgot that increasing the dependency to "0.6.1" (not "=0.6.1") will still disallow the usage of 0.6.0, causing the `common` crate to be updated correctly. I was trying to avoid the case where `nrf53832-hal` WAS updated, but due to an existing lock file, `nrf52-common-hal` would NOT be updated automatically.

IMO, there is no harm in tightly coupling versions of `nrf528xx-hal` and `nrf52-common-hal`, as the former (at least today) is largely a thin wrapper around the latter. However, I don't have a compelling reason to favor strict version dependencies rather than the normal `>=` semver behavior, so I will revert this.

> I am absolutely sure that sooner or later someone (maybe me?) will forget to update this, and then this will cause havoc in someone's Cargo build.

I think failure to update dependency versions is likely to cause unexpected behavior, at least missing functionality, if not havoc in any case. Perhaps we should add a CI check that all three HAL crates have matching version numbers? My [tomlq](https://github.com/jamesmunns/tomlq) binary crate should make this pretty easy to do from the command line.

NOTE: All of my comments above ONLY apply to the `nrf52*-hal` crates, not to the dependencies on `nrf52*-pac` crates.

hannobraun:
@jamesmunns 
> Perhaps we should add a CI check that all three HAL crates have matching version numbers? My [tomlq](https://github.com/jamesmunns/tomlq) binary crate should make this pretty easy to do from the command line.

Sounds good.

Issue #17, opened 2018-09-13
Allow larger SPI buffer sizes on nrf52840 in nrf52-hal-common


wez:
the `maxcnt` register in the SPI hardware is 8 bits wide on the nrf52382 but wider on the 52840.  The boundary checking code currently constrains the maximum buffer size to 8 bit values which is safe but arbitrarily limiting on the 52840.

Come up with a reasonable way to express this range check that allows taking maximum advantage for the different target devices.

jamesmunns:
Would it make sense to have some kind of `target` module that collects all these details? This would make it easier to conditionally compile. In that target module, we could define constants. Then in the driver, we could do something like:

```rust
#[cfg(target = "nrf52832")]
mod target {
    mod spi {
        pub const MAX_BYTE_COUNT: usize = 255;
    }
}

#[cfg(target = "nrf52840")]
mod target {
    mod spi {
        pub const MAX_BYTE_COUNT: usize = 65535; // made up number
    }
}

// ...

use target::spi::MAX_BYTE_COUNT;

for _ in 0..MAX_BYTE_COUNT { }
```

Issue #10, opened 2018-09-12
Investigate implementing embedded-hal traits for SPIM peripheral


hannobraun:
This issue was first opened in the [old nrf52-hal repository](https://github.com/jamesmunns/nrf52-hal). Original discussion: https://github.com/jamesmunns/nrf52-hal/issues/11

---

I didn't do this initially, because I'm not sure whether it's possible, or if it is, if it should be done without modifying embedded-hal first. The problem is that the [embedded-hal trait for SPI](https://docs.rs/embedded-hal/0.2.1/embedded_hal/spi/trait.FullDuplex.html) is very basic, reading and writing only single words. But SPIM supports DMA natively. The most straight-forward way to use it is to actually let it handle the whole SPI transaction.

I now discovered that there are [higher-level SPI traits](https://docs.rs/embedded-hal/0.2.1/embedded_hal/blocking/spi/index.html), but those have problems, too. [Transfer](https://docs.rs/embedded-hal/0.2.1/embedded_hal/blocking/spi/trait.Transfer.html) uses the same buffer for sending and receiving. I'm not sure if this can be implemented for SPIM.

Maybe my doubts are unfounded and there is a reasonable way to implement those traits. Maybe the embedded-hal traits need to be changed. I'm pretty sure there's discussion about SPI somewhere on the embedded-hal issue tracker that might need to be consulted.

PR #39, opened 2018-12-29
Implemented embedded_hal Transfer trait for SPIM and removed EasyDMA limitation for UARTE and TWIM

simonsso:
Hi
Here is an implementation for the embedded_hal transfer trait. #10  It is based on the existing DMA setup code but CS is handled outside in embedded_hal and I set up multiple transfers if size is over the limit.

I also wrote some code for #17 not as suggested in the ticket but that could be refactored later.

### DIFF
--- a/nrf52-hal-common/src/lib.rs
+++ b/nrf52-hal-common/src/lib.rs
@@ -30,7 +30,7 @@ pub mod prelude {
     pub use crate::twim::TwimExt;
     pub use crate::uarte::UarteExt;
 }
-/// Length of Nordoc EasyDMA differs for MCUs
+/// Length of Nordic EasyDMA differs for MCUs
 const fn easy_dma_size() -> usize {
     // NRF52832 8 bits1..0xFF
     #[cfg(feature = "52832")]
--- a/nrf52-hal-common/src/lib.rs
+++ b/nrf52-hal-common/src/lib.rs
@@ -30,15 +30,20 @@ pub mod prelude {
     pub use crate::twim::TwimExt;
     pub use crate::uarte::UarteExt;
 }
+
 /// Length of Nordic EasyDMA differs for MCUs
-const fn easy_dma_size() -> usize {
+#[cfg(feature = "52832")]
+pub mod target_constants{
     // NRF52832 8 bits1..0xFF
-    #[cfg(feature = "52832")]
-    {255}
-    // NRF52840 16 bits 1..0xFFFF
-    #[cfg(feature = "52840")]
-    {65535}
+    pub const EASY_DMA_SIZE:usize = 255;
 }
+    // NRF52840 16 bits 1..0xFFFF
+#[cfg(feature = "52840")]
+pub mod target_constants{
+    pub const EASY_DMA_SIZE:usize = 65535;
+}
+
+
 pub use crate::clocks::Clocks;
 pub use crate::delay::Delay;
 pub use crate::spim::Spim;
--- a/nrf52-hal-common/src/lib.rs
+++ b/nrf52-hal-common/src/lib.rs
@@ -30,8 +30,15 @@ pub mod prelude {
     pub use crate::twim::TwimExt;
     pub use crate::uarte::UarteExt;
 }
-
-
+/// Length of Nordoc EasyDMA differs for MCUs
+const fn easy_dma_size() -> usize {
+    // NRF52832 8 bits1..0xFF
+    #[cfg(feature = "52832")]
+    {255}
+    // NRF52840 16 bits 1..0xFFFF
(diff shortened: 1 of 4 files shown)


nrf52-hal-common/src/spim.rs
```diff
@@ -55,6 +61,92 @@ impl_spim_ext!(
 /// - The over-read character is hardcoded to `0`.
 pub struct Spim<T>(T);
 
+impl<T> Transfer<u8> for Spim<T> where T: SpimExt
+{
+   type Error = Error;
+
+    fn transfer<'w>(&mut self, words: &'w mut [u8]) -> Result<&'w [u8], Error> {
+        // TODO: some targets have a maxcnt whose size is larger
```
jamesmunns: Is this comment still true? I think the feature flags that you do on lines 48-51 means this comment can be removed

nrf52-hal-common/src/spim.rs
```diff
@@ -55,6 +61,92 @@ impl_spim_ext!(
...
+        // than a u8, so this length check is overly restrictive
+        // and could be lifted.
+        if words.len() > MAX_SPI_DMA_SIZE as usize {
+            return Err(Error::TxBufferTooLong);
+        }
+
+        let mut offset:u32 = 0;
+        while offset < words.len() as u32 {
+            let datalen = min(MAX_SPI_DMA_SIZE,(words.len() as u32 )- offset);
```
```suggestion
            let datalen = min(MAX_SPI_DMA_SIZE, (words.len() as u32) - offset);
```

nrf52-hal-common/src/spim.rs
```diff
@@ -55,6 +61,92 @@ impl_spim_ext!(
 /// - The over-read character is hardcoded to `0`.
 pub struct Spim<T>(T);
 
+impl<T> Transfer<u8> for Spim<T> where T: SpimExt
+{
+   type Error = Error;
+
+    fn transfer<'w>(&mut self, words: &'w mut [u8]) -> Result<&'w [u8], Error> {
```
jamesmunns: There is a fair amount of duplicated code here between `Spim<T>::write()` and this method. Would it make sense to refactor some of the common parts into an inline-able function?

nrf52-hal-common/src/spim.rs
simonsso: Thanks, not only the comment but also the code it commented should be removed!

nrf52-hal-common/src/spim.rs
jamesmunns: Another option could be to have the `write` function call the `transfer` function, basically only handling the CS GPIO?

simonsso: I will see if I could make a nice refactoring of it, almost the same code is on three places.

simonsso:
So I have joined all to one helper function and use it from all three places. But I have not been able to verify the other functions. I don't think moving the handling of CS a few instructions will matter

simonsso:
Some more updates, I applied the EasyDMA fix for UARTE too, and created a shared definition. The same code will probably work for TWIM too but I don't have any thing here to verify that with so I leave it as it is for now.

nrf52-hal-common/src/lib.rs
```diff
@@ -30,8 +30,15 @@ pub mod prelude {
     pub use crate::twim::TwimExt;
     pub use crate::uarte::UarteExt;
 }
-
-
+/// Length of Nordoc EasyDMA differs for MCUs
```
Yatekii: Typo ;)

simonsso:
Should I update something more on this PR?

jamesmunns:
Hey @simonsso, sorry for the delay here, I've been working on some other stuff, and haven't been able to take a closer look at the interface you've added here. If @hannobraun has some time, he might be able to provide some suggestions.

simonsso:
SPI is now in new PR #48  -- when the code is reviewed there for SPI the same structure can be applied to TWIM and UART

simonsso:
@jamesmunns  is it ok to make a overwrite-merge (force push to branch) to this PR for the final part, or will it be better if I create a new and abandon this?

hannobraun:
> is it ok to make a overwrite-merge (force push to branch) to this PR for the final part, or will it be better if I create a new and abandon this?

I can't speak for James, but I think it's probably better to open a new PR with the remaining changes. I don't feel strongly about this though, so do what you want :-)

jamesmunns:
@simonsso can this be closed now that #48 has landed?

simonsso:
There are changes on this for i2c and uart too I think but we are close to clean up time for this.

simonsso:
Original change is merged in #48  -- closing this and creating a new for the other changes

PR #40, opened 2019-01-03
Don't Merge: Initial working commit of an async uarte

jamesmunns:
This branch is using (a hacked up version of) [`bbqueue`](https://github.com/jamesmunns/bbqueue) to act as a SPSC.

Won't build because it relies on some hacked up repos I needed to get it working, but worth sharing anyway.

Edit: This is now based on the released 0.2.1 version of BBQueue.


jamesmunns:
Note: This branch uses (essentially) the `0.0.2` release of `BBQueue`. I'll update this PR later today to split the hacked up interface into a new file (`uarte::UarteAsync` or so), so that this can live side by side with the existing blocking `Uarte`.

nrf52-hal-common/src/uarte.rs
```diff
@@ -19,22 +25,37 @@ use crate::gpio::{
...
+
 // Re-export SVD variants to allow user to directly set values
 pub use crate::target::uarte0::{
     baudrate::BAUDRATEW as Baudrate,
     config::PARITYW as Parity,
 };
 
 pub trait UarteExt: Deref<Target = uarte0::RegisterBlock> + Sized {
-    fn constrain(self, pins: Pins, parity: Parity, baudrate: Baudrate) -> Uarte<Self>;
+    fn constrain(self, nvic: NVIC, pins: Pins, parity: Parity, baudrate: Baudrate) -> Uarte<Self>;
```
hannobraun: Won't that be far too limiting, requiring ownership of the whole `NVIC` like that?

I've been thinking for a while that we need a higher-level layer on top of NVIC, where each interrupt is its own ownable/movable thing. It's one of those things I want to prototype, but haven't gotten to yet.

nrf52-hal-common/src/uarte.rs
```diff
@@ -19,22 +25,37 @@ use crate::gpio::{
...
+    fn constrain(self, nvic: NVIC, pins: Pins, parity: Parity, baudrate: Baudrate) -> Uarte<Self> {
+        Uarte::new(self, nvic, pins, parity, baudrate)
     }
 }
 
+lazy_static! {
+    static ref MUH_BUFFAH: BBQueue<U2048> = {
+        BBQueue::new()
+    };
+}
```
hannobraun: This is a single `static`, and this module only supports UARTE0 right now, so this works. But what if we add support for other UARTE instances?

Maybe we could make the user allocate the buffer for us? That would mean the API would need an additional lifetime, but the user can always fall back to allocating the buffer in a `static` (thereby making the lifetime `'static`), if that causes any trouble.

nrf52-hal-common/src/uarte.rs
```diff
@@ -96,75 +121,44 @@ impl<T> Uarte<T> where T: UarteExt {
...
-    )
-        -> Result<(), Error>
-    {
-        // This is overly restrictive. See (similar SPIM issue):
-        // https://github.com/nrf-rs/nrf52/issues/17
-        if tx_buffer.len() > u8::max_value() as usize {
-            return Err(Error::TxBufferTooLong);
-        }
+        unsafe {
+            MAYBE_CONSUMER = Some(cons)
```
hannobraun: Why safe the consumer in a `static` instead of returning it to the user?

nrf52-hal-common/src/uarte.rs
jamesmunns: I moved this to requiring `&mut NVIC`, only needed during the scope of `::new()`.

nrf52-hal-common/src/uarte.rs
jamesmunns: Updated the interface to take a `Producer` and `Consumer`, which no longer require lifetimes (they explicitly require a 'static buffer for backing, but I have some helper methods now for that in v0.2.1).

nrf52-hal-common/src/uarte.rs
jamesmunns: The current implementation of `Uarte` only supports sending data over serial. In this instance, the Producer is the application code, sending data, and the Consumer is used by the DMA/interrupt, to feed the outgoing data.

jamesmunns:
> By the way, how familiar are you with Jorge's article about DMA? I don't recall enough of the details to know whether it will be helpful here, but I've immersed myself in it in the past and believe it is a good approach. (For reference, here's my own partial implementation.)

I'm not terribly familiar, unfortunately. This is approaching the problem in a different way though (I think), leaning less heavily on futures style abstractions, and rather being a more straightforward "stream" implementation. I think it could definitely be possible to impl `Reader` and `Writer` with this.

To be honest, I'm open to having multiple parallel versions of each peripheral, perhaps abstracting some of the common behavior to a central `mod.rs`. Since each "style" of peripheral requires ownership of the `-pac` peripheral, we don't have to worry about the user trying to mix and match.

nrf52-hal-common/src/uarte.rs
jamesmunns: Also if we add support for other UARTE instances, I think those will need their own separate interrupt handler, where we can grab the "proper" `static Consumer`, and combine the logic. Something like

```rust
#[interrupt]
unsafe fn UARTE0_UART0() {
    let uart = &*UARTE0::ptr();
    let cons: &mut Consumer = MAYBE_CONSUMER_0.as_mut().unwrap();
    common_handler(uart, cons);
}

#[interrupt]
unsafe fn UARTE1_UART1() {
    let uart = &*UARTE1::ptr();
    let cons: &mut Consumer = MAYBE_CONSUMER_1.as_mut().unwrap();
    common_handler(uart, cons);
}
```

nrf52-hal-common/src/uarte_async.rs
```diff
@@ -0,0 +1,288 @@
...
+        self.prod.commit(tx_buffer.len(), w_grant);
+
+        NVIC::pend(Interrupt::UARTE0_UART0);
+
+        Ok(())
+    }
+
+    /// Return the raw interface to the underlying UARTE peripheral
+    pub fn free(self) -> T {
+        self.periph
```
jamesmunns: I should probably disable interrupts and close any open DMA transactions here, and we should maybe disable the UARTE peripheral before giving it back.

nrf52-hal-common/src/uarte_async.rs
```diff
@@ -0,0 +1,288 @@
...
+    }
+}
+
+
+pub struct Pins {
```
jamesmunns: Stuff like this should probably move to a shared `mod.rs` between `uarte.rs` and `uarte_async.rs`.

nrf52-hal-common/src/uarte_async.rs
```diff
@@ -0,0 +1,288 @@
...
+
+    if end_tx {
+        uart.events_endtx.write(|w| w.bits(0));
+    }
+    let cons: &mut Consumer = MAYBE_CONSUMER.as_mut().unwrap();
+
+    use crate::uarte_async::DmaState::*;
+    let mut stat = Idle;
+
+    ::core::mem::swap(&mut stat, &mut DMA_STATUS);
```
jamesmunns: I should add a note here that this is "safe" to do, since we are in interrupt condition, and I don't think it is possible for interrupts to be re-entrant.

nrf52-hal-common/src/uarte_async.rs
```diff
@@ -0,0 +1,288 @@
...
+    let rgrant = match cons.read() {
+        Ok(gr) => gr,
+        Err(BBQError::InsufficientSize) => {
+            // No more data to send
+            return;
+        }
+        Err(_) => panic!("queue error"),
+    };
+
+    let sz = min(255, rgrant.buf.len());
```
jamesmunns: This should not be 255, but instead something that makes sense for both the 832 and 840

nrf52-hal-common/src/uarte_async.rs
```diff
@@ -0,0 +1,288 @@
...
+
+    // Conservative compiler fence to prevent optimizations that do not
+    // take in to account actions by DMA. The fence has been placed here,
+    // before any DMA action has started
+    compiler_fence(SeqCst);
+
+    // Set up the DMA write
+    uart.txd.ptr.write(|w|
+        // We're giving the register a pointer to the stack. Since we're
```
jamesmunns: Comments are no longer true

hannobraun:
@jamesmunns 
> > By the way, how familiar are you with Jorge's article about DMA? I don't recall enough of the details to know whether it will be helpful here, but I've immersed myself in it in the past and believe it is a good approach. (For reference, here's my own partial implementation.)
> 
> I'm not terribly familiar, unfortunately. This is approaching the problem in a different way though (I think), leaning less heavily on futures style abstractions, and rather being a more straightforward "stream" implementation. I think it could definitely be possible to impl `Reader` and `Writer` with this.
> 
> To be honest, I'm open to having multiple parallel versions of each peripheral, perhaps abstracting some of the common behavior to a central `mod.rs`. Since each "style" of peripheral requires ownership of the `-pac` peripheral, we don't have to worry about the user trying to mix and match.

Thanks for the explanation. I agree that parallel approaches are probably the right way to go at this point. Maybe one approach will "win" and we can deprecate some or all of the others in time, but right now too much is in flux, I think. Rust's async story is still developing and it's not clear to me how that will affect nb/embedded-hal and HALs in general.

PR #41, opened 2019-01-17
Basic implementation of the nRF52840-DK BSP

Yatekii:
I did a basic impl for the nRF52840-DK.
So far I only tested it with a simple blinky example.
It needs:
- docs
- better examples
- support for buttons (only the pin reservations are done atm)

I am open for any suggestions, as I am not too familiar with the rust embedded ecosystem, so this impl could be suboptimal. I used @hannobraun's DW1001-Dev impl as a starting point. I did refrain from using a feature flag to enable/disable devboard functionalities as I feel there is sooo many IOs with a full mux, we dont need the pins reserved for the COM UART or the flash SPI to have dual functionality, even tho the HW allows it via Jumpers.

Maybe it is suboptimal to use a fixed UART and SPI peripheral for those functionalities as the chip can do a full mux and maybe a user wants a different peripheral for whatever reason ...

edit: formatting


boards/nRF52840-DK/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+license = "MIT OR Apache-2.0"
+name = "nrf52840-dk-bsp"
+version = "0.1.0"
+edition = "2018"
+
+[dependencies]
+cortex-m = { version = "0.5.4", features = [ "const-fn" ] }
+cortex-m-rt = "0.6.5"
+embedded-hal = "0.2.1"
+nrf52840-hal = { path = "../../nrf52840-hal" }
```
hannobraun: I think this path-only-dependency will prevent the crate from being published on crates.io. I believe the correct way to handle this is to keep the path, but add a `version` in addition.

boards/nRF52840-DK/src/lib.rs
```diff
@@ -0,0 +1,580 @@
+//! Board support crate for the Nordic nRF52840-DK
+//!
+//! This crate is in early development. Not much to see here, right now.
+#![no_std]
+
+//#![deny(missing_docs)]
+#![deny(warnings)]
```
hannobraun: I know `#![deny(warnings)]` is everywhere, but it's plain wrong and should be avoided. Detailed explanation and alternative here: https://github.com/rust-unofficial/patterns/blob/master/anti_patterns/deny-warnings.md

boards/nRF52840-DK/src/lib.rs
```diff
@@ -0,0 +1,580 @@
...
+        Uarte,
+        Parity as UartParity,
+        Baudrate as UartBaudrate,
+    },
+};
+
+/// Provides access to all features of the nRF52840-DK board
+#[allow(non_snake_case)]
+#[allow(non_camel_case_types)]
+pub struct nRF52840DK {
```
hannobraun: Just an idea: I've been thinking for a while that `struct`s such as this should be called `Board`, like the peripheral struct PAC crates (and at least one HAL crate) is called `Peripherals`. This would also remove the repetition in the fully qualified name.

Not sure about this. Any comments?

boards/nRF52840-DK/Cargo.toml
Yatekii: Oh, thanks for the hint! I will change this.

boards/nRF52840-DK/src/lib.rs
Yatekii: Actually I am kinda sure I copied this from your crate too ;) Thanks for the hint tho; I will remove it! My OCD dictates me to remove all warnings anyways :D

boards/nRF52840-DK/src/lib.rs
Yatekii: I agree, this would make sense. I will change this up!

What I am not quite sure about yet too is how I make the pins stored in here available to different owners iside an application lateron. The Peripheral allows you to take certain pins from certain places afaik. This is not the case here, right?

boards/nRF52840-DK/src/lib.rs
hannobraun: Yes, I'm as guilty as anyone :)

I'm going to remove these my crates as I'm working on them. So far most of them still have it.

boards/nRF52840-DK/src/lib.rs
hannobraun: After calling `take` or `steal`, the application will own `nRF52840DK`/`Board`, which includes ownership of `Pins` within it. Owning the pins means they can be moved out or borrowed as required.

Does that answer your question? Not sure I understood it correctly.

Yatekii: Yeah, but you cannot move the Pins out of Board as it would make the fields of the struct uninitialized which is forbidden ;) It would consume the entire struct as far as I understand it.

hannobraun: That's not exactly right. Moving a field out of a struct does two things:
- It makes that field inaccessible.
- It will prevent any use that requires the whole struct.

What this means is, you can no longer move or borrow the whole struct (which including calling methods that require `self`/`&self`/`&mut self`). You can still access any other fields though. I'm also relatively sure that you can make the struct usable again as a whole by moving stuff back into the fields you previously moved out of.

The main guideline to pay attention to is, for a given struct, follow one style or the other. Either design your struct with open fields that can be moved out of, or add methods to it that require some form of `self`. Never mix both in the same struct, unless you explicitly want those methods to be uncallable while the struct is incomplete.

Yatekii: Well, there can be issues with moving out.
For example

```
self.s = consume_and_return(self.s);
```

can be hard to get working. There is [take_mut](https://docs.rs/take_mut/0.2.2/take_mut/) to get this working tho.
I also made this working at some places, but I am not quite sure yet when it works and when not.
Surely I will make some experiments :)
But for starters I am trying to get the SPI flash working to test out the SPI :)

nrf52-hal-common/src/spim.rs
```diff
@@ -118,10 +118,15 @@ impl<T> Spim<T> where T: SpimExt {
         // TODO: some targets have a maxcnt whose size is larger
         // than a u8, so this length check is overly restrictive
         // and could be lifted.
```
hannobraun: I think after this commit, this `TODO` no longer applies.

Yatekii: Well, it kind of does, as I did it wrong. Apparently the register is only 10 bits wide, not 16.
I will fix this as soon as I get UART and SPI working. Atm it's behaving strangely. Uart::write() completes without any errors but the pin does not actually toggle (It goes high during the initialization of the UART tho).
I have no clue what this could be but maybe there are some defs wrong. Still trying to figure svd2rust generated code out to verify :)
Thanks for the hint/reminder

boards/nRF52840-DK/src/lib.rs
Yatekii: Yep, worked absolutely great for me in the past =)

boards/nRF52840-DK/Cargo.toml
jamesmunns: @Yatekii still pending

boards/nRF52840-DK/src/lib.rs
```diff
@@ -0,0 +1,580 @@
...
+pub use cortex_m_rt;
+pub use embedded_hal;
+pub use nrf52840_hal as hal;
+
+/// Exports traits that are usually needed when using this crate
+pub mod prelude {
+    pub use nrf52840_hal::prelude::*;
+}
+
+// TODO:
```
jamesmunns: Remove TODO

boards/nRF52840-DK/src/lib.rs
```diff
@@ -0,0 +1,580 @@
...
+
+/// Provides access to all features of the nRF52840-DK board
+#[allow(non_snake_case)]
+#[allow(non_camel_case_types)]
+pub struct nRF52840DK {
+    /// The nRF52's pins which are not otherwise occupied on the nRF52840-DK
+    pub pins: Pins,
+
+    /// The nRF52840-DK UART which is wired to the virtual COM port
+    pub com: Uarte<nrf52::UARTE0>,
```
jamesmunns: `COM` is a bit of a windows-ism :) Maybe just `uart` or `uarte`?

boards/nRF52840-DK/src/lib.rs
```diff
@@ -0,0 +1,580 @@
...
+    pub TPIU: nrf52::TPIU,
+
+    /// nRF52 peripheral: FICR
+    pub FICR: nrf52::FICR,
+
+    /// nRF52 peripheral: UICR
+    pub UICR: nrf52::UICR,
+
+    // /// nRF52 peripheral: BPROT
+    // TODO:
```
jamesmunns: TODO?

boards/nRF52840-DK/src/lib.rs
```diff
@@ -0,0 +1,580 @@
...
+
+        let flash_cs = pins0.p0_17.into_push_pull_output(Level::High).degrade();
+        // let flash_wp = pins.p0_22.into_push_pull_output(Level::High).degrade();
+        // let flash_hold = pins.p0_23.into_push_pull_output(Level::High).degrade();
+
+        // The nRF52840-DK features a virtual COM port.
+        // It features HWFC but does not have to use it.
+        // It can transmit a flexible baudrate of up to 1Mbps.
+        let com_uart = p.UARTE0.constrain(uarte::Pins {
+                txd: pins0.p0_06.into_push_pull_output(Level::High).degrade(),
```
jamesmunns: This will need to change when #43 lands. Also dead code for cts/rts below

nrf52-hal-common/src/gpio.rs
```diff
@@ -51,21 +51,293 @@ pub enum Level {
...
+// ===============================================================
+// Implement Generic Pins for this port, which allows you to use
+// other peripherals without having to be completely rust-generic
+// across all of the possible pins
+// ===============================================================
+/// Generic $PX pin
+pub struct Pin<MODE> {
+    pub pin: u8,
+    #[cfg(feature = "52840")]
+    pub port: bool,
```
jamesmunns: I'd probably prefer a `u8` to a bool, though probably not much difference in practice.

boards/nRF52840-DK/Cargo.toml
Yatekii: Fixed

boards/nRF52840-DK/src/lib.rs
Yatekii: Fixed.

boards/nRF52840-DK/src/lib.rs
Yatekii: Fixed.

boards/nRF52840-DK/Cargo.toml
Yatekii: Fixed.

boards/nRF52840-DK/src/lib.rs
Yatekii: I left this for now, but I elaborated on why it is there. Hope that's ok

boards/nRF52840-DK/src/lib.rs
Yatekii: You are absolutely right. Renamed this to `cdc` as this would be the correct term :)

boards/nRF52840-DK/src/lib.rs
Yatekii: I left this, but elaborated on why it is there :)

boards/nRF52840-DK/src/lib.rs
Yatekii: Yes, I didn't change it yet as I have to wait for #43. No point in fixing it as well ;)
I reenabled HWFC as the CDC UART can do HWFC and it was eliminated as a potential source of error for the broken UART :)

nrf52-hal-common/src/gpio.rs
Yatekii: Well, I wanted an u8 too. BUT: The svd2rust files are generated this way. they use a bool too, guess because it saw it's only 1 bit wide. So I left this as a bool so you can directly use the .port field when you want to set this when selecting pins etc.
Also, I highly doubt there will ever be a third port. The nRF52 is a low power small form factor chip. It already has a lot of pins for this usecase I think. And there is still room for 16 more pins on the second port.

Yatekii:
I addressed all the pointed out flaws. If there is more concern, please just tell me right away :)

boards/nRF52840-DK/src/lib.rs
```diff
@@ -0,0 +1,579 @@
...
+    uarte::{
+        self,
+        Uarte,
+        Parity as UartParity,
+        Baudrate as UartBaudrate,
+    },
+};
+
+/// Provides access to all features of the nRF52840-DK board
+#[allow(non_snake_case)]
```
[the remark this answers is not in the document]
Yatekii: Well, without the snake case one it will generate a ton of warnings for all the fields.
I removed the other one tho.

boards/nRF52840-DK/src/lib.rs
```diff
@@ -0,0 +1,579 @@
...
+    /// Core peripheral: Trace Port Interface Unit
+    pub TPIU: nrf52::TPIU,
+
+    /// nRF52 peripheral: FICR
+    pub FICR: nrf52::FICR,
+
+    /// nRF52 peripheral: UICR
+    pub UICR: nrf52::UICR,
+
+    // /// nRF52 peripheral: BPROT
```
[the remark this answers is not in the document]
Yatekii: Opened #51 any criticism & additions welcome.

nrf52-hal-common/src/gpio.rs
```diff
@@ -51,21 +51,293 @@ pub enum Level {
...
+            &(*{
+                #[cfg(feature = "52832")]
+                { P0::ptr() }
+                #[cfg(feature = "52840")]
+                { if !self.port { P0::ptr() } else { P1::ptr() } }
+            }).pin_cnf[self.pin as usize]
+        }
+        .write(|w| {
+            w.dir().output()
+                .input().connect() // AJM - hack for SPI
```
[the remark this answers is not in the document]
Yatekii: What exactly do you mean with that comment and is there an issue to track this already? I would open one or fix it immediately if I knew.

Yatekii:
All changes should be in place as requested.

PR #43, opened 2019-01-23
Fix/Extend UARTE

### LINKED ISSUE #42: UARTE Pins are wrong
Right now, RXD and CTS of Uarte::Pins are of type output. This is wrong.

https://github.com/nrf-rs/nrf52-hal/blob/master/nrf52-hal-common/src/uarte.rs#L172-L177

From the datasheet:

| UARTE signal | UARTE pin | Direction | Output value |
| :--- | :--- | :--- | :--- |
| RXD | As specified in PSEL.RXD | Input | N/A |
| CTS | As specified in PSEL.CTS | Input | N/A |
| RTS | As specified in PSEL.RTS | Output | 1 |
| TXD | As specified in PSEL.TXD | Output | 1 |

jamesmunns:
Closes #42, also adds RX functionality with optional timeout to Uarte.

@hannobraun I'd be interested to hear your opinion on this approach for receiving with a fixed timeout.

### DIFF
--- a/nrf52-hal-common/src/timer.rs
+++ b/nrf52-hal-common/src/timer.rs
@@ -16,7 +16,7 @@ use crate::target::{
     TIMER3,
     TIMER4,
 };
-use void::Void;
+use void::{unreachable, Void};
 
 
 pub trait TimerExt : Deref<Target=timer0::RegisterBlock> + Sized {
@@ -149,6 +149,9 @@ impl<T> Timer<T> where T: TimerExt {
 
     pub fn delay(&mut self, cycles: u32) {
         self.start(cycles);
-        block!(self.wait()).unwrap();
+        match block!(self.wait()) {
+            Ok(_) => {},
+            Err(x) => unreachable(x),
+        }
     }
 }
--- a/nrf52-hal-common/src/timer.rs
+++ b/nrf52-hal-common/src/timer.rs
@@ -5,7 +5,7 @@
 
 use core::ops::Deref;
 
-use nb;
+use nb::{self, block};
 use crate::target::{
     timer0,
     Interrupt,
@@ -145,5 +145,10 @@ impl<T> Timer<T> where T: TimerExt {
         self.0.events_compare[0].write(|w| w);
 
         Ok(())
+    }
+
+    pub fn delay(&mut self, cycles: u32) {
+        self.start(cycles);
+        block!(self.wait()).unwrap();
     }
 }
--- a/nrf52-hal-common/src/uarte.rs
+++ b/nrf52-hal-common/src/uarte.rs
@@ -192,7 +192,15 @@ impl<T> Uarte<T> where T: UarteExt {
     /// Read via UARTE
     ///
     /// This method fills all bytes in `rx_buffer`, and blocks
-    /// until the buffer is full.
+    /// until the buffer is full or the timeout expires, whichever
+    /// comes first.
+    ///
+    /// If the timeout occurs, an `Error::Timeout(n)` will be returned,
+    /// where `n` is the number of bytes read successfully.
+    ///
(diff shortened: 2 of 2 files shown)


nrf52-hal-common/src/timer.rs
```diff
@@ -146,4 +146,9 @@ impl<T> Timer<T> where T: TimerExt {
 
         Ok(())
     }
+
+    pub fn delay(&mut self, cycles: u32) {
+        self.start(cycles);
+        block!(self.wait()).unwrap();
```
jamesmunns: TODO: Should this result be returned?

nrf52-hal-common/src/uarte.rs
```diff
@@ -162,6 +164,130 @@ impl<T> Uarte<T> where T: UarteExt {
...
+        if self.0.rxd.amount.read().bits() != rx_buffer.len() as u32 {
+            return Err(Error::Receive);
+        }
+
+        Ok(())
+    }
+
+    /// Read via UARTE
+    ///
+    /// This method fills all bytes in `rx_buffer`, and blocks
```
jamesmunns: TODO: Fix this comment

nrf52-hal-common/src/uarte.rs
```diff
@@ -162,6 +164,130 @@ impl<T> Uarte<T> where T: UarteExt {
...
+        }
+
+        // Cleanup, even in the error case
+        self.finalize_read();
+
+        let bytes_read = self.0.rxd.amount.read().bits() as usize;
+
+        if timeout_occured {
+            return Err(Error::Timeout(bytes_read));
+        }
```
jamesmunns: TODO: Do we need to explicitly stop/clear the timer here? Probably only if interrupts are enabled...

nrf52-hal-common/src/timer.rs
Yatekii: I don't think there is any use to this, no?

jamesmunns: I think the unwrap will only occur if something other than Error::WouldBlock is returned. That isn't possible today, but might be later? I'm not too familiar with the `nb` crate.

Yatekii: Well I guess that would need be adressed when there changes anything about this. It makes no sense to have the user always write

`let _  = delay(42);`

imo.

nrf52-hal-common/src/uarte.rs
```diff
@@ -163,6 +164,130 @@ impl<T> Uarte<T> where T: UarteExt {
...
+        self.start_read(rx_buffer)?;
+
+        // Start the timeout timer
+        timer.start(cycles);
+
+        // Wait for transmission to end
+        let mut event_complete = false;
+        let mut timeout_occured = false;
+
+        loop {
```
hannobraun: I think this is fine, but we should probably document that this method involves busy waiting.

hannobraun:
@jamesmunns 
> I'd be interested to hear your opinion on this approach for receiving with a fixed timeout.

Sorry, forgot to answer this in my review comment. I think this is okay. I'm not exactly thrilled about the busy waiting, but I think it's fine for now. In fact, it might be preferable to dealing with an interrupt, as things are currently.

I think we should revisit this once we've reached the magical land of ownable interrupts, or when it becomes clear that's not going to happen :)

jamesmunns:
@hannobraun addressed all open topics. In general, I'd like the "support longer than 255 DMA bytes" issue to be tackled in a separate PR.

Let me know what you think

jamesmunns:
@hannobraun reping for `r+`/merge

hannobraun:
Looks good to me. Thank you, @jamesmunns!

PR #44, opened 2019-01-24
Properly size the memory variants of the nrf53832 

jamesmunns:
as per PS 1.4 table 135

### DIFF
--- a/nrf52832-hal/Cargo.toml
+++ b/nrf52832-hal/Cargo.toml
@@ -40,5 +40,7 @@ version = "0.2.1"
 [features]
 doc = []
 rt = ["nrf52832-pac/rt"]
-default = ["rt"]
+xxAA-package = []
+xxAB-package = []
+default = ["rt", "xxAA-package"]
 
--- a/nrf52832-hal/Cargo.toml
+++ b/nrf52832-hal/Cargo.toml
@@ -42,5 +42,8 @@ doc = []
 rt = ["nrf52832-pac/rt"]
 xxAA-package = []
 xxAB-package = []
-default = ["rt", "xxAA-package"]
+
+# Note: We use the xxAB package because it has the least amount of available resources.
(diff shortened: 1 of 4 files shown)


nrf52832-hal/Cargo.toml
```diff
@@ -40,5 +40,7 @@ version = "0.2.1"
 [features]
 doc = []
 rt = ["nrf52832-pac/rt"]
-default = ["rt"]
+xxAA-package = []
+xxAB-package = []
+default = ["rt", "xxAA-package"]
```
hannobraun: I think this should default to `xxAB-package`. That has the smaller memory, which makes it the safe choice.

jamesmunns: Hm, I see what you mean, however the xxAA package is *much* more widely used, particularly on development boards. This would put the inconvenience on the majority of users.

If you feel strongly about this, I can make the change, however I would prefer not to.

jamesmunns:
@hannobraun `memory.x` was moved to `memory_xxAA.x`.

jamesmunns:
@hannobraun reping for `r+`/merge

nrf52832-hal/Cargo.toml
jamesmunns: I still feel the same, but @hannobraun's opinion and @simonsso's :+1: have convinced me :)

Change made.

hannobraun: I was just about to say that I don't feel strongly enough to overrule you. Better merge this fast, before you change your mind again ;-)

hannobraun:
@jamesmunns 
> `memory.x` was moved to `memory_xxAA.x`.

Ah, thanks, I misread the diff.

hannobraun:
Looks good. Thank you, @jamesmunns!

PR #45, opened 2019-01-28
Option encapsulation for SPI Pins

simonsso:
In some scenarios it is possible to setup unidirectional SPI
communication. When only one way communication is needed there
is no need to connect a GPIO for the unused MISO/MOSI pin.

Encoding pins in an Option<Pin> will support this use case


jamesmunns:
LGTM! Thank you!

PR #46, opened 2019-01-29
Allow larger EasyDMA buffer sizes on nrf52840

### LINKED ISSUE #17: Allow larger SPI buffer sizes on nrf52840 in nrf52-hal-common

simonsso:
Expose const function in nrf52-hal-common which per platform
return limit of EasyDMA.

Closes #17

### DIFF
--- a/nrf52-hal-common/src/lib.rs
+++ b/nrf52-hal-common/src/lib.rs
@@ -35,7 +35,7 @@ pub mod prelude {
 #[cfg(feature = "52832")]
 pub mod target_constants {
     // NRF52832 8 bits1..0xFF
-    pub const EASY_DMA_SIZE:usize = 255;
+    pub const EASY_DMA_SIZE: usize = 255;
 }
     // NRF52840 16 bits 1..0xFFFF
 #[cfg(feature = "52840")]
--- a/nrf52-hal-common/src/lib.rs
+++ b/nrf52-hal-common/src/lib.rs
@@ -30,8 +30,15 @@ pub mod prelude {
     pub use crate::twim::TwimExt;
     pub use crate::uarte::UarteExt;
 }
-
-
+/// Length of Nordic EasyDMA differs for MCUs
+const fn easy_dma_size() -> usize {
+    // NRF52832 8 bits1..0xFF
+    #[cfg(feature = "52832")]
+    {255}
+    // NRF52840 16 bits 1..0xFFFF
+    #[cfg(feature = "52840")]
+    {65535}
+}
 pub use crate::clocks::Clocks;
 pub use crate::delay::Delay;
 pub use crate::spim::Spim;
--- a/nrf52-hal-common/src/lib.rs
+++ b/nrf52-hal-common/src/lib.rs
@@ -40,7 +40,7 @@ pub mod target_constants {
     // NRF52840 16 bits 1..0xFFFF
 #[cfg(feature = "52840")]
 pub mod target_constants {
-    pub const EASY_DMA_SIZE:usize = 65535;
+    pub const EASY_DMA_SIZE: usize = 65535;
 }
 
 
--- a/nrf52-hal-common/src/lib.rs
+++ b/nrf52-hal-common/src/lib.rs
@@ -30,15 +30,20 @@ pub mod prelude {
     pub use crate::twim::TwimExt;
     pub use crate::uarte::UarteExt;
 }
+
 /// Length of Nordic EasyDMA differs for MCUs
-const fn easy_dma_size() -> usize {
+#[cfg(feature = "52832")]
+pub mod target_constants{
     // NRF52832 8 bits1..0xFF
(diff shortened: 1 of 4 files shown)


Yatekii:
Why did you use a const fn instead of just a const variable?
Looks good to me on all other counts

simonsso:
The _const fn_ evaluates to a constant at compile time, and I thought a get-value-function call looked clearer in the code, but it is a matter of taste.

The constant is also only in one place as it is tied to the EasyDMA block in the MCU more than to the individual function blocks.

Yatekii:
> The const fn evaluates to a constant at compile time, and I thought a get-value-function call looked clearer in the code, but it is a matter of taste.

I am aware how it works, I just don't see the use of it. It's more confusing than not imo. Also it's not named get_ so it's not really a getter ;)

> The constant is also only in one place as it is tied to the EasyDMA block in the MCU more than to the individual function blocks.

How do you mean? I am not sure whether the MAXCNT is really an EasyDMA thing or a per-peripheral thing. But since the value is the same for UART/TWI/SPI, I am ok with having it the same constant for all 3 peripherals.

nrf52-hal-common/src/lib.rs
```diff
@@ -31,6 +31,18 @@ pub mod prelude {
     pub use crate::uarte::UarteExt;
 }
 
+/// Length of Nordic EasyDMA differs for MCUs
+#[cfg(feature = "52832")]
+pub mod target_constants{
```
jamesmunns:
```suggestion
pub mod target_constants {
```

Formatting suggestion

nrf52-hal-common/src/lib.rs
```diff
@@ -31,6 +31,18 @@ pub mod prelude {
...
+    // NRF52832 8 bits1..0xFF
+    pub const EASY_DMA_SIZE:usize = 255;
```
jamesmunns:
```suggestion
    pub const EASY_DMA_SIZE: usize = 255;
```

nrf52-hal-common/src/lib.rs
```diff
@@ -31,6 +31,18 @@ pub mod prelude {
...
+}
+    // NRF52840 16 bits 1..0xFFFF
+#[cfg(feature = "52840")]
+pub mod target_constants{
```
jamesmunns:
```suggestion
pub mod target_constants {
```

nrf52-hal-common/src/lib.rs
```diff
@@ -31,6 +31,18 @@ pub mod prelude {
...
+    pub const EASY_DMA_SIZE:usize = 65535;
```
jamesmunns:
```suggestion
    pub const EASY_DMA_SIZE: usize = 65535;
```

nrf52-hal-common/src/spim.rs
```diff
@@ -226,16 +224,14 @@ impl<T> Spim<T> where T: SpimExt {
     /// This method uses the provided chip select pin to initiate the
     /// transaction, then transmits all bytes in `tx_buffer`.
     ///
-    /// The buffer must have a length of at most 255 bytes.
+
```
```suggestion
```

nrf52-hal-common/src/twim.rs
```diff
@@ -263,13 +261,11 @@ impl<T> Twim<T> where T: TwimExt {
     )
         -> Result<(), Error>
     {
-        // This is overly restrictive. See:
-        // https://github.com/nrf-rs/nrf52-hal/issues/17
-        if wr_buffer.len() > u8::max_value() as usize {
+        if wr_buffer.len() > EASY_DMA_SIZE{
```
jamesmunns:
```suggestion
        if wr_buffer.len() > EASY_DMA_SIZE {
```

nrf52-hal-common/src/twim.rs
```diff
@@ -263,13 +261,11 @@ impl<T> Twim<T> where T: TwimExt {
...
             return Err(Error::BufferTooLong);
         }
 
-        if rd_buffer.len() > u8::max_value() as usize {
+        if rd_buffer.len() > EASY_DMA_SIZE{
```
jamesmunns:
```suggestion
        if rd_buffer.len() > EASY_DMA_SIZE {
```

jamesmunns:
Hey @simonsso, I made a couple formatting suggestions, I'm happy to merge once those are fixed!

jamesmunns:
@simonsso looks good to me! Ready to merge, or do you have any other changes planned?

simonsso:
Lets merge this first and I will move on to the dependent pull requests based on this.

Issue #37, opened 2018-12-24
EasyDMA doesn't read from Flash


jamesmunns:
In general, EasyDMA does not work with items in Flash, e.g. a `&'static str`. From the datasheet:

> EasyDMA is an easy-to-use direct memory access module that some peripherals implement to gain direct
access to Data RAM.
> The EasyDMA is an AHB bus master similar to the CPU and it is connected to the AHB multilayer
interconnect for direct access to the Data RAM. The EasyDMA is not able to access the Flash.

I did this accidentally, and did not receive any visible error when sending via UARTE. I think there would be 3 ways to address this:

1. if there is some way to detect this error condition and return an Err
2. `debug_assert` that the slice ptr is in the RAM region
3. Do an explicit check, roughly `if !( RAM_LOWER <= buf.as_ptr() <= RAM_UPPER ) { return Err(...); }`

Yatekii:
Experienced the same just today.
I am not sure what we do best (Error/debug_assert!). Since we already go by errors I think we can keep them even tho debug_assert!s would be better also for buffer length checking etc. I think.

simonsso:
4th option: **Copy to ram**

When implementing embedded_hal traits where one do not have control over the external calling code it could be necessary to copy from flash to ram and then EasyDMA (which of course feels wrong)

jamesmunns:
@simonsso true, copying to RAM is always an option, however I'd like to avoid excess copies when not needed. I think the approach you're taking in https://github.com/simonsso/nrf52-hal/commit/cee4745e399d50b1122b0b32e1baefa486197df0 is definitely interesting!

Yatekii:
I like the fact that your implementation throws an error when the `do_spi_dma_transfer` function detects an out of RAM bounds.

What I do not really like is the implicit copy. It's nice if you just expect the SPI bytes to be pumped out. But for some cases (for example SPI abuse for driving LEDs) this can kill you. At least on the nRF52832 there was a very tiny time-gap between SPI transactions, which really can kill you in terms of timings (we had some major issues, even when directly chaining the transactions via the PPI on-chip bus).

So I would rather have it such that we have a separate method to assemble that buffer which you can use in case of an erroneous transaction due to RAM boundaries. You could even chain that easily via results like this:

```
spi.write(&buf).or_else(|e| if let Error::DMABufferNotInDataMemory = e { spi.write(&copy_to_ram(&buf)) });
```

If you really always want to copy if things go south.

[...]

Issue #11, opened 2018-09-12
Consider supporting extended DMA transfers


hannobraun:
This issue was first opened in the [old nrf52-hal repository](https://github.com/jamesmunns/nrf52-hal). Original discussion: https://github.com/jamesmunns/nrf52-hal/issues/14

---

The nRF52832's DMA channels are limited to buffers with a maximum length of 255 bytes. This is very limiting. For example, in the [DW1000 driver](https://github.com/braun-robotics/rust-dw1000), I need to support registers that are much longer than that (although there are ways to read/write subsets).

There is a way to use multiple consecutive buffers to circumvent this limitation. The product specification alludes to this (see section 10.1), but doesn't really explain how this works. I found the following discussions that provide a clearer picture:
- https://devzone.nordicsemi.com/f/nordic-q-a/18638/easydma-array-list
- https://devzone.nordicsemi.com/f/nordic-q-a/19891/transfer-more-than-255-bytes-with-spi-master-on-nrf52

Since this technique requires additional system resources (PPI, TIMER) and applies to all DMA channels, I think it's probably best to provide this as a separate layer that builds on top of DMA users like SPIM.

jamesmunns:
Hey @hannobraun, do you think a solution like https://github.com/nrf-rs/nrf52-hal/pull/40 would work? My idea would be roughly as follows:

* Continue offering a "blocking" implementation of each peripheral, essentially what we have now
* Offer an "async" version of each peripheral which does the following:
    * Provides the interrupt handler, transparent to the user
    * uses a `static BBQueue` (one for each peripheral and each direction)
    * uses something like [this demo](https://github.com/jamesmunns/mcf-config-demo) to allow the user to enable/configure the buffer size, at least until it is possible to have the `UarteAsync` instance totally own the BBQueue in a way that makes sense

hannobraun:
@jamesmunns 
I left some comments in your pull request. Whether it is an alternative to the approach suggested in this issue, I don't know. My impression is that this approach would enable the blocking API to support larger buffers with little hassle (from the user's perspective). You're suggesting a non-blocking API, which seems somewhat orthogonal.

Yatekii:
I used exactly that feature of the nRF52832 in C already and it's really not that much sorcery.
[...]

PR #48, opened 2019-01-29
Implemenet embedded_hal traits for SPI

simonsso:
Implement embedded_hal traits and refactor SPIM

Implements spi::blocking::Transfer trait (as first posted in #39 ) with refactoring of other methods to use shared helper function.

Implements spi::blocking::write. - This function is could be called with buffer in flash, copy to ram function is also implemented.

implements #10
implements #37 for SPIM
partially solves #11

### DIFF
--- a/nrf52-hal-common/src/spim.rs
+++ b/nrf52-hal-common/src/spim.rs
@@ -64,15 +64,12 @@ impl<T> embedded_hal::blocking::spi::Transfer<u8> for Spim<T> where T: SpimExt
     fn transfer<'w>(&mut self, words: &'w mut [u8]) -> Result<&'w [u8], Error> {
         let mut offset:usize = 0;
         while offset < words.len() {
-            let datalen = min(EASY_DMA_SIZE, words.len() - offset);
-            let dataptr = offset + (words.as_ptr() as usize);
-            offset += datalen;
+            let data_len = min(EASY_DMA_SIZE, words.len() - offset);
+            let data_ptr = offset + (words.as_ptr() as usize);
+            offset += data_len;
 
-            self.do_spi_dma_transfer(dataptr as u32,datalen as u32,dataptr as u32,datalen as u32,|_|{})?;
+            self.do_spi_dma_transfer(data_ptr as u32,data_len as u32,data_ptr as u32,data_len as u32,|_|{})?;
 
-            // Conservative compiler fence to prevent optimizations that do not
-            // take in to account DMA
-            compiler_fence(SeqCst);
         }
         Ok(words)
     }
@@ -88,16 +85,12 @@ impl<T> embedded_hal::blocking::spi::Write<u8> for Spim<T> where T: SpimExt
         if SRAM_LOWER <= p && p < SRAM_UPPER {
             let mut offset:usize = 0;
             while offset < words.len() {
-                let datalen = min(EASY_DMA_SIZE, words.len()  - offset);
-                let dataptr = offset + (words.as_ptr() as usize);
-                offset += datalen;
+                let data_len = min(EASY_DMA_SIZE, words.len()  - offset);
+                let data_ptr = offset + (words.as_ptr() as usize);
+                offset += data_len;
 
                 // setup spi dma tx buffer and 0 for read buffer length
-                self.do_spi_dma_transfer(dataptr as u32,datalen as u32,0,0,|_|{})?;
-
-                // Conservative compiler fence to prevent optimizations that do not
-                // take in to account DMA
-                compiler_fence(SeqCst);
+                self.do_spi_dma_transfer(data_ptr as u32,data_len as u32,0,0,|_|{})?;
             }
         } else {
             // Force copy from flash mode.
@@ -105,17 +98,13 @@ impl<T> embedded_hal::blocking::spi::Write<u8> for Spim<T> where T: SpimExt
             let mut buffer:[u8;FORCE_COPY_BUFFER_SIZE] = [0;FORCE_COPY_BUFFER_SIZE];
             let mut offset:usize = 0;
             while offset < words.len() {
-                let datalen = min(blocksize, words.len()  - offset);
-                for i in 0..datalen{
+                let data_len = min(blocksize, words.len()  - offset);
+                for i in 0..data_len{
                     buffer[i] = words[offset+i];
                 }
                 offset += blocksize;
                 // setup spi dma tx buffer and 0 for read buffer length
-                self.do_spi_dma_transfer(buffer.as_ptr() as u32,datalen as u32,0,0,|_|{})?;
-
-                // Conservative compiler fence to prevent optimizations that do not
-                // take in to account DMA
-                compiler_fence(SeqCst);
+                self.do_spi_dma_transfer(buffer.as_ptr() as u32,data_len as u32,0,0,|_|{})?;
             }
         }
         Ok(())
--- a/nrf52-hal-common/src/spim.rs
+++ b/nrf52-hal-common/src/spim.rs
@@ -66,7 +66,7 @@ impl<T> embedded_hal::blocking::spi::Transfer<u8> for Spim<T> where T: SpimExt
         while offset < words.len() {
             let datalen = min(EASY_DMA_SIZE, words.len() - offset);
             let dataptr = offset + (words.as_ptr() as usize);
-            offset += EASY_DMA_SIZE;
+            offset += datalen;
 
             self.do_spi_dma_transfer(dataptr as u32,datalen as u32,dataptr as u32,datalen as u32,|_|{})?;
 
@@ -90,7 +90,7 @@ impl<T> embedded_hal::blocking::spi::Write<u8> for Spim<T> where T: SpimExt
             while offset < words.len() {
                 let datalen = min(EASY_DMA_SIZE, words.len()  - offset);
                 let dataptr = offset + (words.as_ptr() as usize);
-                offset += EASY_DMA_SIZE;
+                offset += datalen;
 
                 // setup spi dma tx buffer and 0 for read buffer length
                 self.do_spi_dma_transfer(dataptr as u32,datalen as u32,0,0,|_|{})?;
--- a/nrf52-hal-common/src/spim.rs
+++ b/nrf52-hal-common/src/spim.rs
@@ -187,7 +187,7 @@ impl<T> Spim<T> where T: SpimExt {
         // Check If buffer is in data RAM, compiler sometimes put static data
         // in flash this area is not accessable by EasyDMA
         if (tx_data_ptr as usize) < SRAM_LOWER || tx_data_ptr as usize >= SRAM_UPPER {
-            return Err(Error::DMABufferInUnaccessableMemory)
+            return Err(Error::DMABufferNotInDataMemory)
         }
         // Conservative compiler fence to prevent optimizations that do not
         // take in to account actions by DMA. The fence has been placed here,
@@ -281,7 +281,7 @@ impl<T> Spim<T> where T: SpimExt {
         }
 
         self.do_spi_dma_transfer(
-        // Set up the DMA write
+            // Set up the DMA write
             // We're giving the register a pointer to the stack. Since we're
             // waiting for the SPI transaction to end before this stack pointer
             // becomes invalid, there's nothing wrong here.
@@ -302,20 +302,21 @@ impl<T> Spim<T> where T: SpimExt {
             // The MAXCNT field is thus at least 8 bits wide and accepts the full
             // range of values that fit in a `u8`.
 
-            //tx_len:
+            // tx_len:
             tx_buffer.len() as _,
 
             // Set up the DMA read
             // This is safe for the same reasons that writing to TXD.PTR is
             // safe. Please refer to the explanation there.
 
-            //rx_data_ptr:
+            // rx_data_ptr:
             rx_buffer.as_mut_ptr() as u32,
             // This is safe for the same reasons that writing to TXD.MAXCNT is
             // safe. Please refer to the explanation there.
 
-            //rx_len:
+            // rx_len:
             rx_buffer.len() as _,
+            // chip select callback
             |cs|{if cs {chip_select.set_low()} else {chip_select.set_high()} }
         )
     }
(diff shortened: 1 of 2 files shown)


nrf52-hal-common/src/spim.rs
```diff
@@ -56,6 +57,71 @@ impl_spim_ext!(
...
+impl<T> embedded_hal::blocking::spi::Transfer<u8> for Spim<T> where T: SpimExt
+{
+   type Error = Error;
+
+    fn transfer<'w>(&mut self, words: &'w mut [u8]) -> Result<&'w [u8], Error> {
+        let mut offset:usize = 0;
+        while offset < words.len() {
+            let datalen = min(EASY_DMA_SIZE, words.len() - offset);
+            let dataptr = offset + (words.as_ptr() as usize);
+            offset += EASY_DMA_SIZE;
```
hannobraun: This line is correct (I think), but it made me pause at first. Maybe using `datalen` would be a bit clearer.

```suggestion
            offset += datalen;
```

Just a suggestion, feel free to ignore.

nrf52-hal-common/src/spim.rs
```diff
@@ -56,6 +57,71 @@ impl_spim_ext!(
...
 pub struct Spim<T>(T);
 
+impl<T> embedded_hal::blocking::spi::Transfer<u8> for Spim<T> where T: SpimExt
+{
+   type Error = Error;
+
+    fn transfer<'w>(&mut self, words: &'w mut [u8]) -> Result<&'w [u8], Error> {
+        let mut offset:usize = 0;
+        while offset < words.len() {
+            let datalen = min(EASY_DMA_SIZE, words.len() - offset);
```
hannobraun: Minor nit-pick: "data" and "len" are two words, so `data_len` would be more idiomatic, I believe. Not a big deal, feel free to leave it.

nrf52-hal-common/src/spim.rs
```diff
@@ -56,6 +57,71 @@ impl_spim_ext!(
...
+                let datalen = min(EASY_DMA_SIZE, words.len()  - offset);
+                let dataptr = offset + (words.as_ptr() as usize);
+                offset += EASY_DMA_SIZE;
+
+                // setup spi dma tx buffer and 0 for read buffer length
+                self.do_spi_dma_transfer(dataptr as u32,datalen as u32,0,0,|_|{})?;
+
+                // Conservative compiler fence to prevent optimizations that do not
+                // take in to account DMA
+                compiler_fence(SeqCst);
```
hannobraun: Same as above: Aren't the compiler fences in `do_spi_dma_transfer` enough?

nrf52-hal-common/src/spim.rs
```diff
@@ -56,6 +57,71 @@ impl_spim_ext!(
...
+            }
+        } else {
+            // Force copy from flash mode.
+            let blocksize = min(EASY_DMA_SIZE,FORCE_COPY_BUFFER_SIZE); 
+            let mut buffer:[u8;FORCE_COPY_BUFFER_SIZE] = [0;FORCE_COPY_BUFFER_SIZE];
+            let mut offset:usize = 0;
+            while offset < words.len() {
+                let datalen = min(blocksize, words.len()  - offset);
```
hannobraun: Same as above: Minor nit-pick, `data_len` would look nicer, feel free to ignore.

nrf52-hal-common/src/spim.rs
```diff
@@ -56,6 +57,71 @@ impl_spim_ext!(
...
+                for i in 0..datalen{
+                    buffer[i] = words[offset+i];
+                }
+                offset += blocksize;
+                // setup spi dma tx buffer and 0 for read buffer length
+                self.do_spi_dma_transfer(buffer.as_ptr() as u32,datalen as u32,0,0,|_|{})?;
+
+                // Conservative compiler fence to prevent optimizations that do not
+                // take in to account DMA
+                compiler_fence(SeqCst);
```
hannobraun: Same as above: Aren't the compiler fences in `do_spi_dma_transfer` enough?

nrf52-hal-common/src/spim.rs
```diff
@@ -56,6 +57,71 @@ impl_spim_ext!(
...
+        while offset < words.len() {
+            let datalen = min(EASY_DMA_SIZE, words.len() - offset);
+            let dataptr = offset + (words.as_ptr() as usize);
+            offset += EASY_DMA_SIZE;
+
+            self.do_spi_dma_transfer(dataptr as u32,datalen as u32,dataptr as u32,datalen as u32,|_|{})?;
+
+            // Conservative compiler fence to prevent optimizations that do not
+            // take in to account DMA
+            compiler_fence(SeqCst);
```
hannobraun: Why is this compiler fence here? The ones in `do_spi_dma_transfer` should be enough, no?

nrf52-hal-common/src/spim.rs
```diff
@@ -56,6 +57,71 @@ impl_spim_ext!(
...
+   type Error = Error;
+
+    fn write<'w>(&mut self, words: &'w [u8]) -> Result<(), Error> {
+        let p = words.as_ptr() as usize;
+        // Mask on segment where Data RAM is located on nrf52840 and nrf52832
+        // Upper limit is choosen to entire area where DataRam can be placed
+        if SRAM_LOWER <= p && p < SRAM_UPPER {
+            let mut offset:usize = 0;
+            while offset < words.len() {
+                let datalen = min(EASY_DMA_SIZE, words.len()  - offset);
```
hannobraun: Same as above: Minor nit-pick, `data_len` would look nicer, feel free to ignore.

nrf52-hal-common/src/spim.rs
```diff
@@ -56,6 +57,71 @@ impl_spim_ext!(
 /// - The over-read character is hardcoded to `0`.
 pub struct Spim<T>(T);
 
+impl<T> embedded_hal::blocking::spi::Transfer<u8> for Spim<T> where T: SpimExt
+{
+   type Error = Error;
+
+    fn transfer<'w>(&mut self, words: &'w mut [u8]) -> Result<&'w [u8], Error> {
```
hannobraun: Why doesn't this method copy code to SRAM, if needed, just like `write` does?

nrf52-hal-common/src/spim.rs
```diff
@@ -201,100 +252,124 @@ impl<T> Spim<T> where T: SpimExt {
...
-    /// transaction, then transmits all bytes in `tx_buffer`.
+    /// This method implements a complete read transaction, which consists of
+    /// the master transmitting what it wishes to read, and the slave responding
+    /// with the requested data.
     ///
-    pub fn write(&mut self,
+    /// Uses the provided chip select pin to initiate the transaction. Transmits
+    /// all bytes in `tx_buffer`, then receives bytes until `rx_buffer` is full.
+    /// Both buffer must have a length of at most 255 bytes.
+    pub fn read(&mut self,
```
hannobraun: Same question as above: Why doesn't this method copy code to SRAM, if needed, just like `write` does?

nrf52-hal-common/src/spim.rs
```diff
@@ -56,6 +57,71 @@ impl_spim_ext!(
...
+    fn write<'w>(&mut self, words: &'w [u8]) -> Result<(), Error> {
+        let p = words.as_ptr() as usize;
+        // Mask on segment where Data RAM is located on nrf52840 and nrf52832
+        // Upper limit is choosen to entire area where DataRam can be placed
+        if SRAM_LOWER <= p && p < SRAM_UPPER {
+            let mut offset:usize = 0;
+            while offset < words.len() {
+                let datalen = min(EASY_DMA_SIZE, words.len()  - offset);
+                let dataptr = offset + (words.as_ptr() as usize);
+                offset += EASY_DMA_SIZE;
```
hannobraun: Same suggestion as above, feel free to ignore.

```suggestion
                offset += datalen;
```

nrf52-hal-common/src/spim.rs
```diff
@@ -119,79 +185,64 @@ impl<T> Spim<T> where T: SpimExt {
...
+            rx_len:u32,
+            mut cs_n:CSFun
+            ) -> Result<(), Error>
+            where CSFun: FnMut(bool)
     {
-        if tx_buffer.len() > EASY_DMA_SIZE {
-            return Err(Error::TxBufferTooLong);
+        // Check If buffer is in data RAM, compiler sometimes put static data
+        // in flash this area is not accessable by EasyDMA
+        if (tx_data_ptr as usize) < SRAM_LOWER || tx_data_ptr as usize >= SRAM_UPPER {
```
hannobraun: If all the methods calling this one made sure to copy to SRAM, if needed, we could replace this with an `assert!` and remove the new error variants again.

nrf52-hal-common/src/spim.rs
```diff
@@ -322,6 +397,7 @@ pub struct Pins {
 pub enum Error {
     TxBufferTooLong,
     RxBufferTooLong,
+    DMABufferInUnaccessableMemory,
```
hannobraun: Typo: unaccess**i**ble

Although Google suggests that **in**accessible is the more common word, by far.

nrf52-hal-common/src/spim.rs
```diff
@@ -119,79 +185,64 @@ impl<T> Spim<T> where T: SpimExt {
...
-        chip_select.set_low();
         self.0.tasks_start.write(|w|
             // `1` is a valid value to write to task registers.
             unsafe { w.bits(1) }
         );
 
+        // Conservative compiler fence to prevent optimizations that do not
+        // take in to account actions by DMA. The fence has been placed here,
+        // after all possible DMA actions have completed
+        compiler_fence(SeqCst);
```
hannobraun: What's the reason for this compiler fence? I don't think we need it (but I'm not sure).

nrf52-hal-common/src/spim.rs
simonsso: Keeping style of the code is good,I will change.

nrf52-hal-common/src/spim.rs
simonsso: I intended this check to be propagated to all the outer (API) functions calling this. I placed it in the  inner function so the code would be shared.

There are still cases where this code can be called with buffer in flash which is not handled, when all cases are handled correctly this could be an assert until then I think it is an error.

nrf52-hal-common/src/spim.rs
simonsso: (bad excuse again )The compiler fences came form the originating code, and I kept them where they landed. Will give it a try and see which could be removed.

nrf52-hal-common/src/spim.rs
simonsso: I have tried not modify the behavior of the existing read/write functions. They are only refactored with the shared code. The write function which implements the copy is a new implementation of the embedde_hal write trait

nrf52-hal-common/src/spim.rs
simonsso: The buffer is mutable and will always be placed in data memory by the compiler, so there will never be the same need for a copy

nrf52-hal-common/src/spim.rs
simonsso: I have not been able to observe this error, where the compiler removed too much. I removed the one I dared to remove.

nrf52-hal-common/src/spim.rs
hannobraun: Ah, makes sense! Adding a comment that states just that would be great.

nrf52-hal-common/src/spim.rs
hannobraun: Makes sense. I'm happy to merge this PR without these modifications, but we should remember to open an issue about it, if we do that.

nrf52-hal-common/src/spim.rs
hannobraun: If in doubt, leaving compiler fences in is a good idea, I think. The worst they can do is prevent some optimizations.

The perfect scenario for me would be, if someone took a detailed look and proclaimed, with great certainty, which compiler fences are required and which aren't. As long as that doesn't happen, I'm happy to merge code that has an additional compiler fence here or there :-)

hannobraun:
Thanks for the changes! Haven't had a chance to take a look yet, but I've added it to my task list. Feel free to ping me, if it takes too long.

In the meantime, more reviews are welcome!

simonsso:
> Thanks for the PR, @simonsso! I left a lot of comments, but I do like this a lot :-)
> 

Good review comments makes the code better and makes programming fun.

> Have you tested this in real-use scenarios? I'd like to see confirmation that EasyDMA can use the same buffer for reading/writing.
> 

Yes I have tested both traits on hardware. The same buffer will always work: because bytes are clocked out and in from SPI block with the same clock; the data must be moved from RAM before being shifted out and the read data will not be written to RAM until it have been fully received.

A side note: a common behavior for SPI slaves is to have one instruction for both read and write. Reading a register will at the same time write whatever was in the buffer, but then the same buffer can be used one more time to restore the data. Sometimes confusing but very simple.

nrf52-hal-common/src/spim.rs
```diff
@@ -201,100 +241,123 @@ impl<T> Spim<T> where T: SpimExt {
...
-        compiler_fence(SeqCst);
+    /// Write to an SPI slave
+    ///
+    /// This method uses the provided chip select pin to initiate the
+    /// transaction, then transmits all bytes in `tx_buffer`.
+    ///
 
-        // End SPI transaction
-        chip_select.set_high();
+    pub fn write(&mut self,
```
simonsso: This  old method now have the same name as write in the embedded_hal trait, it causes some problem for a user of the library.

It is still works -- it is possible to choose which function to use and the number of parameters are not the same so the user will notice.

hannobraun: I'm inclined to accept this for now, so we can get this PR merged. But yes, it would be better if there weren't any name collisions.

jamesmunns:
Looks great to me! Thanks @simonsso!

jamesmunns:
Also thank you @hannobraun for reviewing :D

PR #50, opened 2019-02-05
UART and TWIM: Implement large buffer splitting and  check for readonly buffers in flash

simonsso:
Same type of changes as done for SPI.

I have not split up the dma transfer for: **pub fn write_then_read()**
it is not clear to me what assumptions can be made about read and write - for instance if a write and read always completes in sync or if all writes shall be done before data starts to be revived.

There have been no attempts to verify this change -- other than it is almost the same as I did for SPI.


simonsso:
@jamesmunns / @hannobraun / @Yatekii Should we try to do something about this change? 

This patch includes the fail-if-DMA is not in RAM check and replaces the 255 limit on easy DMA by splitting it up in several if there is a limit.

hannobraun:
Sorry, I don't have time to take a closer look right now. I've added it to my list.

nrf52-hal-common/src/twim.rs
```diff
@@ -110,151 +112,151 @@ impl<T> Twim<T> where T: TwimExt {
...
     pub fn write(&mut self,
         address: u8,
         buffer:  &[u8],
     )
         -> Result<(), Error>
     {
-
-        if buffer.len() > EASY_DMA_SIZE {
-            return Err(Error::TxBufferTooLong);
+        if SRAM_LOWER <= ( buffer.as_ptr() as usize ) && (buffer.as_ptr() as usize) < SRAM_UPPER {
```
somemetricprefix: Instead of using the address of the very first element, this could check if the end of the buffer is still at an address <SRAM_UPPER.

Yatekii: Still, good catch. just adding the `buffer.len()` should work.

Yatekii:
@simonsso I tried to speed up discussions about the implicit copy and batching.
We did also think about possible compiletime solutions.

Maybe it should be merged for now and adapted later on (even tho I think we'll never adapt it then :P).
If @jamesmunns  and @hannobraun agree, I am ok with merging this for now if you fix the minor issue with bondary checks pointed out.

What do you think about this @jamesmunns @hannobraun? Appart from the ongoing issue about implicity, this seems to be sound otherwise (except for the minor issue pointed out by @somemetricprefix which can be fixed easily.).

nrf52-hal-common/src/twim.rs
Yatekii: As pointed out in https://github.com/nrf-rs/nrf52-hal/pull/50/files/968be9f40c1b04f3b7715a7f947a20707f8580b4#r262476494 this should be fixed.

nrf52-hal-common/src/twim.rs
```diff
@@ -276,6 +278,9 @@ impl<T> Twim<T> where T: TwimExt {
         if rd_buffer.len() > EASY_DMA_SIZE {
             return Err(Error::RxBufferTooLong);
         }
+        if SRAM_LOWER <= ( wr_buffer.as_ptr() as usize ) && (wr_buffer.as_ptr() as usize) < SRAM_UPPER {
```
Yatekii: Same issue here.

nrf52-hal-common/src/twim.rs
simonsso: If this is an issue there are other problems with the buffer halfway outside of the RAM, but I makes sense include buffer length for readability.

simonsso:
Close in favor of #74 and #75

PR #64, opened 2019-02-27
feat: SPIM configuration

chocol4te:
This PR allows the configuration of the frequency, mode and over-read character of the SPIM interfaces.

### DIFF
--- a/nrf52-hal-common/src/spim.rs
+++ b/nrf52-hal-common/src/spim.rs
@@ -4,6 +4,7 @@
 use core::ops::Deref;
 use core::sync::atomic::{compiler_fence, Ordering::SeqCst};
 use core::cmp::min;
+pub use embedded_hal::spi::{Mode, Phase, Polarity, MODE_0, MODE_1, MODE_2, MODE_3};
 
 use crate::target::{spim0, SPIM0};
 
@@ -21,15 +22,15 @@ use crate::gpio::{
 };
 
 pub trait SpimExt : Deref<Target=spim0::RegisterBlock> + Sized {
-    fn constrain(self, pins: Pins) -> Spim<Self>;
+    fn constrain(self, pins: Pins, frequency: Frequency, mode: Mode, orc: u8) -> Spim<Self>;
 }
 
 macro_rules! impl_spim_ext {
     ($($spim:ty,)*) => {
         $(
             impl SpimExt for $spim {
-                fn constrain(self, pins: Pins) -> Spim<Self> {
-                    Spim::new(self, pins)
+                fn constrain(self, pins: Pins, frequency: Frequency, mode: Mode, orc: u8) -> Spim<Self> {
+                    Spim::new(self, pins, frequency, mode, orc)
                 }
             }
         )*
@@ -92,7 +93,7 @@ impl<T> embedded_hal::blocking::spi::Write<u8> for Spim<T> where T: SpimExt
             }
         } else {
             // Force copy from flash mode.
-            let blocksize = min(EASY_DMA_SIZE,FORCE_COPY_BUFFER_SIZE); 
+            let blocksize = min(EASY_DMA_SIZE,FORCE_COPY_BUFFER_SIZE);
             let mut buffer:[u8;FORCE_COPY_BUFFER_SIZE] = [0;FORCE_COPY_BUFFER_SIZE];
             let mut offset:usize = 0;
             while offset < words.len() {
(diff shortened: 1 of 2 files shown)


nrf52-hal-common/src/spim.rs
```diff
@@ -155,25 +156,58 @@ impl<T> Spim<T> where T: SpimExt {
...
+        spim.config.write(|w| {
+            // Can't match on `mode` as embedded_hal::spi::Mode only derives PartialEq, not Eq
+            if mode == MODE_0 {
+                w.order().msb_first().cpol().active_high().cpha().leading()
+            } else if mode == MODE_1 {
+                w.order().msb_first().cpol().active_high().cpha().trailing()
+            } else if mode == MODE_2 {
+                w.order().msb_first().cpol().active_low().cpha().leading()
+            } else {
+                w.order().msb_first().cpol().active_low().cpha().trailing()
```
jamesmunns: Maybe panic or return Err here if not mode 0..3?

nrf52-hal-common/src/spim.rs
```diff
@@ -155,25 +156,58 @@ impl<T> Spim<T> where T: SpimExt {
...
-        // Set to SPI mode 0
-        spim.config.write(|w|
-            w
-                .order().msb_first()
-                .cpha().leading()
-                .cpol().active_high()
-        );
+        // Configure mode
+        spim.config.write(|w| {
+            // Can't match on `mode` as embedded_hal::spi::Mode only derives PartialEq, not Eq
```
jamesmunns: That's really annoying. Would you mind opening an issue upstream to add this derive?

nrf52-hal-common/src/spim.rs
chocol4te: There are only 4 modes possible (parity and phase are booleans), so if it is not mode 0 through 2, it must be 3, and there are no other possibilities. I agree this is ugly and unclear, adding the `Eq` derive would mean being able to use a match which would be much nicer.

nrf52-hal-common/src/spim.rs
chocol4te: https://github.com/rust-embedded/embedded-hal/pull/126 My colleague appears to have already done so :)

chocol4te:
Oops, I forgot to fix the examples

nrf52-hal-common/src/spim.rs
Yatekii: I would prefer a match here ;) It's the rusty way to do this :)

jonas-schievink: We can't match on the value since it doesn't implement `Eq`. The next embedded-hal release will fix that, but I'd prefer not to wait for that.

```
error: to use a constant of type `X` in a pattern, `X` must be annotated with `#[derive(PartialEq, Eq)]`
```

Yatekii: Oh, I am very sorry, it's even written down below ... I didn't read on first ... I somehow thought it's an enum.

nrf52-hal-common/src/spim.rs
```diff
@@ -155,25 +156,58 @@ impl<T> Spim<T> where T: SpimExt {
...
+                w.order().msb_first().cpol().active_low().cpha().trailing()
+            }
+        });
 
         // Configure frequency
-        spim.frequency.write(|w|
-            w.frequency().k500() // 500 kHz
-        );
+        spim.frequency.write(|w| match frequency {
+            Frequency::K125 => {
```
Yatekii: I wonder if this huge match is nicer or if values for the enum fields & an `unsafe { bits(enum.field) }` was nicer.

jonas-schievink: This would also match what `Uarte` does, I believe

jamesmunns: Yup, we re-export the PAC type with a nicer name, it would be good to see that here

nrf52-hal-common/src/spim.rs
```suggestion
            // Can't match on `mode` due to embedded-hal, see https://github.com/rust-embedded/embedded-hal/pull/126
```

nrf52-hal-common/src/spim.rs
Yatekii: > we re-export the PAC type

Oh, even better!

chocol4te: I have made these changes :)

nrf52-hal-common/src/spim.rs
chocol4te: Done :)

Yatekii:
Looks good to me now.

The changes with the embedded-hal can be added once its new version is live.

PR #65, opened 2019-02-27
Implement fmt::Write for Uarte

jonas-schievink:
I did not implement the traits in `embedded_hal::serial` because I couldn't tell whether they're allowed to block (and their documentation didn't help).

### DIFF
--- a/nrf52-hal-common/src/uarte.rs
+++ b/nrf52-hal-common/src/uarte.rs
@@ -6,6 +6,7 @@
 //! - nrf52840: Section 6.34
 use core::ops::Deref;
 use core::sync::atomic::{compiler_fence, Ordering::SeqCst};
+use core::fmt;
 
 use crate::target::{
     uarte0,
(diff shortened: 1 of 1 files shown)


jonas-schievink:
Ah apparently we could get this "for free" if we'd instead implement the hal trait now since https://github.com/rust-embedded/embedded-hal/pull/111 was merged. Any preference here? (this would need to wait for a new embedded-hal release in that case)

jonas-schievink:
Yes, but here the simplest *implementation* would be a blocking one, not the user. I'll look into building a "proper" non-blocking implementation, but is that really that useful given that we'd always only DMA a single byte?

jonas-schievink:
Actually it looks like the next embedded-hal release might still take a while, so I'd like to get the blocking `fmt::Write` impl merged if possible, independently of whether we also add an impl for the embedded-hal traits.

jamesmunns:
CC https://github.com/nrf-rs/nrf52-hal/issues/37

Yatekii:
> This looks good to me, we should probably re-evaluate eager copies like this again in the future, or at least refactor "is this in RAM" detection into a single util, and be consistent about how we do this.

Yep, as stated on IRC, I think here eagerly copying is okay as we expicitely operate on strings. And not really big chunks of streamed data. Otherwise we would use &[u8] I think.
I agree that we should make this an util to be consistent about it.

PR #66, opened 2019-02-27
Add reference manual links to the README

### LINKED ISSUE #54: Add resources for the two nrf52 chips to the README/repo
I think it would be useful to add resources such as reference manuals to the repo.

Question is if it should be links to the PDFs in the README or even included PDFs in the repo that are omitted in a cargo publish.
I would really welcome that as google et. al always show prelim. versions as first hits, potentially fooling users (e.g the 52840er chip had 255byte wide EDMA transfer buffers for 2/3 peripherals when in fact it really has 2^16 - 1 bytes wide transfer buffers for all peripherals (at least at release (v1.0))).

Adding just a link keeps the storage footprint of the repo low and doesn't clutter it so much, where as committing the actual file would prevent broken links in case nordic moving stuff (which they seem to do some time; didn't happen for the reference manual yet tho; afaik).

What do y'all think about this?

Including [0, 1] would up the repo size about 15 Megs.

[2] http://infocenter.nordicsemi.com/pdf/nRF52840_PS_v1.0.pdf
[1] http://infocenter.nordicsemi.com/pdf/nRF52832_PS_v1.4.pdf

Yatekii:
Title.

Closes #54

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,12 @@
 * [`nrf52832-hal`](https://docs.rs/nrf52832-hal)
 * [`nrf52840-hal`](https://docs.rs/nrf52840-hal)
 
+## Resources on the nRF52 devices
+
+- [nRF52840 Reference Manual](http://infocenter.nordicsemi.com/pdf/nRF52840_PS_v1.0.pdf)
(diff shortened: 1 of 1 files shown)


simonsso:
LGTM

PR #68, opened 2019-02-28
feat: add temperature interface

chocol4te:
Virtually copy-pasted from @jonas-schievink's PR, https://github.com/nrf-rs/nrf51-hal/pull/17.

### DIFF
--- a/nrf52-hal-common/src/temp.rs
+++ b/nrf52-hal-common/src/temp.rs
@@ -2,6 +2,7 @@
 
 use fpa::I30F2;
 use nb;
+use void::Void;
 use crate::target::TEMP;
 
 /// Integrated temperature sensor.
@@ -45,7 +46,7 @@ impl Temp {
     /// Before calling this, `start_measurement` must be called.
     ///
     /// Returns the measured temperature in °C.
-    pub fn read(&mut self) -> nb::Result<I30F2, ()> {
+    pub fn read(&mut self) -> nb::Result<I30F2, Void> {
         if self.0.events_datardy.read().bits() == 0 {
             return Err(nb::Error::WouldBlock);
         } else {
--- a/nrf52-hal-common/src/temp.rs
+++ b/nrf52-hal-common/src/temp.rs
@@ -1,0 +1,61 @@
+//! Temperature sensor interface.
+
+use fpa::I30F2;
+use nb;
+use crate::target::TEMP;
+
+/// Integrated temperature sensor.
+pub struct Temp(TEMP);
+
+impl Temp {
+    /// Creates a new `Temp`, taking ownership of the temperature sensor's register block.
+    pub fn new(raw: TEMP) -> Self {
+        Temp(raw)
+    }
+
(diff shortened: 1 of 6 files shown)


nrf52-hal-common/src/temp.rs
```diff
@@ -0,0 +1,61 @@
...
+    /// If a measurement was already started, it will be canceled.
+    pub fn measure(&mut self) -> I30F2 {
+        self.stop_measurement();
+        self.start_measurement();
+
+        while self.0.events_datardy.read().bits() == 0 {}
+
+        self.0.events_datardy.reset(); // clear event
+        let raw = self.0.temp.read().bits();
+        I30F2::from_bits(raw as i32)
```
hannobraun: Why not call `read` in a loop instead? This seems to be the same code, basically.

jonas-schievink: Yes, this can just use `nb::block!`

chocol4te: Done :)

chocol4te:
@hannobraun I have tested this on 2 nRF52810's, I do not have any other chips but the datasheets indicate they behave identically on all variants.

Yatekii:
I can test the nRF52840 in the evening/this weekend. The question is how you verify that the readings are correct ;)

chocol4te:
Well I put one in the freezer 🤷‍♂️

Yatekii:
Ah, I wrongfully assumed we want to veryfiy the accuracy, which of course isn't necessary to verify correct driver functionality. Will test ;)

jonas-schievink:
I've checked the datasheets and the nrf52 version is equivalent in data format and interface to the nrf51 sensor, except that the nrf52 chips also have additional registers for linearity compensation (which we don't use, and don't have to use).

chocol4te:
We can open an issue for adding that additional functionality, if there's anyone who needs it

nrf52-hal-common/src/temp.rs
```diff
@@ -0,0 +1,57 @@
...
+            self.0.events_datardy.reset();
+        }
+    }
+
+    /// Tries to read a started measurement (non-blocking).
+    ///
+    /// Before calling this, `start_measurement` must be called.
+    ///
+    /// Returns the measured temperature in °C.
+    pub fn read(&mut self) -> nb::Result<I30F2, ()> {
```
hannobraun: It seems like this function can't fail (except for `WouldBlock`) and `()` is not the right type to represent this. I don't think using the never type (`!`) in this position is stable yet, but `void::Void` would work. I believe `void` is already a dependency of this crate.

```suggestion
    pub fn read(&mut self) -> nb::Result<I30F2, Void> {
```
Note: This suggestion only covers the change required in this line. `Void` would also need to be imported above..

chocol4te: Done :)

chocol4te:
I'm not sure how to mark your suggestion as added,  I just thought it would be easier to manually commit :/

PR #69, opened 2019-03-02
Implement rand::RngCore for Rng

chocol4te:
This PR implements rand's `RngCore` trait for `Rng`. This will not have an additional cost for those depending on the HAL, but does allow generic drivers and libraries. The embedded-hal `rng::Read` trait is not final and is currently feature-gated, while `RngCore` has more widespread adoption.

### DIFF
--- a/nrf52-hal-common/Cargo.toml
+++ b/nrf52-hal-common/Cargo.toml
@@ -19,7 +19,10 @@ edition = "2018"
 [dependencies]
 cortex-m = "0.5.8"
 nb = "0.1.1"
-rand = { version = "0.6.5", default-features = false }
+
+[dependencies.rand]
+default-features = false
+version = "0.6.5"
 
 [dependencies.void]
 default-features = false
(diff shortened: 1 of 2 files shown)


jamesmunns:
Hey @chocol4te, could you please open the formatting into another PR? That will make this change easier to review (but will also cause merge conflicts for all other open PRs).

I'll try and land at least the temperature PR now to avoid some of the conflicts.

jamesmunns:
Also, I'm generally positive on using cargo fmt, but you might get some (reasonable) resistance from @hannobraun :)

I agree with him it can make some code less readable, but in general I am personally okay with the tradeoff.

chocol4te:
Ah okay :D

nrf52-hal-common/Cargo.toml
```diff
@@ -19,6 +19,7 @@ edition = "2018"
 [dependencies]
 cortex-m = "0.5.8"
 nb = "0.1.1"
+rand = { version = "0.6.5", default-features = false }
```
jamesmunns: Could you please use the `[dependencies.rand]` syntax here for consistency?

chocol4te: Done :)

PR #71, opened 2019-03-03
Allow 10 minutes for caching

jamesmunns:


jamesmunns:
Overall, it seems to take more time to download, install, and re-pack the cache, than it actually takes to just download and build everything from scratch. See #72. Closing this in favor of that PR. We can re-evaluate if our list of dependencies or total build time grows too large.

PR #72, opened 2019-03-03
No cache at all!

jamesmunns:
As opposed to #71, this disables caching completely. This is meant as a baseline comparison.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ rust:
   - stable
   - beta
   - nightly
-cache: cargo
 addons:
   apt:
     packages:


jamesmunns:
So, without caching, a build takes about 5 minutes for each of stable/beta/nightly, meaning a total of 15 minutes of build time.

jamesmunns:
It takes less time to download fresh and build than it does to use the cache. This PR disables caching, reducing the total build time from 25-30 minutes down to 15 or so.

hannobraun:
@jamesmunns 
> It takes less time to download fresh and build than it does to use the cache. This PR disables caching, reducing the total build time from 25-30 minutes down to 15 or so.

Thanks for confirming. This matches my intuition, but I never bothered to actually measure it :-)

PR #73, opened 2019-03-03
Use the LLVM linker (only applies to examples and CI builds)

jamesmunns:


Yatekii:
Looks good to me.
I didn't actually use lld yet, but I think we should definitely go for it as a default.

PR #75, opened 2019-03-09
Add embedded_hal::blocking::i2c trait implementation

Istar-Eldritch:
I was trying to use the Twin strut with other crates that use the i2c traits defined in embedded_hal (embedded_graphics) and I found that Twin was not implementing such traits.

This PR adds the trait implementations linking the logic to the existing functions in the default implementation of th eTwin strut. I didn't move the code to the trait implementation because I was not sure if it would break existing code, but I think this would be the right thing to do.

This PR also adds a new example using the TWI peripheral to use an I2C device. The example only uses the write mechanism.

I would like to add more examples using I2C but I do not have any devices I can use to test atm.

If choose to accept this PR, I would like to clarify if the current implementation of the traits is the right approach. Any other feedback or changes are welcome.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ members = [
   "nrf52840-hal",
   "examples/rtfm-demo",
   "examples/spi-demo",
+  "examples/twi-ssd1306",
 ]
 
 [profile.dev]
@@ -19,3 +20,4 @@ lto = false
 debug = true
 lto = true
 opt-level = "s"
+
--- a/examples/twi-ssd1306/README.md
+++ b/examples/twi-ssd1306/README.md
@@ -1,0 +1,18 @@
+# TWI-SSD1306 
+
+TWI write example using an SSD1306 OLED display:
+https://cdn-shop.adafruit.com/datasheets/SSD1306.pdf
+
+After running it you should see the words "Hello Rust!" on the display.
+
+## HW connections
+Pin     Connecton
+P0.26   SCL
+P0.27   SDA
+
+This is designed for the nRF52-DK & the nRF52840-DK board:
+https://www.nordicsemi.com/Software-and-Tools/Development-Kits/nRF52-DK
+https://www.nordicsemi.com/Software-and-Tools/Development-Kits/nRF52840-DK
+
+The TWI device is a 128x64px SSD1306
+https://cdn-shop.adafruit.com/datasheets/SSD1306.pdf
--- a/nrf52-hal-common/src/twim.rs
+++ b/nrf52-hal-common/src/twim.rs
@@ -372,6 +372,31 @@ impl<T> Twim<T> where T: TwimExt {
     }
 }
 
+/// Implementation of embedded_hal::blocking::i2c Traits
+
+impl<T> embedded_hal::blocking::i2c::Write for Twim<T> where T: TwimExt {
+    type Error = Error;
+
+    fn write<'w>(&mut self, addr: u8, bytes: &'w [u8]) -> Result<(), Error> {
+        self.write(addr, bytes)
+    }
+}
+
+impl<T> embedded_hal::blocking::i2c::Read for Twim<T> where T: TwimExt {
+    type Error = Error;
+
+    fn read<'w>(&mut self, addr: u8, bytes: &'w mut [u8]) -> Result<(), Error> {
+        self.read(addr, bytes)
+    }
+}
+
+impl<T> embedded_hal::blocking::i2c::WriteRead for Twim<T> where T: TwimExt {
+    type Error = Error;
+
+    fn write_read<'w>(&mut self, addr: u8, bytes:&'w[u8], buffer: &'w mut [u8]) -> Result<(), Error> {
+        self.write_then_read(addr, bytes, buffer)
+    }
+}
 
 /// The pins used by the TWIN peripheral
 ///
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,4 +14,7 @@ script:
   - cargo build --manifest-path examples/rtfm-demo/Cargo.toml --no-default-features --features="52810" --target thumbv7em-none-eabi
   - cargo build --manifest-path examples/rtfm-demo/Cargo.toml --no-default-features --features="52840"
   - cargo build --manifest-path examples/spi-demo/Cargo.toml
+  - cargo build --manifest-path examples/twi-ssd1306/Cargo.toml
+  - cargo build --manifest-path examples/twi-ssd1306/Cargo.toml --no-default-features --features="52840" --target thumbv7em-none-eabi
+
 
--- a/examples/twi-ssd1306/.cargo/config
+++ b/examples/twi-ssd1306/.cargo/config
@@ -1,0 +1,2 @@
+[target.thumbv7em-none-eabihf]
+runner = "arm-none-eabi-gdb -tui"
--- a/examples/twi-ssd1306/.gdbinit
+++ b/examples/twi-ssd1306/.gdbinit
@@ -1,0 +1,10 @@
+set remotetimeout 60000
+target remote :2331
+set arm force-mode thumb
+
+# Uncomment to enable semihosting, when necessary
+monitor semihosting enable
+
+layout split
+monitor reset
+load
--- a/examples/twi-ssd1306/Cargo.toml
+++ b/examples/twi-ssd1306/Cargo.toml
@@ -1,0 +1,30 @@
+[package]
+name = "twi-ssd1306"
+version = "0.1.0"
+authors = ["James Waples <james@wapl.es>", "Ruben Paz <me@ruben.io>"]
+edition = "2018"
+
+[dependencies]
+cortex-m-rt = "0.6.5"
+ssd1306 = "0.2.4"
+embedded-graphics = "0.4.7"
+panic-semihosting = "0.5.1"
+
+[dependencies.embedded-hal]
+features = ["unproven"]
+version = "0.2"
+
+[dependencies.nrf52832-hal]
+version = "0.7.0"
+path = "../../nrf52832-hal"
+optional = true
+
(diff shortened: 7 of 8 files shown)


Cargo.toml
```diff
@@ -18,4 +19,4 @@ lto = false
 [profile.release]
 debug = true
 lto = true
-opt-level = "s"
```
Yatekii: Please keep the newline.

examples/i2c-demo/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+ssd1306 = "0.2.4"
+embedded-graphics = "0.4.7"
+panic-semihosting = "0.5.1"
+panic-itm = "0.4.0"
+
+[dependencies.embedded-hal]
+features = ["unproven"]
+version = "0.2"
+
+
```
Yatekii: Remove empty line.

examples/i2c-demo/README.md
```diff
@@ -0,0 +1,18 @@
+# i2c-demo
+
+I2C write example code usin a ssd1306 oled display:
```
Yatekii: I2C write example code using an ssd1306 OLED display:

examples/i2c-demo/README.md
```diff
@@ -0,0 +1,18 @@
...
+https://cdn-shop.adafruit.com/datasheets/SSD1306.pdf
+
+After running it you should see the words "Hello Rust!" in the display.
```
Yatekii: After running it you should see the words "Hello Rust!" on the display.

examples/i2c-demo/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+## HW connections
+Pin     Connecton
+P0.26   SCL
+P0.27   SDA
+
+This is designed for nRF52-DK & nRF52840-DK board:
```
Yatekii: This is designed for the nRF52-DK & the nRF52840-DK board:

examples/i2c-demo/README.md
jamwaffles:
```suggestion
I2C write example using an SSD1306 OLED display:
```

examples/i2c-demo/README.md
jamwaffles:
```suggestion
After running it you should see the words "Hello Rust!" on the display.
```

examples/i2c-demo/src/main.rs
```diff
@@ -0,0 +1,63 @@
...
+#[cfg(feature = "52840")]
+use nrf52840_hal::{
+    nrf52840_pac as pac,
+    gpio::*,
+    twim,
+    prelude::TwimExt,
+};
+
+
+/// I2C write example code usin a ssd1306 oled display:
```
Yatekii: I2C write example code using an ssd1306 OLED display:

examples/i2c-demo/README.md
```diff
@@ -0,0 +1,18 @@
...
+https://www.nordicsemi.com/Software-and-Tools/Development-Kits/nRF52-DK
+https://www.nordicsemi.com/Software-and-Tools/Development-Kits/nRF52840-DK
+
+The I2C device is an ssd1306
```
jamwaffles:
```suggestion
The I2C device is a 128x64px SSD1306
```

examples/i2c-demo/src/main.rs
```diff
@@ -0,0 +1,63 @@
...
+/// https://cdn-shop.adafruit.com/datasheets/SSD1306.pdf
+///
+/// Connect sda wire to pin 27 and scl wire to pin 26
+///
+/// You should see the words "Hello Rust!" in the display.
```
Yatekii: You should see the words "Hello Rust!" on the display.

examples/i2c-demo/src/main.rs
jamwaffles:
```suggestion
/// I2C write example code using an SSD1306 OLED display:
```

examples/i2c-demo/src/main.rs
```diff
@@ -0,0 +1,63 @@
...
+    gpio::*,
+    twim,
+    prelude::TwimExt,
+};
+
+
+/// I2C write example code usin a ssd1306 oled display:
+/// https://cdn-shop.adafruit.com/datasheets/SSD1306.pdf
+///
+/// Connect sda wire to pin 27 and scl wire to pin 26
```
jamwaffles:
```suggestion
/// Connect SDA to P0.27 and SCL to pin P0.26
```

examples/i2c-demo/src/main.rs
jamwaffles:
```suggestion
/// You should see the words "Hello Rust!" on the display.
```

examples/i2c-demo/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+name = "i2c-demo"
+version = "0.1.0"
+authors = ["James Waples <james@wapl.es>", "Ruben Paz <me@ruben.io>"]
+edition = "2018"
+
+[dependencies]
+cortex-m-rt = "0.6.5"
+embedded-hal-spy = "0.0.3"
```
jamwaffles: I don't think this is used by the example code so can be removed.

examples/i2c-demo/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+ssd1306 = "0.2.4"
+embedded-graphics = "0.4.7"
+panic-semihosting = "0.5.1"
+panic-itm = "0.4.0"
```
jamwaffles: The example should use _either_ panic-semihosting or panic-itm, not both.

examples/i2c-demo/README.md
Istar-Eldritch: Add display resolution to readme

Istar-Eldritch:
Thank you for the fast feedback @Yatekii, @jamwaffles.

I did the proposed changes. I was aware of the TWI naming issue, as far as I know TWI is a superset of I2C and difference in name is mainly a political and/or copyright issue. It was confusing for me when I first encounter it because I am pretty new to the embedded world, Its a matter of experience and I think other newbies will feel the hit too. I do not think we can do much on this repository about it, so I wont dwell on it.

I propose squashing the typo fixing and the rename into the add example commit to keep a clean merge history.

examples/twi-ssd1306/README.md
```diff
@@ -0,0 +1,18 @@
...
+https://cdn-shop.adafruit.com/datasheets/SSD1306.pdf
+
+After running it you should see the words "Hello Rust!" on the display.
+
+## HW connections
+Pin     Connecton
+P0.26   SCL
+P0.27   SDA
+
+This is designed for nRF52-DK & the nRF52840-DK board:
```
Yatekii:
```suggestion
This is designed for the nRF52-DK & the nRF52840-DK board:
```

Yatekii:
Yeah, the I2C/TWI issue also got me when I tried to mod China PWM ESCs as my first embedded project ;)
I think it's favorable to call it TWI as it is the actual meaning in most cases (even tho people often use I2C) and Nordic uses it too. Afaik the embedded-hal calls it I2C too, which I do not agree with ;)
Maybe @nrf-rs/all do not feel like this, If so, please raise concern.

Appart from that, I think all the grammatical errors but one have been erased (I have suggested a change for that one) and the suggested code "beauty" issues too.
Thanks a lot!

I think you can rebase everything and I'll merge it after that.

nrf52-hal-common/src/twim.rs
```diff
@@ -372,6 +372,31 @@ impl<T> Twim<T> where T: TwimExt {
...
+/// Implementation of embedded_hal::blocking::i2c Traits
+
+impl<T> embedded_hal::blocking::i2c::Write for Twim<T> where T: TwimExt {
+    type Error = Error;
+
+    fn write<'w>(&mut self, addr: u8, bytes: &'w [u8]) -> Result<(), Error> {
+        self.write(addr, bytes)
+    }
+}
+
```
simonsso: Just had a quick glance at the code, but is it not possible to add the implements traits- clause to existing functions? It looks like they already have the correct signature.

Istar-Eldritch: The implemented functions are part of the default implementation of the `Twim` trait. We need to implement multiple new traits. My intuition tells me the correct way of doing so is by moving the implementation details of the function to the trait implementation. This however is likely to break existing implementations using the library as now the code will require importing the Trait for the implementation to be available. So I've done a call to the existing implementation instead.

I do not completely understand your question, can you be more explicit with what you mean by `add the implement traits clause to existing functions`?

Yatekii: I think it's ok to implement it like this for now and change it later for exactly the reason of possible breakage.
What do you think @jamesmunns @hannobraun ?

```
I do not completely understand your question, can you be more explicit with what do you mean by add the implement traits clause to existing functions?
```
I think by that he simply meant what you just described, impl the Trait instead of just functions.

Istar-Eldritch:
@Yatekii I squashed the resolved changes. I there is no rush to close this PR so I would suggest waiting for the https://github.com/nrf-rs/nrf52-hal/pull/75#discussion_r264084800 to be clarified.

nrf52-hal-common/src/twim.rs
hannobraun: > I think it's ok to implement it like this for now and change it later for exactly the reason of possible breakage.
> What do you think @jamesmunns @hannobraun ?

I agree. I think having the two method versions, as done here, is probably the better option in general, although I don't feel strongly about this.

In any case, I'm fine with merging now (assuming the reviewers are happy; I haven't fully reviewed). We can then decide to leave it as is or change it at a later point.

Yatekii:
LGTM! Thanks for your contribution!

PR #76, opened 2019-03-10
Retooling of the SPI code

jamesmunns:
This code has not yet been tested, nor has it been checked for size regressions.

### DIFF
--- a/nrf52-hal-common/src/lib.rs
+++ b/nrf52-hal-common/src/lib.rs
@@ -41,18 +41,49 @@ pub mod target_constants {
     // NRF52832 8 bits1..0xFF
     pub const EASY_DMA_SIZE: usize = 255;
     // Easy DMA can only read from data ram
-    pub const SRAM_LOWER:usize = 0x2000_0000;
-    pub const SRAM_UPPER:usize = 0x3000_0000;
-    pub const FORCE_COPY_BUFFER_SIZE:usize = 255;
+    pub const SRAM_LOWER: usize = 0x2000_0000;
+    pub const SRAM_UPPER: usize = 0x3000_0000;
+    pub const FORCE_COPY_BUFFER_SIZE: usize = 255;
 }
 #[cfg(feature = "52840")]
 pub mod target_constants {
     // NRF52840 16 bits 1..0xFFFF
     pub const EASY_DMA_SIZE: usize = 65535;
     // Limits for Easy DMA - it can only read from data ram
-    pub const SRAM_LOWER:usize = 0x2000_0000;
-    pub const SRAM_UPPER:usize = 0x3000_0000;
-    pub const FORCE_COPY_BUFFER_SIZE:usize = 1024;
+    pub const SRAM_LOWER: usize = 0x2000_0000;
+    pub const SRAM_UPPER: usize = 0x3000_0000;
+    pub const FORCE_COPY_BUFFER_SIZE: usize = 1024;
+}
+
+/// Does this slice resize entirely within RAM?
+pub(crate) fn slice_in_ram(slice: &[u8]) -> bool {
+    let ptr = slice.as_ptr() as usize;
+    ptr >= target_constants::SRAM_LOWER &&
+        (ptr + slice.len()) < target_constants::SRAM_UPPER
+}
+
+/// A handy structure for converting rust slices into ptr and len pairs
+/// for use with EasyDMA. Care must be taken to make sure mutability
+/// guarantees are respected
+pub(crate) struct DmaSlice {
+    ptr: u32,
+    len: u32,
+}
+
+impl DmaSlice {
+    pub fn null() -> Self {
+        Self {
+            ptr: 0,
+            len: 0,
+        }
+    }
+
+    pub fn from_slice(slice: &[u8]) -> Self {
+        Self {
+            ptr: slice.as_ptr() as u32,
+            len: slice.len() as u32,
+        }
+    }
 }
 
 pub use crate::clocks::Clocks;
--- a/nrf52-hal-common/src/lib.rs
+++ b/nrf52-hal-common/src/lib.rs
@@ -55,7 +55,7 @@ pub mod target_constants {
     pub const FORCE_COPY_BUFFER_SIZE: usize = 1024;
 }
 
-/// Does this slice resize entirely within RAM?
+/// Does this slice reside entirely within RAM?
 pub(crate) fn slice_in_ram(slice: &[u8]) -> bool {
     let ptr = slice.as_ptr() as usize;
     ptr >= target_constants::SRAM_LOWER &&
--- a/nrf52-hal-common/src/spim.rs
+++ b/nrf52-hal-common/src/spim.rs
@@ -54,38 +54,21 @@ where
     type Error = Error;
 
     fn transfer<'w>(&mut self, words: &'w mut [u8]) -> Result<&'w [u8], Error> {
-        // Mask on segment where Data RAM is located on nrf52840 and nrf52832
-        // Upper limit is choosen to entire area where DataRam can be placed
-        let needs_copy = !slice_in_ram(words);
-        let chunk_sz = if needs_copy {
-            EASY_DMA_SIZE
-        } else {
-            FORCE_COPY_BUFFER_SIZE
-        };
+        // If the slice isn't in RAM, we can't write back to it at all
+        ram_slice_check(words)?;
 
-        for chunk in words.chunks(chunk_sz) {
-            if needs_copy {
-                // We don't really need to initialize this...
-                let mut buf = [0u8; FORCE_COPY_BUFFER_SIZE];
-                buf[..chunk.len()].copy_from_slice(chunk);
-
-                self.do_spi_dma_transfer(
-                    DmaSlice::from_slice(&buf[..chunk.len()]),
-                    DmaSlice::from_slice(&buf[..chunk.len()]),
-                    |_| {},
-                )?;
-            } else {
-                self.do_spi_dma_transfer(
-                    DmaSlice::from_slice(chunk),
-                    DmaSlice::from_slice(chunk),
-                    |_| {},
-                )?;
-            }
+        for chunk in words.chunks(EASY_DMA_SIZE) {
+            self.do_spi_dma_transfer(
+                DmaSlice::from_slice(chunk),
+                DmaSlice::from_slice(chunk),
+                |_| {},
+            )?;
         }
 
         Ok(words)
     }
 }
+
 impl<T> embedded_hal::blocking::spi::Write<u8> for Spim<T>
 where
     T: SpimExt,
--- a/nrf52-hal-common/src/spim.rs
+++ b/nrf52-hal-common/src/spim.rs
@@ -61,7 +61,6 @@ where
             self.do_spi_dma_transfer(
                 DmaSlice::from_slice(chunk),
                 DmaSlice::from_slice(chunk),
-                |_| {},
             )
         })?;
 
@@ -100,7 +99,7 @@ where
     T: SpimExt,
 {
     fn spi_dma_no_copy(&mut self, chunk: &[u8]) -> Result<(), Error> {
-        self.do_spi_dma_transfer(DmaSlice::from_slice(chunk), DmaSlice::null(), |_| {})
+        self.do_spi_dma_transfer(DmaSlice::from_slice(chunk), DmaSlice::null())
     }
 
     fn spi_dma_copy(&mut self, chunk: &[u8]) -> Result<(), Error> {
@@ -110,7 +109,6 @@ where
         self.do_spi_dma_transfer(
             DmaSlice::from_slice(&buf[..chunk.len()]),
             DmaSlice::null(),
-            |_| {},
         )
     }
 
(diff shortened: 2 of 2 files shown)


jamesmunns:
Checking the size, there is a mix of code size improvements/regressions, depending on your optimization level. opt-level 3 is larger, but opt-levels 's' and 'z' are smaller. Using [`dwm1001`](https://github.com/braun-robotics/rust-dwm1001)'s `dw1000_rx_tx` example, with the following data points for each check:

1. Debug Build
2. Release Build, opt-level 3
3. Release Build, opt-level z
4. Release Build, opt-level s

## Unmodified

```text
   text    data     bss     dec     hex filename
 232048       0      16  232064   38a80 target/thumbv7em-none-eabihf/debug/examples/dw1000_rx_tx
  32358       0      16   32374    7e76 target/thumbv7em-none-eabihf/release/examples/dw1000_rx_tx
  39622       0      16   39638    9ad6 target/thumbv7em-none-eabihf/release/examples/dw1000_rx_tx
  30112	      0	     16	  30128	   75b0	target/thumbv7em-none-eabihf/release/examples/dw1000_rx_tx
```

## Current nrf52-hal master

```text
   text    data     bss     dec     hex filename
 233144       0      16  233160   38ec8 target/thumbv7em-none-eabihf/debug/examples/dw1000_rx_tx
  32358       0      16   32374    7e76 target/thumbv7em-none-eabihf/release/examples/dw1000_rx_tx
  39622       0      16   39638    9ad6 target/thumbv7em-none-eabihf/release/examples/dw1000_rx_tx
  30228	      0	     16	  30244	   7624	target/thumbv7em-none-eabihf/release/examples/dw1000_rx_tx
```

## `retool-spi` branch

```text
   text    data     bss     dec     hex filename
 232760       0      16  232776   38d48 target/thumbv7em-none-eabihf/debug/examples/dw1000_rx_tx
  33478       0      16   33494    82d6 target/thumbv7em-none-eabihf/release/examples/dw1000_rx_tx
  39570       0      16   39586    9aa2 target/thumbv7em-none-eabihf/release/examples/dw1000_rx_tx
  29920	      0	     16	  29936	   74f0	target/thumbv7em-none-eabihf/release/examples/dw1000_rx_tx
```

A summary of all of my tests, including the patches necessary to reproduce are available here: https://gist.github.com/jamesmunns/05395e7fa5c72b509664f8923e58c2a1.

jamesmunns:
Still to-do: Make sure this code actually still works.

jamesmunns:
I also ran similar checks on the `nrf52-hal/examples/spi-demo`, which exercises more of the methods that I changed.

This is just release builds, with test cases:

* Opt Level 3
* Opt Level Z
* Opt Level S

## Current `nrf52-hal` master branch

```text
   text	   data	    bss	    dec	    hex	filename
   2580	      0	      4	   2584	    a18	/home/james/nrf-rs/nrf52-hal/target/thumbv7em-none-eabihf/release/spi-demo
   1858	      0	      4	   1862	    746	/home/james/nrf-rs/nrf52-hal/target/thumbv7em-none-eabihf/release/spi-demo
   1472	      0	      4	   1476	    5c4	/home/james/nrf-rs/nrf52-hal/target/thumbv7em-none-eabihf/release/spi-demo
```

## `retool-spi` branch

```text
   text	   data	    bss	    dec	    hex	filename
   2818	      0	      4	   2822	    b06	/home/james/nrf-rs/nrf52-hal/target/thumbv7em-none-eabihf/release/spi-demo
   2340	      0	      4	   2344	    928	/home/james/nrf-rs/nrf52-hal/target/thumbv7em-none-eabihf/release/spi-demo
   1846	      0	      4	   1850	    73a	/home/james/nrf-rs/nrf52-hal/target/thumbv7em-none-eabihf/release/spi-demo
```

See https://gist.github.com/jamesmunns/408033f736191079c5ec7ac52768b92f for more details

nrf52-hal-common/src/lib.rs
```diff
@@ -41,18 +41,49 @@ pub mod target_constants {
...
+
+/// Does this slice resize entirely within RAM?
+pub(crate) fn slice_in_ram(slice: &[u8]) -> bool {
+    let ptr = slice.as_ptr() as usize;
+    ptr >= target_constants::SRAM_LOWER &&
+        (ptr + slice.len()) < target_constants::SRAM_UPPER
+}
+
+/// A handy structure for converting rust slices into ptr and len pairs
+/// for use with EasyDMA. Care must be taken to make sure mutability
```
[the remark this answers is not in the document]
jamesmunns: Not sure how, it is necessary (at least in SPI) to have two pointers to the same slice - one for the read pointer and one for the write pointer.

nrf52-hal-common/src/lib.rs
```diff
@@ -41,18 +41,49 @@ pub mod target_constants {
...
     // Limits for Easy DMA - it can only read from data ram
-    pub const SRAM_LOWER:usize = 0x2000_0000;
-    pub const SRAM_UPPER:usize = 0x3000_0000;
-    pub const FORCE_COPY_BUFFER_SIZE:usize = 1024;
+    pub const SRAM_LOWER: usize = 0x2000_0000;
+    pub const SRAM_UPPER: usize = 0x3000_0000;
+    pub const FORCE_COPY_BUFFER_SIZE: usize = 1024;
+}
+
+/// Does this slice resize entirely within RAM?
```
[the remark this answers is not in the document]
jamesmunns: Use the suggest feature and then I can just click accept :)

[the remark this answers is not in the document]
Yatekii: I didn't find out how that works; I can try again tho!

Yatekii:
```suggestion
/// Does this slice reside entirely within RAM?
```

jamesmunns:
> I think the code looks nice.

Aw shucks, thanks :D

> What I do not understand is why transfer and write exist as separate functions as those seem to do the same in my view.

Yeah, I could have one complicated transfer function that fits all the use cases, but is complicated. I chose to have progressively more complicated functions, which allow the user to use the one that they need, and let the linker remove the rest.

The methods in question here:

* embedded-hal::transfer
    * Sends all the bytes in the slice, puts the responses back in the same buffer
    * tx.len() == rx.len() always
    * Does not toggle the CS pin automatically
* embedded-hal::write
    * Sends all the bytes in the slice, discards all the responses
    * Does not toggle the CS pin automatically
* Spim::read() - see `transfer_split_uneven()`. This is an old, less descriptive name kept for compatibility
* Spim::transfer()
    * Sends all the bytes in the slice, puts the responses back in the same buffer
    * tx.len() == rx.len() always
    * Does toggle the CS pin automatically
* Spim::transfer_split_even()
    * Sends all the bytes in the outgoing slice, then puts the responses into the (separate) incoming slice.
    * tx's/rx's min(tx.len(), rx.len())
    * Does toggle the CS pin automatically
* Spim::transfer_split_uneven()
    * Sends all the bytes in the outgoing slice, then puts the responses into the (separate) incoming slice.
    * If tx.len() > rx.len(), then the first tx.len() incoming bytes are put in the buffer, and the rest are discarded
    * If rx.len() > rx.len(), then all bytes in tx are sent, and all remaining bytes transferred are the ORC (defined by ::new()
    * Does toggle the CS pin automatically

> I can't really judge if the code actually works. Maybe a sign that we need some form of testing :/

I still plan to test this manually, but we would need some kind of hardware in the loop testing or simulator to test this realistically.

nrf52-hal-common/src/lib.rs
```diff
@@ -41,18 +41,49 @@ pub mod target_constants {
...
+
+impl DmaSlice {
+    pub fn null() -> Self {
+        Self {
+            ptr: 0,
+            len: 0,
+        }
+    }
+
+    pub fn from_slice(slice: &[u8]) -> Self {
```
jamesmunns: This should really be unsafe.

nrf52-hal-common/src/lib.rs
```diff
@@ -41,18 +41,49 @@ pub mod target_constants {
     // NRF52832 8 bits1..0xFF
     pub const EASY_DMA_SIZE: usize = 255;
     // Easy DMA can only read from data ram
-    pub const SRAM_LOWER:usize = 0x2000_0000;
-    pub const SRAM_UPPER:usize = 0x3000_0000;
-    pub const FORCE_COPY_BUFFER_SIZE:usize = 255;
+    pub const SRAM_LOWER: usize = 0x2000_0000;
+    pub const SRAM_UPPER: usize = 0x3000_0000;
+    pub const FORCE_COPY_BUFFER_SIZE: usize = 255;
```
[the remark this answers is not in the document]
Yatekii: Well, but for the nRF52840 this is not the `EASY_DMA_SIZE` ;) Maybe it would be feasible to use it there too but on the nRF52840 this would occupy 16kb of ram ;)

nrf52-hal-common/src/lib.rs
Yatekii: Should it tho? Shouldn't just the user of DmaSlice be aware/unsafe declaring? Because the DmaSlice itself is actually perfectly safe until you use it :)

nrf52-hal-common/src/spim.rs
```diff
@@ -41,77 +37,77 @@ macro_rules! impl_spim_ext {
...
-        let mut offset:usize = 0;
-        while offset < words.len() {
-            let data_len = min(EASY_DMA_SIZE, words.len() - offset);
-            let data_ptr = offset + (words.as_ptr() as usize);
-            offset += data_len;
-
-            self.do_spi_dma_transfer(data_ptr as u32,data_len as u32,data_ptr as u32,data_len as u32,|_|{})?;
-
+        // If the slice isn't in RAM, we can't write back to it at all
+        ram_slice_check(words)?;
```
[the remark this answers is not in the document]
jamesmunns: hmmm, thats a good point.

nrf52-hal-common/src/lib.rs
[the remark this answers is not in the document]
jamesmunns: Yeah, I did this for @Yatekii's reason, I don't think it's reasonable to eat 16k of ram here. Though I do agree I should document that 255 bytes is a semi-arbitrary value that is <= the lowest EASY_DMA_SIZE (== the 810/832 len)

jamesmunns:
By the way, I have left this PR on WIP until I have time to sit down with a logic analyzer and make sure this actually all still works on hardware. If any of you have time to verify that, it would be appreciated :)

nrf52-hal-common/src/spim.rs
```diff
@@ -41,77 +37,83 @@ macro_rules! impl_spim_ext {
...
+where
+    T: SpimExt,
+{
+    fn spi_dma_no_copy(&mut self, chunk: &[u8]) -> Result<(), Error> {
+        self.do_spi_dma_transfer(DmaSlice::from_slice(chunk), DmaSlice::null(), |_| {})
+    }
+
+    fn spi_dma_copy(&mut self, chunk: &[u8]) -> Result<(), Error> {
+        let mut buf = [0u8; FORCE_COPY_BUFFER_SIZE];
+        buf[..chunk.len()].copy_from_slice(chunk);
```
jamesmunns: TODO: make sure `chunk.len() <= FORCE_COPY_BUFFER_SIZE`, or at least document that limitation

nrf52-hal-common/src/spim.rs
```diff
@@ -270,105 +251,153 @@ impl<T> Spim<T> where T: SpimExt {
...
+                (
+                    t.map(|t| DmaSlice::from_slice(t))
+                        .unwrap_or_else(|| DmaSlice::null()),
+                    r.map(|r| DmaSlice::from_slice(r))
+                        .unwrap_or_else(|| DmaSlice::null()),
+                )
+            })
+            .try_for_each(|(t, r)| {
+                self.do_spi_dma_transfer(t, r, |cs| {
+                    if cs {
```
jamesmunns: TODO: We probably don't want to toggle CS between each block, just once for the whole transaction

nrf52-hal-common/src/spim.rs
```diff
@@ -270,105 +251,153 @@ impl<T> Spim<T> where T: SpimExt {
...
+    ) -> Result<(), Error> {
+        ram_slice_check(tx_buffer)?;
+        ram_slice_check(rx_buffer)?;
+
+        let txi = tx_buffer.chunks(EASY_DMA_SIZE);
+        let rxi = rx_buffer.chunks_mut(EASY_DMA_SIZE);
+
+        txi.zip(rxi).try_for_each(|(t, r)| {
+            self.do_spi_dma_transfer(DmaSlice::from_slice(t), DmaSlice::from_slice(r), |cs| {
+                if cs {
```
jamesmunns: Also probably don't want to toggle every time either here.

nrf52-hal-common/src/spim.rs
```diff
@@ -270,105 +251,153 @@ impl<T> Spim<T> where T: SpimExt {
...
-            // safe. Please refer to the explanation there.
+        buffer: &mut [u8],
+    ) -> Result<(), Error> {
+        ram_slice_check(buffer)?;
+
+        buffer.chunks(EASY_DMA_SIZE).try_for_each(|chunk| {
+            self.do_spi_dma_transfer(
+                DmaSlice::from_slice(chunk),
+                DmaSlice::from_slice(chunk),
+                |cs| {
```
jamesmunns: Also not here. Do we want to do this ever?

nrf52-hal-common/src/lib.rs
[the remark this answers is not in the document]
simonsso: @Yatekii  yes, I was only thinking about the 810 and 832 here, as the comment was about code inside the conditional block. @jamesmunns yes a single comment here for the future maintainer of this code.

nrf52-hal-common/src/spim.rs
simonsso: Both to toggle CS and leave it low while waiting to restart should be ok.

nrf52-hal-common/src/spim.rs
simonsso: Interesting concern, `copy_from_slice` will panic it length are not equal, is it possible to compile misconfigured?

Yatekii:
Ok, so on the different functions for transfers topic:

I still do not 100% feel the need for 6 different functions. In my opinion one should always do CS themselves as it sometimes needs non-standard behavior and in the other cases it's really not too much asked to toggle it before the transfer. I can see that others feel differently about this tho.
What bothers me more about all of this is, that the all of the transfer methods have - at least to me - arbitrary names. It could either be `::transfer()` or `::send()` which toggle the CS line implicitely or not. The same goes for the other methods.
In my opinion the method names should reflecta little bit more what they do. I didn't have any specific issues with the current API, except that I can't really make sense when just looking at the function signatures :)

As for testing, if you have any specific wishes, I can do some tests in the lab. I have an nRF52840-DK, an Oscilloscope and a Logic Analyzer here.

hannobraun:
Due to the ongoing impl days and my plan to make some architectural changes today and tomorrow, we (me, @Yatekii, @jamesmunns) decided to merge as many PRs as possible and deal with further changes and potential fixes later. If you disagree with anything that is being merged here, feel free to open issues and new PRs.

PR #80, opened 2019-03-20
SAADC Implementation

chocol4te:
Initial PR for my SAADC work, there are still a few things I need some advice/input on.

- Reading VDD (Probably make a unit struct and implement channel for it)
- Configuration is currently hardcoded with reasonable values, how should this be configured?
- Only a single channel is utilised, and only one AIN can be read at a time, how should multiple be used? What about configurations for multiple values? How will values be returned when the sample task ends?
- Calibration is probably not implemented correctly, make it optional when all the channels are known at creation?
- Abuse of the `Channel` trait, any analog input can be connected to any channel's negative or positive side, this isn't well reflected in the current implementation

Squashing all the commits might also be a good idea, I made too many small changes.

### DIFF
--- a/nrf52-hal-common/src/saadc.rs
+++ b/nrf52-hal-common/src/saadc.rs
@@ -1,7 +1,12 @@
 //use crate::gpio::{Floating, Input, Pin};
-use crate::gpio::{Floating, Input};
-use crate::target::{saadc, SAADC};
-use core::ops::Deref;
+use crate::{
+    gpio::{Floating, Input},
+    target::{saadc, SAADC},
+};
+use core::{
+    ops::Deref,
+    sync::atomic::{compiler_fence, Ordering::SeqCst},
+};
 use embedded_hal::adc::{Channel, OneShot};
 
 // Only 1 channel is allowed right now, a discussion needs to be had as to how
@@ -51,7 +56,7 @@ where
     PIN: Channel<Saadc, ID = u8>,
 {
     type Error = ();
-    fn read(&mut self, pin: &mut PIN) -> nb::Result<u16, Self::Error> {
+    fn read(&mut self, _pin: &mut PIN) -> nb::Result<u16, Self::Error> {
         match PIN::channel() {
             0 => self.0.ch[0].pselp.write(|w| w.pselp().analog_input0()),
             1 => self.0.ch[0].pselp.write(|w| w.pselp().analog_input1()),
@@ -62,10 +67,12 @@ where
             6 => self.0.ch[0].pselp.write(|w| w.pselp().analog_input6()),
             7 => self.0.ch[0].pselp.write(|w| w.pselp().analog_input7()),
             // This can never happen as if another pin was used, there would be a compile time error
-            _ => panic!("Invalid pin"),
+            _ => {
+                return Err(nb::Error::Other(()));
+            }
         }
 
-        let mut val = 0u16;
+        let mut val: u16 = 0;
         self.0
             .result
             .ptr
@@ -74,6 +81,10 @@ where
             .result
             .maxcnt
             .write(|w| unsafe { w.maxcnt().bits(1) });
+
+        // Conservative compiler fence to prevent starting the ADC before the
+        // pointer and maxcount have been set
+        compiler_fence(SeqCst);
 
         self.0.tasks_start.write(|w| unsafe { w.bits(1) });
         self.0.tasks_sample.write(|w| unsafe { w.bits(1) });
--- a/nrf52-hal-common/src/saadc.rs
+++ b/nrf52-hal-common/src/saadc.rs
@@ -1,0 +1,82 @@
+//use crate::gpio::{Floating, Input, Pin};
+use crate::gpio::{Floating, Input};
+use crate::target::{saadc, SAADC};
+use core::ops::Deref;
+use embedded_hal::adc::{Channel, OneShot};
+
+// Only 1 channel is allowed right now, a discussion needs to be had as to how
+// multiple channels should work (See SCAN_MODE)
+
+pub trait SaadcExt: Deref<Target = saadc::RegisterBlock> + Sized {
+    fn constrain(self) -> Saadc;
+}
+
+impl Channel<Saadc> for crate::gpio::p0::P0_02<Input<Floating>> {
+    type ID = u8;
+
+    fn channel() -> <Self as embedded_hal::adc::Channel<Saadc>>::ID {
+        0u8
+    }
+}
+
+pub struct Saadc(SAADC);
+
+impl Saadc {
+    pub fn new(saadc: SAADC) -> Self {
+        saadc.enable.write(|w| w.enable().enabled());
+        saadc.resolution.write(|w| w.val()._12bit());
+        saadc.oversample.write(|w| w.oversample().bypass());
+        saadc.samplerate.write(|w| w.mode().task());
+
+        saadc.ch[0].config.write(|w| {
+            w.refsel()
+                .internal()
+                .gain()
+                .gain1_6()
+                .tacq()
+                ._20us()
+                .mode()
+                .se()
+                .resp()
+                .bypass()
+                .resn()
+                .bypass()
+                .burst()
+                .enabled()
+        });
+        saadc.ch[0].pselp.write(|w| w.pselp().analog_input0());
+        saadc.ch[0].pseln.write(|w| w.pseln().nc());
+
+        // Calibrate
+        saadc.tasks_calibrateoffset.write(|w| unsafe { w.bits(1) });
+        while saadc.events_calibratedone.read().bits() == 0 {}
+
+        Saadc(saadc)
+    }
+}
+
+impl<PIN> OneShot<Saadc, u16, PIN> for Saadc
+where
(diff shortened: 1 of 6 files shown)


simonsso:
There are 25 changed files in this commit, it looks like you have reformatted the existing code too, it is easier to review and if new commits and reformatting is not mixed in the same commit.

nrf52-hal-common/src/saadc.rs
```diff
@@ -0,0 +1,111 @@
...
+            0 => self.0.ch[0].pselp.write(|w| w.pselp().analog_input0()),
+            1 => self.0.ch[0].pselp.write(|w| w.pselp().analog_input1()),
+            2 => self.0.ch[0].pselp.write(|w| w.pselp().analog_input2()),
+            3 => self.0.ch[0].pselp.write(|w| w.pselp().analog_input3()),
+            4 => self.0.ch[0].pselp.write(|w| w.pselp().analog_input4()),
+            5 => self.0.ch[0].pselp.write(|w| w.pselp().analog_input5()),
+            6 => self.0.ch[0].pselp.write(|w| w.pselp().analog_input6()),
+            7 => self.0.ch[0].pselp.write(|w| w.pselp().analog_input7()),
+            // This can never happen as if another pin was used, there would be a compile time error
+            _ => panic!("Invalid pin"),
```
simonsso: Even if this should not be possible I would prefer an Error over an Panic

nrf52-hal-common/src/saadc.rs
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        let mut val = 0u16;
+        self.0
+            .result
+            .ptr
+            .write(|w| unsafe { w.ptr().bits(((&mut val) as *mut _) as u32) });
```
simonsso: In other places of this module` compiler_fence(SeqCst);`  - compiler fence have been used to prevent optimizations as the compiler was not too aware of EasyDMA. Not sure if this way of getting the pointer will prevent this error or not. Could be worth looking into.

nrf52-hal-common/src/saadc.rs
```diff
@@ -0,0 +1,111 @@
...
+            3 => self.0.ch[0].pselp.write(|w| w.pselp().analog_input3()),
+            4 => self.0.ch[0].pselp.write(|w| w.pselp().analog_input4()),
+            5 => self.0.ch[0].pselp.write(|w| w.pselp().analog_input5()),
+            6 => self.0.ch[0].pselp.write(|w| w.pselp().analog_input6()),
+            7 => self.0.ch[0].pselp.write(|w| w.pselp().analog_input7()),
+            // This can never happen as if another pin was used, there would be a compile time error
+            _ => panic!("Invalid pin"),
+        }
+
+        let mut val = 0u16;
```
simonsso: For me, it would be easier to read this as a u16 variable with:

         let mut val:u16 = 0;

given how the pointer is taken from this and then casted to u32 a few lines down

chocol4te:
> There are 25 changed files in this commit, it looks like you have reformatted the existing code too, it is easier to review and if new commits and reformatting is not mixed in the same commit.

Ah, my bad, I've dropped those commits and force pushed. I will address your comments now :)

nrf52-hal-common/src/saadc.rs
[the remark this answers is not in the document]
chocol4te: Done :)

nrf52-hal-common/src/saadc.rs
chocol4te: I agree, done :)

nrf52-hal-common/src/saadc.rs
chocol4te: I've added a fence between setting the pointer and max count, and starting and sampling the ADC, to ensure that we never try to sample into an old pointer. Does that sound correct?

nrf52-hal-common/src/saadc.rs
```diff
@@ -0,0 +1,122 @@
...
+    fn constrain(self) -> Saadc;
+}
+
+pub struct Saadc(SAADC);
+
+impl Saadc {
+    pub fn new(saadc: SAADC) -> Self {
+        saadc.enable.write(|w| w.enable().enabled());
+        saadc.resolution.write(|w| w.val()._12bit());
+        saadc.oversample.write(|w| w.oversample().bypass());
```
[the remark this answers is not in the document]
jamesmunns: Something like this: https://github.com/braun-robotics/rust-dwm1001/blob/master/src/lib.rs#L169-L176

nrf52-hal-common/src/saadc.rs
```diff
@@ -0,0 +1,122 @@
...
+            1 => self.0.ch[0].pselp.write(|w| w.pselp().analog_input1()),
+            2 => self.0.ch[0].pselp.write(|w| w.pselp().analog_input2()),
+            3 => self.0.ch[0].pselp.write(|w| w.pselp().analog_input3()),
+            4 => self.0.ch[0].pselp.write(|w| w.pselp().analog_input4()),
+            5 => self.0.ch[0].pselp.write(|w| w.pselp().analog_input5()),
+            6 => self.0.ch[0].pselp.write(|w| w.pselp().analog_input6()),
+            7 => self.0.ch[0].pselp.write(|w| w.pselp().analog_input7()),
+            // This can never happen as if another pin was used, there would be a compile time error
+            _ => {
+                return Err(nb::Error::Other(()));
```
jamesmunns: If you're 100% sure about this, you can use https://doc.rust-lang.org/core/hint/fn.unreachable_unchecked.html, which will prevent the compiler from even generating code for this - though the tradeoff is that you get UB if you are wrong :)

nrf52-hal-common/src/saadc.rs
```diff
@@ -0,0 +1,122 @@
...
+
+        // Conservative compiler fence to prevent starting the ADC before the
+        // pointer and maxcount have been set
+        compiler_fence(SeqCst);
+
+        self.0.tasks_start.write(|w| unsafe { w.bits(1) });
+        self.0.tasks_sample.write(|w| unsafe { w.bits(1) });
+
+        while self.0.events_end.read().bits() == 0 {}
+        self.0.events_end.reset();
```
jamesmunns: I suggest adding a second fence here, before you return, to force the compiler to re-evaluate that `val` may have changed.

nrf52-hal-common/src/saadc.rs
```diff
@@ -0,0 +1,122 @@
...
+
```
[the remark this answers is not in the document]
chocol4te: Done :)

nrf52-hal-common/src/saadc.rs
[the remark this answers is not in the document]
chocol4te: I'll start on this now :)

nrf52-hal-common/src/saadc.rs
```diff
@@ -0,0 +1,122 @@
+//use crate::gpio::{Floating, Input, Pin};
```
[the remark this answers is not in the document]
chocol4te: Done

nrf52-hal-common/src/saadc.rs
```diff
@@ -0,0 +1,122 @@
...
+    gpio::{Floating, Input},
+    target::{saadc, SAADC},
+};
+use core::{
+    ops::Deref,
+    sync::atomic::{compiler_fence, Ordering::SeqCst},
+};
+use embedded_hal::adc::{Channel, OneShot};
+
+// Only 1 channel is allowed right now, a discussion needs to be had as to how
```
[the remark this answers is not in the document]
chocol4te: Done

nrf52-hal-common/src/saadc.rs
chocol4te: for PIN::channel to not return a value from 0 to 7 someone would have to modify the channel_mappings macro in some way, and not reflect this back here. Nobody *should* modify it, as those mappings are fixed in hardware for all 3 supported variants, and is unlikely to change even for future variants.

I'm not saying I'm 100% sure, I'm saying that if written *correctly*, this will never happen. Of course if a PR or commit slips through that changes this in some way, then it could be a real problem. I think this is unlikely to happen as there will neither be a need to modify it, nor any future changes that would necessitate modifying it, PLUS the fact that seeing the `unreachable_unchecked()` and comment would alert any future reviewer to pay close attention.

So on that basis I have made the change.

nrf52-hal-common/src/saadc.rs
chocol4te: Done :)

chocol4te:
@jamesmunns Is this what you had in mind for the config?

nrf52-hal-common/src/saadc.rs
```diff
@@ -0,0 +1,180 @@
...
+        // The write enums do not implement clone/copy/debug, only the
+        // read ones, hence the need to pull out and move the values
+        let SaadcConfig {
+            resolution,
+            oversample,
+            reference,
+            gain,
+            resistor,
+            time
+        } = config;
```
jamesmunns: Woah, awesome destructuring.

PR #83, opened 2019-03-26
Implement RTC, retool the clocks module

jamesmunns:
Throw another breaking change on the pile


jamesmunns:
Note to reviewers: I am bad, and didn't add doc comments. I'll try to clean that up in the next day so this is ready to merge. if you have any comments now, I'd still appreciate it!

I didn't cover 100% of the options here, mostly just the ones I needed to get external clocks used instead of the internal oscillators. I didn't implement a "freeze" method, because I'm not sure that's a good thing (I think you might need to re-enable the use of the external clocks after sleep? I'd appreciate feedback from someone like @Yatekii or @jonas-schievink who probably know better).

nrf52-hal-common/src/clocks.rs
```diff
@@ -1,61 +1,105 @@
...
+            periph: self.periph,
         }
     }
 }
 
-impl Clocks {
-    pub fn hfclk(&self) -> Hertz {
-        self.hfclk
+impl<H, L> Clocks<H, L, LfOscStopped> {
+    pub fn set_lfclk_src(self, src: LfOscSource, external: bool, bypass: bool) -> Clocks<H, Internal, LfOscStopped> {
```
jamesmunns: This signature is wrong. It always returns LFCLK: Internal, which is not true depending on `src`, I'll likely need to turn this into three methods to get the type signatures correct.

jonas-schievink:
> I think you might need to re-enable the use of the external clocks after sleep?

The datasheet doesn't seem to mention this (unless I missed it), but it should be easy to check by reading `HFCLKSTAT` after wakeup.

jamesmunns:
@jonas-schievink this is the specific section where I got that idea:

> These clocks are only available when the system is in ON mode. When the system enters ON mode, the
internal oscillator (HFINT) clock source will automatically start to be able to provide the required HFCLK
clock(s) for the system.
> 
> The HFINT will be used when HFCLK is requested and HFXO has not been started. The HFXO is started by triggering the HFCLKSTART task and stopped using the HFCLKSTOP task.

jamesmunns:
@nrf-rs/nrf52 this is now ready for review :tada:

jonas-schievink:
Yeah that does sound like it would switch back to `HFINT` then. However, waking up from OFF mode will also reset the chip, so it might not be something that has to be modeled in the clock configuration.

jamesmunns:
Proof this works on (nrf52832/dwm1001) hardware [available here (youtube)](https://youtu.be/UB8OiEFdYgQ?t=3435).

Could I please get an r+?

jamesmunns:
@Yatekii afaik, it's not possible to use an enum to select the oscillator, etc., and still use the type states. For example, something like this wouldn't compile:

```rust
fn select_osc<OscSource>(self, src: Source) -> Clock<OscSource> {
    match src {
        Source::Internal => Clock { src: InternalTypeState },
        Source::External => Clock { src: ExternalTypeState },
    }
}
```

This is because then the function would be able to return two different types, and generics currently only work when monomorphized to a single type.

jamesmunns:
Here's a simplified [playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=5ba00c9de09e2fee9d8b5e98c14a1976).

the code:

```rust
struct VariantA;
struct VariantB;

struct Outer<V> {
    var: V
}

impl<V> Outer<V> {
    fn exchange(self, use_var_a: bool) -> Outer<V> {
        if use_var_a {
            Outer { var: VariantA }
        } else {
            Outer { var: VariantB }
        }
    }
}

fn main() { }
```

It gives these errors:

```text
   Compiling playground v0.0.1 (/playground)
error[E0308]: mismatched types
  --> src/main.rs:11:26
   |
11 |             Outer { var: VariantA }
   |                          ^^^^^^^^ expected type parameter, found struct `VariantA`
   |
   = note: expected type `V`
              found type `VariantA`

error[E0308]: mismatched types
  --> src/main.rs:13:26
   |
13 |             Outer { var: VariantB }
   |                          ^^^^^^^^ expected type parameter, found struct `VariantB`
   |
   = note: expected type `V`
              found type `VariantB`

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0308`.
error: Could not compile `playground`.

To learn more, run the command again with --verbose.
```

jamesmunns:
The other option would be not not use Zero Sized Types (a reasonable choice), but then we have some wasted RAM usage and are limited to run-time correctness checking.

jonas-schievink:
This is a common problem when using typestate. You can't use *values* to decide on what to return, but you can use *types* if you have something like a `LfclkSource` trait:

```rust
trait LfclkSource {}

struct InternalRC;
struct Synth;
struct External;

impl LfclkSource for InternalRC {}
impl LfclkSource for Synth {}
impl LfclkSource for External {}

struct Clocks<L> {
    lfclk: L,
}

impl Clocks<Synth> {
    pub fn default() -> Self {
        Self {
            lfclk: Synth,
        }
    }
}

impl<L> Clocks<L> {
    pub fn set_flclk_src<S: LfclkSource>(self, src: S) -> Clocks<S> {
        Clocks {
            lfclk: src,
        }
    }
}
```

Now you can call `set_flclk_src` with any of the 3 structs.

PR #86, opened 2019-04-12
fix: enable RNG bias correction

### LINKED ISSUE #85: `Rng` is not cryptographically secure
`Rng` implements `CryptoRng` without turning on the bias correction. This is a bad idea since its output is not guaranteed to be uniformly distributed without doing that.

chocol4te:
Closes #85.

### DIFF
--- a/nrf52-hal-common/src/rng.rs
+++ b/nrf52-hal-common/src/rng.rs
@@ -18,6 +18,7 @@ pub trait RngExt : Deref<Target=rng::RegisterBlock> + Sized {
 
 impl RngExt for RNG {
     fn constrain(self) -> Rng {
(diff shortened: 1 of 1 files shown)


nrf52-hal-common/src/rng.rs
```diff
@@ -18,6 +18,7 @@ pub trait RngExt : Deref<Target=rng::RegisterBlock> + Sized {
 
 impl RngExt for RNG {
     fn constrain(self) -> Rng {
+        self.deref().config.write(|w| w.dercen().enabled());
```
jonas-schievink: `.deref()` should not need to be called manually

chocol4te: 🤦‍♂️ fixed

PR #87, opened 2019-04-16
Add board support crate for Adafruit's nrf52832 Bluefruit LE board.

bjc:
This was basically culled from boards/adafruit_nrf52pro. I added a 'cdc' to the board, for the USB/UARTE0 output and wired up the NFC pins.

The example compiles and runs successfully on my board. I am unable to test NFC functionality directly, but the pins should match the board documentation (such as it is from adafruit).

The memory layout assumes you've got the s132 6.1.1 SoftDevice installed on it from the adafruit bootloader, and is required for the `adafruit-nrfutil` to work. I don't have a j-link, and the board itself is kind of a pain to set up with one, so that's what I have to use. The README has instructions for how to get it all loaded.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,7 @@
 [workspace]
 members = [
   "boards/adafruit_nrf52pro",
+  "boards/adafruit-nrf52-bluefruit-le",
   "boards/nRF52-DK",
   "boards/nRF52840-DK",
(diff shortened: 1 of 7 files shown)


Yatekii:
Looks good to me.

More content is always better and this looks well documented and clean.

Thanks for your contribution :)

PR #89, opened 2019-04-17
Do not allow stopping the Low Frequency clock when it's not started.

bjc:
Sorry about the rustfmt changes. The important one is moving `stop_lfclk` into an impl for `Clocks<H, L, LfOscStarted`. This will allow the compiler to prevent stopping the clock when it's not started.


jamesmunns:
Overall LGTM, is there any negative effect to stopping the clock if it isn't running? Or is this aimed at preventing developer mistakes?

Also the rustfmt changes look small enough here, though in the future I would prefer if they were at least separate commits, to make the reviewing easier.

Thanks for the contribution!

bjc:
I actually didn't try stopping the clock while it was stopped. I just noticed this while I was going through the manual for this MCU and figured it should probably be enforced if it's being called out.

Yatekii:
LGTM.

Thanks for the contribution :)

Issue #91, opened 2019-04-19
Add nrf52 team as nrf52810-hal owner


jamesmunns:
CC @hannobraun, could you please add the nrf-rs/nrf52 team as an owner for the nrf52810 crate? Right now you are the only owner.

Edit: On crates.io. You can do this by running the following command from within the crate locally:

```shell
cargo owner --add github:nrf-rs:nrf52
```

hannobraun:
Added the team as owner. Sorry for the oversight.

PR #92, opened 2019-04-19
Commit with pre-release

jamesmunns:
Note: the 810-hal beta has not yet been published due to #91.

CC @nrf-rs/nrf52 @bjc 

If there are no complaints in the next 24 hours or so, I will plan to release this as-is (after removing the `-beta1` suffix)


jamesmunns:
All HAL crates published. CC @hannobraun

PR #95, opened 2019-04-24
 Pass NVIC peripheral into RTC*.{enable, disable}_interrupt.

bjc:
This change makes the rtc module mirror the timer module, which
requires the NVIC to be passed in to enable and disable its
interrupts.

There are some issues I have with this:

  1) RTC has multiple event types tied to its interrupt, and the
     current API requires multiple calls to enable each of those
     events. This means that we enable the interrupt on the NVIC,
     potentially, many times in a row. However, I've checked the
     Cortex-M4 Technical Reference Manual §6, and couldn't find any
     prohibition with doing that. Additionally, I've tested doing this
     on a 52832 and it functions as expected. So, while kind of gross,
     it seems to work ok.

  2) While the motivation for the change is to make RTC similar to
     TIMER, which it conceptually is, I'm not completely sold on how
     TIMER does its thing either. It's definitely convenient to have
     the TIMER API handle the NVIC for you, but it does make it more
     awkward to handle the NVIC independently, IMHO. Additionally,
     there's no obvious way to handle interrupt priorities. In
     general, I think there needs to be a better story around using
     the NVIC.

  3) This breaks the current RTC API. Since we're still 0.x, I think
     this is worth it.

Given the above, I think there's a choice between having peripheral
modules handle their usage of the NVIC versus having the NVIC managed
separately from them. The purpose of this commit is not to answer that
question, but to have symmetry with the existing API.


jamesmunns:
One thing that wasn't caught at this time, this new signature is not compatible with RTFM. In RTFM, the user never has access to the NVIC as this is managed by the framework.

Apologies for not catching this at the time.

I will be opening a PR to make the use of the NVIC optional.

Yatekii:
Cood catch, thx! I am not familiar enough with RTFM :/

PR #97, opened 2019-04-28
Correctly stop incomplete UARTE reception after timeout.

danielwippermann:
I encountered a problem with the `read_timeout` function during the OxidizeConf workshop. When I passed it a four byte buffer two weird things could happen:

1) I send one byte which is received and put in the buffer, but the function returned `Err(Timeout(0))`
2) I send four bytes, correctly resulting in `Ok(())`, but from then on returning `Err(Timeout(4))` without having to send anything.

After reading through the nRF52832 datasheet it seems that an incomplete reception must be `STOPRX`ed and the `FLUSHRX`ed to cause the `RXD.AMOUNT` to be updated correctly.

This PR implements that sequence if the `ENDRX` event is not set / the timeout elapsed.

If you need any additional work done to this PR, just get in contact.

### DIFF
--- a/nrf52-hal-common/src/uarte.rs
+++ b/nrf52-hal-common/src/uarte.rs
@@ -236,6 +236,11 @@ impl<T> Uarte<T> where T: UarteExt {
             }
         }
 
+        if !event_complete {
+            // Cancel the reception if it did not complete until now
+            self.cancel_read();
+        }
+
         // Cleanup, even in the error case
         self.finalize_read();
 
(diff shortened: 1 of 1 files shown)


hannobraun:
Thank you for the PR, @danielwippermann! I'll look into it shortly.

Travis fails because of linker errors. Looks like multiple `cortex-m` versions are imported into the build. Seems unrelated to this PR.

Yatekii:
Yes, the failing builds are most likely because of the bumped versions and @jamesmunns already made PRs for that I think, which we just didn't merge yet.

I think we should really do that during impl days.

PR #103, opened 2019-05-02
Fixed missing export

korken89:
The timer instance is missing, which caused errors when I tried to fix the dwm1001 crate

### DIFF
--- a/nrf52-hal-common/src/lib.rs
+++ b/nrf52-hal-common/src/lib.rs
@@ -83,6 +83,7 @@ impl DmaSlice {
 
 pub use crate::clocks::Clocks;
 pub use crate::delay::Delay;
+pub use crate::rng::Rng;
 pub use crate::rtc::Rtc;
 pub use crate::saadc::Saadc;
 pub use crate::spim::Spim;


korken89:
The Rng export was also missing

nrf52-hal-common/src/lib.rs
```diff
@@ -83,9 +79,10 @@ impl DmaSlice {
 
 pub use crate::clocks::Clocks;
 pub use crate::delay::Delay;
+pub use crate::rng::Rng;
 pub use crate::rtc::Rtc;
 pub use crate::saadc::Saadc;
 pub use crate::spim::Spim;
-pub use crate::timer::Timer;
+pub use crate::timer::{Instance, Timer};
```
bjc: None of the other re-exports pull in symbols from their modules. Is that desirable?

korken89: Not sure, but the Instance is needed to have a generic function taking `Timer<T: Instance>`.
If there is a better fix we can do that, I just followed the error-train and ended up here :)

Yatekii: It would be best if @hannobraun can give his thought to this as he is restructuring things currently and I am not 100% in the loop how he's laying out things.

hannobraun: `Instance` shouldn't be re-exported, as that would conflict with the `Instance` traits of other peripherals. The intended use is `timer::Instance`.

Of course, we could call it `TimerInstance` and re-export it, or re-export it with `timer::Instance as TimerInstance`. I prefer using it as `timer::Instance`, but that's a matter of taste.

korken89: I tested using `timer::Instance`, and this did not compile (it cannot find it). I then wonder what it is causing the error.

korken89: I will test some more, hold this PR for now

korken89: Hmm, if I remove the export (testing in my `rust-dwm1001` PR), the error becomes:

```
error[E0405]: cannot find trait `Instance` in module `timer`
   --> src/lib.rs:823:19
    |
823 |         T: timer::Instance,
    |                   ^^^^^^^^ not found in `timer`

error: aborting due to previous error
```
And `lib.rs` has `use nrf52832_hal::timer`, which should be enough, so something is wrong in its exporting.
Do you have any idea @hannobraun ?

hannobraun: @korken Your downstream PR depends on version `0.8.1`, which doesn't include my changes. Those are not released yet.

korken89: Yeah, you were correct. Fixed in force push!

korken89:
Please hold this PR, I need to fix the branch mess (2 PRs are mixing)

korken89:
Sorry, forgot to tell that only the missing `Rng` is now properly exported, I think it is ready for merge

Yatekii:
LGTM.

PR #104, opened 2019-05-04
Update to be compatible with RTFM

jamesmunns:
Semi-reverts https://github.com/nrf-rs/nrf52-hal/pull/95.

This updates the interfaces to be compatible with RTFM. RTFM does not give the user code access to the NVIC, and enables interrupts automatically.

I suggest we maintain this pattern moving forward for peripherals that have their own interrupt enable bits.

CC @hannobraun @bjc

### DIFF
--- a/nrf52-hal-common/src/rtc.rs
+++ b/nrf52-hal-common/src/rtc.rs
@@ -77,7 +77,7 @@ where
     }
 
     /// Enable the generation of a hardware interrupt from a given stimulus
-    pub fn enable_interrupt(&mut self, int: RtcInterrupt, nvic: &mut NVIC) {
+    pub fn enable_interrupt(&mut self, int: RtcInterrupt, nvic: Option<&mut NVIC>) {
         match int {
             RtcInterrupt::Tick => self.periph.intenset.write(|w| w.tick().set()),
             RtcInterrupt::Overflow => self.periph.intenset.write(|w| w.ovrflw().set()),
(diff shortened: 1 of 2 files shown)


bjc:
This is fine, but a bit clunky, IMHO. Since these are the only two peripherals that have `enable_interrupt` methods, maybe it'd be better to just remove the method entirely?

In the case of `rtc`, we're needlessly enabling the interrupt potentially multiple times, for each possible event type, and in the case of `timer` it hides the use of the comparator, which isn't necessarily desirable either.

I don't have a problem with this merge, but it does make it more important to figure out how to abstract the NVIC, I think.

korken89:
Hi, can we decide on this and move forward?
Currently I am extending the HAL manually for our use, and it would be nice to have support for systems with external interrupt setup (such as RTFM).

hannobraun:
To clarify: I'd like to give others the chance to comment for a few days. No need to leave this open for a few more months :-)

korken89:
Sounds good :)

Yatekii:
I have nothing agindt this (yet). Imo we should merge this. Most people use RTFM I guess and it shouldn't hinder others if I understand right.

hannobraun:
Okay, I stumbled over something today that turns out to be relevant to this issue: `NVIC::enable` and `disable` have been deprecated in the latest `cortex-m` version, and the replacement methods don't need an `NVIC` instance.

So the correct thing to do here is to upgrade to the latest `cortex-m` version and figure out whether we want to do the NVIC stuff in those methods. One way or another, it won't interfere with RTFM, if I understand correctly.

That said, this pull request has been open for months (+ 2 days), so I'm just going to merge this now to get @korken89 unblocked, and leave other improvements for later.

korken89:
Thanks!

PR #105, opened 2019-05-04
Help the type checker out

jamesmunns:
With the new interface, it seems that the type checker gets confused by constructors with fully generic parameters, even when the return type is certain.

Given the following code:

```rust
// Start the clocks
let _clocks: Clocks<ExternalOscillator, ExternalOscillator, LfOscStarted> = Clocks::new(device.CLOCK)
    .enable_ext_hfosc()
    .set_lfclk_src_external(LfOscConfiguration::NoExternalNoBypass)
    .start_lfclk();
```

I get the error:

```
error[E0282]: type annotations needed
   --> nodes/alarm-clock/src/main.rs:155:85
    |
155 |         let _clocks: Clocks<ExternalOscillator, ExternalOscillator, LfOscStarted> = Clocks::new(device.CLOCK)
    |                                                                                     ^^^^^^^^^^^ cannot infer type for `H`
```

This still fails with the same error even with full type ascription:

```rust
// Start the clocks
let _clocks: Clocks<Internal, Internal, LfOscStopped> = Clocks::new(device.CLOCK);
```

This change allows it to work.

### DIFF
--- a/nrf52-hal-common/src/clocks.rs
+++ b/nrf52-hal-common/src/clocks.rs
@@ -33,7 +33,7 @@ pub struct Clocks<H, L, LSTAT> {
     periph: CLOCK,
 }
 
-impl<H, L, LSTAT> Clocks<H, L, LSTAT> {
+impl Clocks<Internal, Internal, LfOscStopped> {
     pub fn new(clock: CLOCK) -> Clocks<Internal, Internal, LfOscStopped> {
         Clocks {
             hfclk: Internal,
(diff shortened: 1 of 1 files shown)


Yatekii:
LGTM.

nrf52-hal-common/src/clocks.rs
```diff
@@ -33,7 +33,7 @@ pub struct Clocks<H, L, LSTAT> {
     periph: CLOCK,
 }
 
-impl<H, L, LSTAT> Clocks<H, L, LSTAT> {
+impl Clocks<Internal, Internal, LfOscStopped> {
     pub fn new(clock: CLOCK) -> Clocks<Internal, Internal, LfOscStopped> {
```
hannobraun: For future reference: If the type arguments is fully specified in the `impl` declaration, the return type of the constructor can be `Self`, which reduces duplication.

PR #110, opened 2019-08-09
Add support for Actinius Icarus board

jonathanpallant:
The Actinius Icarus is a small form factor nRF9160 board - see https://www.actinius.com/get-started.

This PR adds a BSP for the Actinius Icarus board.


boards/actinius-icarus/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+keywords = ["arm", "nrf9160"]
+license = "MIT OR Apache-2.0"
+readme = "README.md"
+name = "actinius-icarus-bsp"
+version = "0.1.0"
+edition = "2018"
+
+[dependencies]
+cortex-m = "0.6.0"
+cortex-m-rt = "0.6.7"
```
jamesmunns: Could you please set this to 0.6.10? We should do this for all other crates as well, but that can be another PR.

boards/actinius-icarus/src/lib.rs
```diff
@@ -0,0 +1,653 @@
+//! Board support crate for the Actinius Icarus -
+//! https://www.actinius.com/get-started
+//!
+//! NB: There is currently no Rust driver for the on-board ST LIS2DH12
```
jamesmunns: @hannobraun is this the same as the dwm1001 accelerometer?

hannobraun: Looks like it. The DWM1001 data sheet say LIS2DH12TR.

boards/actinius-icarus/Cargo.toml
jonathanpallant: Done

boards/actinius-icarus/src/lib.rs
```diff
@@ -0,0 +1,653 @@
...
+    // Accel IRQ0 is on P0_28
+    // Accel IRQ1 is on P0_29
+    pub D30: p0::P0_30<Input<Floating>>,
+    pub D31: p0::P0_31<Input<Floating>>,
+}
+
+/// The LEDs on the Actinius Icarus board
+pub struct Leds {
+    /// Red LED
+    pub led_red: Led,
```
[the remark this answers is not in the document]
jonathanpallant: Fair point :)

jonathanpallant:
I don't really know anything about RTFM, but I don't think anything in my BSP is substantially different to the existing BSPs so I'm happy to leave it as is. If someone wants to put in a PR making them work with RTFM that would be great.

hannobraun:
@jonathanpallant 
> I don't really know anything about RTFM, but I don't think anything in my BSP is substantially different to the existing BSPs so I'm happy to leave it as is. If someone wants to put in a PR making them work with RTFM that would be great.

I think that's reasonable. Merging now.

Yatekii:
> I don't really know anything about RTFM, but I don't think anything in my BSP is substantially different to the existing BSPs so I'm happy to leave it as is. If someone wants to put in a PR making them work with RTFM that would be great.

Yes, I think that's because the issue was only discovered later, after all the other BSPs were created :) And they just never got updated. Not sure tho.

PR #111, opened 2019-08-12
Add board support for nrf52840-mdk.

bjc:


jonas-schievink:
What's the difference between this and https://github.com/nrf-rs/nrf52840-mdk-rs?

jonas-schievink:
We are planning to move all the BSPs to their own repos, so I'll close this PR.

PR #112, opened 2019-08-12
Enable UARTE1 for 52840 MCUs.

bjc:
This is mentioned in the datasheet in §4.2 Table 3 (Instantiation
Table) and exists in the PAC for this chipset.

Tested with a nrf52840-mdk board.

### DIFF
--- a/nrf52-hal-common/src/uarte.rs
+++ b/nrf52-hal-common/src/uarte.rs
@@ -8,6 +8,9 @@ use core::ops::Deref;
 use core::sync::atomic::{compiler_fence, Ordering::SeqCst};
 use core::fmt;
 
+#[cfg(feature="52840")]
+use crate::target::UARTE1;
+
 #[cfg(feature="9160")]
 use crate::target::{
     uarte0_ns as uarte0,
@@ -393,5 +396,5 @@ pub trait Instance: Deref<Target = uarte0::RegisterBlock> {}
 
 impl Instance for UARTE0 {}
 
-#[cfg(feature="9160")]
+#[cfg(any(feature="52840", feature="9160"))]
 impl Instance for UARTE1 {}


jacobrosenthal:
Checks out on an nrf52840 dk. (and works with #102 with a fix hopefully @korken89 saw https://github.com/korken89/nrf52-hal/pull/1)

One annoying note I hadn't seen before, unlike UARTE0, both the interrupt and the handler are named UARTE1 

so with UARTE0 it looks like this
```
use hal::target::{interrupt, TIMER0 as TIM0, UARTE1 as UARTE0};

    #[interrupt(priority = 2, resources = [RX, TX], spawn = [printer, rx_error])]
    fn UARTE0_UART0() {
    }
```

but on UARTE1 you have to import UARTE1 as something else

```
use hal::target::{interrupt, TIMER0 as TIM0, UARTE1 as UARTE1_OTHER};

    #[interrupt(priority = 2, resources = [RX, TX], spawn = [printer, rx_error])]
    fn UARTE1() {
    }
```

bjc:
In general, the peripherals and their interrupts share names. The exceptions are the peripherals which share instances (e.g., `UART0`/`UARTE0` share an instance, and the peripherals are named differently, but the interrupt handler is named `UARTE0_UART0`).

`UARTE1` doesn't share any instances with anything else (there is no `UART1`), so, like most other things, the peripheral and the interrupt have the same name.

I don't use rtfm, so I don't run into this issue. I also only import the `$pac::Interrupt` enum, so it doesn't conflict with the peripheral name itself (e.g., I refer to the peripheral type as `UARTE0`, the interrupt as `Interrupt::UARTE0_UART0` and the isr is `fn UARTE0_UART0() {}`).

korken89:
Hi, is there something blocking this PR?

PR #114, opened 2019-10-01
Fix support for SAADC on nRF9160

jonathanpallant:
* nRF9160 has ADC0..7 on pins 13 through 20.
* The ADC returns a signed value.
* The Actinius Icarus puts battery voltage on A0 (P0_13)
* You need the fields of SaadcConfig to be public, otherwise no-one can change them...

I haven't verified whether the pin/channel mapping for non-9160 devices is correct - I just left that as-is.

### DIFF
--- a/nrf52-hal-common/src/saadc.rs
+++ b/nrf52-hal-common/src/saadc.rs
@@ -1,12 +1,10 @@
-#[cfg(feature="9160")]
-use crate::target::{SAADC_NS as SAADC, saadc_ns as saadc};
+#[cfg(feature = "9160")]
+use crate::target::{saadc_ns as saadc, SAADC_NS as SAADC};
 
-#[cfg(not(feature="9160"))]
-use crate::target::{SAADC, saadc};
+#[cfg(not(feature = "9160"))]
+use crate::target::{saadc, SAADC};
 
-use crate::{
-    gpio::{Floating, Input},
-};
+use crate::gpio::{Floating, Input};
 use core::{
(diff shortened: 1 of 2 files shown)


jonathanpallant:
Any thoughts on this one?

nrf52-hal-common/src/saadc.rs
```diff
@@ -96,12 +92,12 @@ impl Default for SaadcConfig {
     }
 }
 
-impl<PIN> OneShot<Saadc, u16, PIN> for Saadc
+impl<PIN> OneShot<Saadc, i16, PIN> for Saadc
```
Yatekii: Why is this an i16? I am kinda sure counters can't do negative numbers? Or is the Nordic one kinda special? :)

thejpster: https://infocenter.nordicsemi.com/index.jsp?topic=%2Fcom.nordic.infocenter.sdk5.v11.0.0%2Fgroup__nrf__saadc__hal.html

The ADC is differential and I'm sure my testing showed it could go below zero. The existing code would give a number in the high 65k region.

Yatekii: Oh cool, I didn't know that =)
I hope this is true for all the devices we support currently. But since it's in the generic HAL I think it's safe to assume so.
Maybe in the future we could encode the mode (differential, single ended) in a typestate.

Yatekii:
LGTM. Thanks for your contribution :)

PR #116, opened 2019-10-15
Updating to embedded-hal digital::v2

careyk007:
I have updated the GPIO pins to use the embedded-hal v2 traits. I only have an Adafruit nRF52832 board so I haven't been able to test other chips, but since most of the changes were in `nrf52-hal-common` I am confident that this will work for other chips as well.

### DIFF
--- a/nrf52-hal-common/src/gpio.rs
+++ b/nrf52-hal-common/src/gpio.rs
@@ -63,7 +63,6 @@ use crate::target::P0;
 use crate::target::{ P1 };
 
 use crate::hal::digital::v2::{OutputPin, StatefulOutputPin, InputPin};
-use core::convert::Infallible;
 
 impl<MODE> Pin<MODE> {
     /// Convert the pin to be a floating input
@@ -221,7 +220,7 @@ impl<MODE> Pin<MODE> {
 }
 
 impl<MODE> InputPin for Pin<Input<MODE>> {
-    type Error = Infallible;
+    type Error = ();
 
     fn is_high(&self) -> Result<bool, Self::Error> {
         self.is_low().map(|v| !v)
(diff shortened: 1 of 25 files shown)


nrf52-hal-common/src/gpio.rs
```diff
@@ -62,7 +62,8 @@ use crate::target::P0;
 #[cfg(feature = "52840")]
 use crate::target::{ P1 };
 
-use crate::hal::digital::{OutputPin, StatefulOutputPin, InputPin};
+use crate::hal::digital::v2::{OutputPin, StatefulOutputPin, InputPin};
+use core::convert::Infallible;
```
thejpster: Does using this increase our MSRV? IIRC this came in in 1.34.

hannobraun: I don't know what the MSRV is, but if `Infallible` needs to be replaced, it should be replaced with `void::Void`, which is already a dependency of this crate.

careyk007: Awesome, I went ahead and replaced `Infallible` with `void::Void`.

hannobraun:
Thank you for the pull request, @careyk007!

I have one more nit-pick: You added `embedded-hal` as a direct dependency to some crates, to import the new traits from there. I think it would be better to add the new traits to the prelude (`nrf52-hal-common/lib.rs`) and import them through that.

Other than that, I think this is good to merge!

hannobraun:
Thanks for the changes! Travis is failing. I think you're missing an import somewhere.

Also, I was suggesting to re-export the `digital::v2` traits from the `prelude` module in `lib.rs`, but that can also be done in a later step. I think this is good to merge as soon as the CI build is green.

hannobraun:
Looks good to, thanks!

Also see https://github.com/nrf-rs/nrf52-hal/pull/117#issuecomment-543518176.

Issue #115, opened 2019-10-05
Provide periodic timer APIs


jonas-schievink:
The current timer APIs unconditionally enable the `compare0_stop` shortcut:

https://github.com/nrf-rs/nrf52-hal/blob/b2302446fb0317cd46a403c3eb36d5d7d1f69032/nrf52-hal-common/src/timer.rs#L37-L39

If this would be made configurable via a type parameter, `Timer` could provide a `Periodic` timer as well as the current one-shot implementation.

careyk007:
I'm interested in tackling this. Disabling the `compare0_stop` shortcut makes the timer periodic (as expected), but there are probably users who are expecting the default one-shot implementation. The options I have so far for exposing both approaches to the user are outlined below.

a) Pass an Enum to the `new` method. This would look something like this:
``` Rust
pub enum TimerType {
    OneShot,
    Periodic,
}

pub fn new(timer: T, type: TimerType) -> Self {
    match type {
        OneShot => {
            timer
                .shorts
                .write(|w| w.compare0_clear().enabled().compare0_stop().enabled());
        },
        Periodic => {
            timer
                .shorts
                .write(|w| w.compare0_clear().enabled().compare0_stop().disabled());
        }
    }
    // ...
}
```

b) Provide functions for starting a repeating timer and a one-shot timer:
``` Rust
pub fn start_oneshot(&mut self, cycles: u32) {
    self.0
        .shorts
        .write(|w| w.compare0_stop().enabled());

    self.start(cycles);
}

pub fn start_periodic(&mut self, cycles: u32) {
    self.0
        .shorts
        .write(|w| w.compare0_stop().disabled());

    self.start(cycles);
}
```

c) Use a state-machine approach similar to `embedded-hal`'s `into_push_pull_output` and `into_floating_input`, etc.:
``` Rust
pub fn into_oneshot(self) -> OneShotTimer {
    self.0
        .shorts
        .write(|w| w.compare0_stop().enabled());

    OneShotTimer(self.0)
}

pub fn into_periodic(self) -> PeriodicTimer {
    self.0
        .shorts
        .write(|w| w.compare0_stop().disabled());

    PeriodicTimer(self.0)
}
```

Option (a) would limit the use of the timer to a specific approach (`Periodic` or `OneShot`). I'm not sure how many users would be switching between a periodic and one-shot timer at runtime though, so it could be a valid approach. The major downside is that it would break backwards API compatibility.

[...]

PR #117, opened 2019-10-16
Periodic timer

careyk007:
In response to #115, I have implemented 2 new types: `OneShotTimer` and `PeriodicTimer`. Additionally, backwards compatibility with `Timer` is maintained.

In order to create a new `OneShotTimer`, there are 3 options:
``` Rust
// Same as before
let timer = Timer::new(p.TIMER0);

// More explicitly
let timer = OneShotTimer::new(p.TIMER0);
// or
let timer = PeriodicTimer::new(p.TIMER0).into_oneshot();
```

In order to create a new `PeriodicTimer`, there are 2 options:
```Rust
let timer = PeriodicTimer::new(p.TIMER0);
// or
let timer = Timer::new(p.TIMER0).into_periodic();
```

Conversion between a `OneShotTimer` and a `PeriodicTimer` is facilitated by the functions `into_oneshot()` and `into_periodic()`:

``` Rust
let oneshot_timer = PeriodicTimer::new(p.TIMER0).into_oneshot();
let periodic_timer = OneShotTimer::new(p.TIMER0).into_periodic();
// This will cause an error since conversion from PeriodicTimer -> PeriodicTimer is forbidden:
let periodic_timer = PeriodicTimer::new(p.TIMER0).into_periodic();
// Similarly, conversion from OneShotTimer -> OneShotTimer is forbidden:
let oneshot_timer = OneShotTimer::new(p.TIMER0).into_oneshot();
```

### DIFF
--- a/nrf52-hal-common/src/timer.rs
+++ b/nrf52-hal-common/src/timer.rs
@@ -17,6 +17,11 @@ use void::{unreachable, Void};
 #[cfg(any(feature = "52832", feature = "52840"))]
 use crate::target::{TIMER3, TIMER4};
 
+use core::marker::PhantomData;
+
+pub struct OneShotTimer;
+pub struct PeriodicTimer;
+
 
 /// Interface to a TIMER instance
 ///
@@ -25,15 +30,13 @@ use crate::target::{TIMER3, TIMER4};
 ///
 /// CC[0] is used for the current/most-recent delay period and CC[1] is used
 /// to grab the current value of the counter at a given instant.
-pub struct Timer<T>(T);
+pub struct Timer<T, U>(T, PhantomData<U>);
 
-impl<T> Timer<T>
+impl<T> Timer<T, OneShotTimer>
 where
     T: Instance,
 {
-    pub const TICKS_PER_SECOND: u32 = 1_000_000;
-
-    pub fn new(timer: T) -> Self {
+    pub fn new(timer: T) -> Timer<T, OneShotTimer> {
         timer
             .shorts
             .write(|w| w.compare0_clear().enabled().compare0_stop().enabled());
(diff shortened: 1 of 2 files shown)


hannobraun:
Thank you, @careyk007!

I would have done this a bit differently: Instead of adding new timer types (`BaseTimer`, `OneShotTimer`, `PerdiodicTimer`), add an additional type parameter to `Timer`. It can be `OneShot` by default, to preserve backward compatibility. Add new methods to timer (`one_shot`, `periodic`) to create a timer of that type. `Timer::new` can just call `one_shot` and be deprecated, or removed outright.

I think it's mostly a matter of taste, but I think it's nicer to have fewer types. However, I'd be fine to merge this as-is, if you prefer.

Yatekii:
I like the typestate approach more too, just because all the things are implemented like that already.

Sorry I didn't think of that earlier :/

careyk007:
I think my first commit may have been similar to what you were thinking. I added an additional type parameter to `Timer<T>` to become `Timer<T, U>` and created `OneShotTimer` and `PeriodicTimer` as empty structs. 

I realized that broke compatibility with code that expected a type `Timer<T>` and so the subsequent commits changed `Timer<T, U>` to `BaseTimer<T, U>` to allow me to create the type `Timer<T>` as `BaseTimer<T, TimerOneShot>`. I then added the types `OneShotTimer` and `PeriodicTimer` as wrappers around `BaseTimer<T, TimerOneShot>` and `BaseTimer<T, TimerPeriodic>` respectively.

If going back to `Timer<T, U>` would be acceptable, I'd be happy to revert my changes.

hannobraun:
Yes, that first commit is pretty much what I was thinking. I'd shorten the type names a bit (`OneShot` and `Periodic`; they're already in a module called `timer`, so no need to be redundant).

You can set a default type for `U` (U = `OneShot`) in the struct declaration. That should preserve backwards compatibility.

careyk007:
Awesome, I don't think I knew you could set default types. I'll revert my changes and shorten the type names.

hannobraun:
Looks good to me! Since mine and @Yatekii's comments have been adressed and nobody else chimed in, I'm merging. If any other maintainers would have liked more time to review, please let me know, so I can adjust my behavior in the future.

I just think we have enough pull requests open as it is, so I don't want to risk forgetting about this and leaving it open.

nrf52-hal-common/src/timer.rs
```diff
@@ -17,6 +17,11 @@ use void::{unreachable, Void};
 #[cfg(any(feature = "52832", feature = "52840"))]
 use crate::target::{TIMER3, TIMER4};
 
+use core::marker::PhantomData;
+
+pub struct OneShot;
+pub struct Periodic;
```
jonas-schievink: These could've been empty `enum`s

careyk007: I'm pretty new to Rust. Out of curiosity, what are the benefits of empty `enum`s versus empty `struct`s?

jonas-schievink: Empty enums can't be constructed, which makes it explicit that these types only exist as compile-time markers. It's a minor thing though, so don't worry about it.

careyk007: That makes sense, thanks!

PR #121, opened 2019-11-23
Bumping  nrf52832 to latest nrf52-hal-common 0.8.1

richardanaya:


Yatekii:
Hey, thanks for this PR. Is there any specific reason you are bumping this? As far as I know, cargo would automatically select 0.8.1.

richardanaya:
My bad, i saw an issue I thought was related due to out of date version :)

Yatekii:
I am not saying it is bad to change this, I am genuinely curious.

I thought best practice was to always leave the most room for choice, such that the user wont have any conflicts if other crates depend on lower versions for whatever reason and that the patch index should always be 0. Maybe I am mistaken.

What do others think about this?

hannobraun:
When I update dependency versions  in Cargo.toml, I usually select the latest available version at that time. If you don't do that, you risk depending on features from versions that the Cargo.toml doesn't explicitly depend on.

Example:
1. Crate A's Cargo.toml has `dependency = "0.1"`. Uses features only available in `0.1.3`.
2. Everything works, everyone's happy.
3. Crate B's Cargo.toml has `dependency = ">= 0.1.0, < 0.1.3"`, to work around some bug.
4. Someone depends on crates A and B, crate A fails to build.

If crate A had `dependency = 0.1.3`, this should at least give the user an explicit error message about the version mismatch.

Other situations also expose the same problem, for example if someone does `git pull` without `cargo update` (and still has an old Cargo.lock).

I think this is mostly a fringe issue though. Most crates don't have overly specific version requirements in Cargo.toml, so if a problem shows up, `cargo update` will usually fix it.

(Off-topic: Some time ago I read an article about a Go package system that always resolves the oldest possible version. So `0.1` and `0.1.2` would resolve to `0.1.2`, even if `0.1.3` is out. This approach prevents this problem from happening and, if I understand correctly, also makes lock files completely unnecessary. Pretty neat approach, I think, provided the tooling makes upgrades easy and nags people about old versions.)

PR #123, opened 2019-11-23
Don't take access to the nvic, no longer necessary

jamesmunns:
Due to updates in `cortex-m`, it is no longer necessary to take access to the NVIC. This PR makes the option a boolean instead.

### DIFF
--- a/nrf52-hal-common/src/timer.rs
+++ b/nrf52-hal-common/src/timer.rs
@@ -110,16 +110,16 @@ where
     /// Enables an interrupt that is fired when the timer reaches the value that
     /// is given as an argument to `start`.
     ///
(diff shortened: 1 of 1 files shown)


jamesmunns:
No objections here, I think either approach is valid. I'll see if anyone else has thoughts on the topic.

jonas-schievink:
This looks unsound to me; can't unmasking an interrupt break critical sections? The docs say so.

hannobraun:
Good point. Another reasons to leave it to the user.

Yatekii:
Personally, I think, I'd prefer letting go of the argument completely, as Hanno suggests, too

jamesmunns:
I totally agree, I just need to update the PR :) this will probably happen this weekend.

PR #126, opened 2019-11-28
Update to svd2rust 0.16-based PACs

JJJollyjim:
This was surprisingly easy! (assuming it actually works on hardware...)

Marking as a draft, since it includes a `patch` section in the root `Cargo.toml` to use the git versions of the PACs. That commit should be reverted once the PAC updates are released.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,3 +26,9 @@ debug = true
 lto = true
 opt-level = "s"
 
+
+[patch.crates-io]
+nrf52810-pac = { git = "https://github.com/JJJollyjim/nrf52810-pac.git" }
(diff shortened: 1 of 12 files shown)


jamesmunns:
Note, for the nrf91 part of this, we are blocked by https://github.com/themadinventor/nrf91-rs/issues/3.

jonas-schievink:
The nRF52 PACs are now at 0.9.0, which is generated by svd2rust 0.16 or 0.17 (they should be compatible). I don't have access to fix the nrf91 crate though.

jonas-schievink:
This was finished in https://github.com/nrf-rs/nrf52-hal/pull/134

PR #127, opened 2019-11-30
[WIP] Support the nRF5340

jonas-schievink:
~This is based on / blocked on #126~

Things left to do:
* [ ] Some GPIOs default to non-GPIO modes, this is not exposed in the HAL
* [ ] Ensure all _NS/_S peripherals are right (_S for app, _NS for net)
* [ ] Make all peripherals set the GPIO mux for pins they use
  * This is a new requirement for this chip, unfortunately
* [ ] Test that things work on the application core
* [ ] Add HAL shim for the network core
* [ ] Test that things work on the network core
* [ ] Add an API to transfer ownership of GPIOs to the network core (can also be done later)


thejpster:
I've updated this using latest master.

jonas-schievink:
I won't finish this PR, but if anyone wants to pick this up, feel free!

Issue #128, opened 2020-01-22
Support non-blocking serial/SPI/... communication


dflemstr:
I'm working on [`embedded-platform`](https://github.com/dflemstr/embedded-platform) and am targeting `nrf52840` devices as the first use-case.

Currently, I'm using this code to support serial writes: https://github.com/dflemstr/embedded-platform/blob/f9ae5e7b686fdbf66ee3f172c755aea543d4c102/platforms/nrf52840/src/serial.rs#L19-L23

I haven't started on SPI/I²C/... mappings but I suspect there will be similar challenges in making the code non-blocking.

The execution model I have chosen is to have the main thread always run `wfi` when there is nothing happening, and rely on interrupts to wake it up, poll a main future, and do some work, before going back to `wfi`.

Ideally, the baseline behavior I would like to see is to break apart some of the blocking transactions into multiple steps.  Let's look at this section for example:

https://github.com/nrf-rs/nrf52-hal/blob/90a8c3c7dbdb478b49580fd2bf2f89bdb9286cc2/nrf52-hal-common/src/uarte.rs#L145-L154

This could either be modeled as an explicit state machine (e.g. an `enum`) where during each state transition, there is some sort of polling method instead of a while loop.  So imagine a `SerialTransaction` object with a `poll()` method that would eventually return a successful status.

If this approach sounds sane, I can make it into a PR.

PR #129, opened 2020-01-23
Implement a polling async SPI interface

dflemstr:
Addresses a part of #128.  The existing code was moved around and with an optimizing compiler I expect it to generate the same code as before.

There are also new polling versions of the transfer functions that don't block.  The idea is that one could call these from interrupt handlers to get non-blocking/low power SPI transfers.  The returned types intentionally don't implement `core::future::Future` since that would require waker management.

See https://github.com/dflemstr/embedded-platform/tree/master/platforms/nrf52840 for higher-level async code that could be supported by this change.


hannobraun:
Thank you for the pull request, @dflemstr! It seems that nobody (including me) has the time to look into it right now. Sorry!

Feel free to ping @nrf-rs/nrf52 and @nrf-rs/nrf91, to remind us, if you feel it's taken too long.

dflemstr:
Pinging @nrf-rs/nrf52 and @nrf-rs/nrf91 :)

nrf52-hal-common/src/spim.rs
```diff
@@ -158,61 +220,12 @@ where
...
-        // after all possible DMA actions have completed
-        compiler_fence(SeqCst);
-
-        if self.0.txd.amount.read().bits() != tx.len {
-            return Err(Error::Transmit);
-        }
-        if self.0.rxd.amount.read().bits() != rx.len {
-            return Err(Error::Receive);
+        while !transfer.poll_complete(&mut self.0)? {
+            core::sync::atomic::spin_loop_hint();
```
jonas-schievink: This is not a spinlock, and I'm not sure this actually emits any instructions on thumbv7, so it probably can be removed (same for the others). If that's not the case and it should be kept, importing the function is a good idea.

nrf52-hal-common/src/spim.rs
```diff
@@ -29,8 +27,71 @@ use embedded_hal::digital::v2::OutputPin;
...
+/// An ongoing transfer that was initiated by a call to `.transfer_even_polling()`.
+///
+/// This transfer must be polled to completion.  Failing to poll it until completed might leave the
+/// peripheral in an inconsistent state.
+#[derive(Debug)]
+#[must_use = "This transfer must be polled to completion.  Failing to poll it until completed might leave the peripheral in an inconsistent state."]
+pub struct SpiEvenTransfer<'a, T> {
+    spim: &'a mut T,
+    state: TransferState,
+    chunks: core::iter::Zip<core::slice::Chunks<'a, u8>, core::slice::ChunksMut<'a, u8>>,
```
jonas-schievink: It would be easier to read if you use a type alias for this (and the other iterators)

nrf52-hal-common/src/spim.rs
```diff
@@ -375,6 +351,304 @@ where
...
+    fn advance<T>(
+        &mut self,
+        spim: &mut T,
+        mut next_chunk: impl FnMut() -> Option<(DmaSlice, DmaSlice)>,
+    ) -> Result<bool, Error>
+    where
+        T: Instance,
+    {
+        loop {
+            match core::mem::replace(self, TransferState::Inconsistent) {
```
jonas-schievink: `use core::mem;` or `use core::mem::replace;` at the top of the module

nrf52-hal-common/src/spim.rs
```diff
@@ -397,6 +671,8 @@ pub enum Error {
     DMABufferNotInDataMemory,
     Transmit,
     Receive,
+    /// The peripheral is in an inconsistent state, because of encountering an error mid-transfer.
```
jonas-schievink:
```suggestion
    /// The peripheral is in an inconsistent state, because it encountered an error mid-transfer.
```

jonas-schievink:
I couldn't find any place where this code calls `set_low` on the chip select pin. Is that handled differently or is this missing?

dflemstr:
@jonas-schievink you are right, that was previously happening in one of the `Drop` impls but it must have disappeared, will fix!

nrf52-hal-common/src/spim.rs
```diff
@@ -375,6 +351,304 @@ where
...
+        tx_buffer: &'b [u8],
+        rx_buffer: &'b mut [u8],
+    ) -> Result<SpiUnevenTransfer<'b, T>, Error> {
+        SpiUnevenTransfer::new(self.spim, tx_buffer, rx_buffer)
+    }
+}
+
+impl<'a, T> Drop for SpiTransaction<'a, T> {
+    fn drop(&mut self) {
+        self.chip_select.set_high().unwrap();
```
dflemstr: @jonas-schievink this is actually where the CS line is pulled high!  I had to move it here to make the code panic safe...

nrf52-hal-common/src/spim.rs
dflemstr: Hmm I couldn't find a good name for these...  In this particular case, for example, we could alias `Chunks` as `SentChunks` or something, but we would then still need the `Zip` type.  I also didn't find a good alias for `ChunksMut` because in some cases, it will be the `ReceivedChunks`, while in some cases, it will be both sent and received chunks...

nrf52-hal-common/src/spim.rs
dflemstr: I added it as a hint to avoid https://github.com/rust-lang/rust/issues/28728.  I have seen that bug being triggered in embedded environments if the surrounding function (in this case `do_spi_dma_transfer`) is called from a function that diverges (i.e. has return type `!`), so I always add this as a precaution (but I'm not sure it always helps...).

In this case, it is a spinlock because all we are really doing is repeatedly doing a volatile read and waiting for a condition to be true (the holder of the lock just happens to be the hardware device, so to speak).  If it weren't a spinlock, we would have to be doing something in the loop instead of spinning (so any empty loop is sort of by definition a spinlock in my opinion, unless you do some IO logic in the actual `while` condition).

Another option here might be to do a `wfi` instead, do you think that would be sound?

I'll add imports in the meantime, at least.

nrf52-hal-common/src/spim.rs
dflemstr: I will at least simplify the imports a bit!

nrf52-hal-common/src/spim.rs
jonas-schievink: A volatile read is a sufficient side effect to avoid rust-lang/rust#28728. `wfi` would be sound, but it would simply lock up the processor if interrupts are disabled (which they might be as far as I can tell).

`spin_loop_hint` only really does anything if you have multiple hardware threads as far as I know. It might not even be sufficient for avoid that LLVM bug and only works by chance, like `loop { continue; }` seems to do today.

nrf52-hal-common/src/spim.rs
```diff
@@ -375,6 +354,305 @@ where
...
+    /// the master transmitting what it wishes to read, and the slave responding
+    /// with the requested data.
+    ///
+    /// Uses the provided chip select pin to initiate the transaction. Transmits
+    /// all bytes in `tx_buffer`, then receives bytes until `rx_buffer` is full.
+    ///
+    /// This method is more complicated than the other `transfer` methods because
+    /// it is allowed to perform transactions where `tx_buffer.len() != rx_buffer.len()`.
+    /// If this occurs, extra incoming bytes will be discarded, OR extra outgoing bytes
+    /// will be filled with the `orc` value.
```
jonas-schievink: What's the `orc` value?

nrf52-hal-common/src/spim.rs
```diff
@@ -242,18 +268,14 @@ where
...
-        let res = buffer.chunks(EASY_DMA_SIZE).try_for_each(|chunk| {
-            self.do_spi_dma_transfer(DmaSlice::from_slice(chunk), DmaSlice::from_slice(chunk))
-        });
+        let mut tx = self.transaction(chip_select);
+        let mut transfer = tx.transfer_polling(buffer)?;
 
-        chip_select.set_high().unwrap();
+        while !transfer.poll_complete()? {
+            spin_loop_hint();
+        }
```
jonas-schievink: It might not hurt to add a method to the transfer types that does this loop, that way users can block on transfers if they want (eg. if the transfer was started in the background, but has to be completed before some operation can start).

nrf52-hal-common/src/spim.rs
```diff
@@ -29,8 +30,71 @@ use embedded_hal::digital::v2::OutputPin;
 /// - The SPIM instances share the same address space with instances of SPIS,
 ///   SPI, TWIM, TWIS, and TWI. You need to make sure that conflicting instances
 ///   are disabled before using `Spim`. See product specification, section 15.2.
+#[derive(Debug)]
 pub struct Spim<T>(T);
```
jonas-schievink: In which cases does `T` implement `Debug` here?

nrf52-hal-common/src/spim.rs
dflemstr: Okay, not sure what happened in my case. I will remove the code and trust you!

nrf52-hal-common/src/spim.rs
```diff
@@ -29,8 +30,71 @@ use embedded_hal::digital::v2::OutputPin;
...
+/// peripheral in an inconsistent state.
+#[derive(Debug)]
+#[must_use = "This transfer must be polled to completion.  Failing to poll it until completed might leave the peripheral in an inconsistent state."]
+pub struct SpiTransfer<'a, T> {
+    spim: &'a mut T,
+    state: TransferState,
+    chunks: slice::ChunksMut<'a, u8>,
+}
+
+/// An ongoing transfer that was initiated by a call to `.transfer_even_polling()`.
```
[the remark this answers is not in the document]
dflemstr: Will fix

nrf52-hal-common/src/spim.rs
```diff
@@ -29,8 +30,71 @@ use embedded_hal::digital::v2::OutputPin;
...
+/// peripheral in an inconsistent state.
+#[derive(Debug)]
+#[must_use = "This transfer must be polled to completion.  Failing to poll it until completed might leave the peripheral in an inconsistent state."]
+pub struct SpiEvenTransfer<'a, T> {
+    spim: &'a mut T,
+    state: TransferState,
+    chunks: iter::Zip<slice::Chunks<'a, u8>, slice::ChunksMut<'a, u8>>,
+}
+
+/// An ongoing transfer that was initiated by a call to `.transfer_uneven_polling()`.
```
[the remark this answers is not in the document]
dflemstr: Will fix

nrf52-hal-common/src/spim.rs
dflemstr: I use this pattern frequently in my own code bases when writing tests.  The `T` would be a mocked instance that debug prints all of the current pin states to aid in debugging, for example.  I've also developed muscle memory to always derive `Debug` because it is quite common to `#[deny(missing_debug_implementations)]` :)

nrf52-hal-common/src/spim.rs
dflemstr: Yep, makes sense

nrf52-hal-common/src/spim.rs
dflemstr: This is copy pasted from the previous code, it's a constructor argument given when creating the SPI master.  I can refer to the constructor here if that helps.

nrf52-hal-common/src/spim.rs
jonas-schievink: Ah, that sounds really useful actually!

nrf52-hal-common/src/spim.rs
jonas-schievink: Looks good, thanks!

jonas-schievink:
I think this looks basically fine now. Have you tested this on real hardware?

dflemstr:
I have tested it on a Particle Argon, but the SPI slave device (which is custom made...) has a few power glitches so I would appreciate some secondary testing by someone with a more reliable SPI slave device

thalesfragoso:
Isn't this API susceptible to the `mem::forget` problem ? See https://docs.rust-embedded.org/embedonomicon/dma.html#memforget

dflemstr:
@thalesfragoso how would you recommend that we fix it in this case?

thalesfragoso:
Right now, I see two solutions:

- `'static` borrows, also explained in the embedonomicon: https://docs.rust-embedded.org/embedonomicon/dma.html#static-bound, you can see an example implementation here: https://github.com/stm32-rs/stm32f1xx-hal/blob/master/src/dma.rs (`Pin` probably wasn't used because it wasn't stabilized back then).

- Using owning pointers, like this implementation: https://github.com/nrf-rs/nrf52-hal/pull/102, I also did an adaption of that to get rid of the size features, you can see it here:
https://github.com/thalesfragoso/usbd-serial/blob/writer-impl/src/pool_serial.rs

dflemstr:
@thalesfragoso @jonas-schievink as you might see from the PR metadata, I've decided to get some clarity around async HAL in https://github.com/rust-embedded/embedded-hal/pull/206 before I'd feel comfortable with picking a direction for how to continue this PR.

PR #132, opened 2020-03-04
Add nRF9160 to README, and use new name for PAC crate.

thejpster:


jonas-schievink:
Looks like more crates need to be updated to the new PAC. Other than that this looks good!

thejpster:
It's worse - some of the peripherals have been renamed. And the 
nrf9160-pac didn't build on docs.rs because .cargo/config specifies a 
target they don't have.

I'll do a 0.2.1 of the PAC without the default target set. No-one should 
be building the PAC by itself anyway, as it's a library not a binary. 
Then I'll work out what peripherals have changed.

thejpster:
Looks likes this got fixed up in e7940e3e!

thejpster:
Ping @jonas-schievink

PR #133, opened 2020-03-09
[WIP] USB support

jonas-schievink:
I have not managed to get this to enumerate yet, but it tries. Also had to work around an erratum that just makes the peripheral Not Work :tm: unless some magic values are written to magic addresses.

### DIFF
--- a/nrf52-hal-common/src/usbd/errata.rs
+++ b/nrf52-hal-common/src/usbd/errata.rs
@@ -1,0 +1,57 @@
+/// Writes `val` to `addr`. Used to apply Errata workarounds.
+unsafe fn poke(addr: u32, val: u32) {
+    *(addr as *mut u32) = val;
+}
+
+/// Reads 32 bits from `addr`.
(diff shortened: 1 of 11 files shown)


nrf52-hal-common/src/usbd/errata.rs
```diff
@@ -0,0 +1,57 @@
+/// Writes `val` to `addr`. Used to apply Errata workarounds.
+unsafe fn poke(addr: u32, val: u32) {
+    *(addr as *mut u32) = val;
+}
+
+/// Reads 32 bits from `addr`.
+unsafe fn peek(addr: u32) -> u32 {
+    *(addr as *mut u32)
+}
+
```
thalesfragoso: Shouldn't this be volatile operations ?

jonas-schievink: Yeah, I thought I made them volatile...

jonas-schievink:
Closing in favor of https://github.com/nrf-rs/nrf-hal/pull/144

PR #134, opened 2020-03-11
update hals for 0.9.0 (0.2.0) release

almindor:
Bump PAC and minor dependencies and fix code for 0.9 release.

NOTE: `TIMER3` and `TIMER4` of the `nrf528*` MCUs now have separate `RegisterBlock` types and cannot be unified. I just commented them out for now, a better solution is welcome.

### DIFF
--- a/nrf52-hal-common/src/timer.rs
+++ b/nrf52-hal-common/src/timer.rs
@@ -2,8 +2,6 @@
 //!
 //! See product specification, chapter 24.
 
-use core::ops::Deref;
-
 #[cfg(feature = "9160")]
 use crate::target::{
     timer0_ns as timer0, Interrupt, NVIC, TIMER0_NS as TIMER0, TIMER1_NS as TIMER1,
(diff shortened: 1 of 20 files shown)


nrf52-hal-common/src/timer.rs
```diff
@@ -17,8 +17,9 @@ use embedded_hal::{prelude::*, timer};
 use nb::{self, block};
 use void::{unreachable, Void};
 
-#[cfg(any(feature = "52832", feature = "52840"))]
-use crate::target::{TIMER3, TIMER4};
+// TODO: TIMER3 and TIMER4 are now timer3::RegisterBlock incompatible with timer0::RegisterBlock
```
almindor: I don't know how to best handle this

jonas-schievink: Probably by either using a macro to generate the impls here (and removing the `: Deref<Target = timer0::RegisterBlock>` bound), or via transmute from one register block to the other. They're compatible, the only difference is that TIMER3 and TIMER4 only have CC0-CC3, so that needs to be accounted for.

The macro approach would be safer though.

almindor: Wouldn't using a macro be a breaking change? The `Instance` trait is public and seems to be used elsewhere as a bound for a timer instance as well.

I understand I need to be able to deref `TIMERX` into it's `RegisterBlock` for the function calls in the end, but I don't see how this can be done without breaking compatibility with the previous `Instance`

jonas-schievink: The PAC update is by itself a breaking change, and there's a lot of stuff in master that's also breaking, so no need to worry about that.

Some trait should probably be kept to identify timer instances. The methods the need to access registers could be moved into the trait and be generated by the `impl_instance` macro.

almindor: Yes putting most of the specific code into the trait can be done. I'm a bit concerned with code duplication that way tho.

Another way which seems to work is adding a `<RB>` generic `RegisterBlock` type to the trait (and all the users of the trait), which should result in smaller overall code, I'll experiment.

almindor:
This should be good to go please check the Timer::Instance trait to see if it makes sense this way.

jonas-schievink:
Looks good. If nobody else has any objections I'll go ahead and merge this tomorrow.

PR #135, opened 2020-03-21
Remove NVIC from timer interrupt methods

jonas-schievink:
This was a soundness hole, since unmasking an interrupt in the NVIC is unsafe.

Closes https://github.com/nrf-rs/nrf52-hal/pull/123


Yatekii:
I guess some formatting is required according to rustfmt :)

Yatekii:
LGTM, if Travis then decides so too :)

PR #136, opened 2020-03-22
add customer UICR registers abstraction

almindor:
Adds a basic erase/store/load UICR abstraction for accessing the 32 `customer` registers on `nrf528**` MCUs.

### DIFF
--- a/nrf52-hal-common/src/uicr.rs
+++ b/nrf52-hal-common/src/uicr.rs
@@ -13,8 +13,14 @@ impl<T> Uicr<T>
 where
     T: Instance,
 {
+    /// Construct a new `Uicr` from `pac::UICR`
     pub fn new(uicr: T) -> Self {
         Self(uicr)
+    }
+
+    /// Release the `pac::UICR` instance back
+    pub fn free(self) -> T {
+        self.0
     }
 
     /// Erase the UICR registers. UICR registers can only be set to `0` bits, additional
--- a/nrf52-hal-common/src/uicr.rs
+++ b/nrf52-hal-common/src/uicr.rs
@@ -31,6 +31,8 @@ impl Uicr {
 
     /// Store a slice of `&[u32]` values to the customer registers with given offset
     /// - offset + slice length must be less than 32
+    /// - initial value after erase is 0xFFFF_FFFFu32
+    /// - UICR registers can only be set to `0` bits, additional overrides back to `1` can only be performed by erasing the UICR registers
     pub fn store_customer(&mut self, nvmc: &mut NVMC, offset: usize, values: &[u32]) {
         assert!(values.len() + offset <= self.0.customer.len()); // ensure we fit
         assert!(nvmc.config.read().wen().is_een() == false); // write + erase is forbidden!
--- a/nrf52-hal-common/src/uicr.rs
+++ b/nrf52-hal-common/src/uicr.rs
@@ -20,7 +20,7 @@ impl Uicr {
 
     /// Erase the UICR registers. UICR registers can only be set to `0` bits, additional
     /// overrides back to `1` can only be performed by erasing the UICR registers.
-    /// - Sets all registers to 0xFFFFu32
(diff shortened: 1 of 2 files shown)


nrf52-hal-common/src/uicr.rs
```diff
@@ -0,0 +1,46 @@
...
+            values[i] = self.0.customer[reg_i].read().customer().bits()
+        }
+
+        values
+    }
+}
+
+pub trait Instance: Deref<Target = uicr::RegisterBlock> {}
+
+impl Instance for UICR {}
```
jonas-schievink: Is there ever more than 1 instance of this?

almindor:
> Cool, seems great to have an API for this!
> 
> Maybe we want to make it part of an NVMC API in the future though, that would make the asserts unnecessary since the NVMC state is under our control.

Hmm I didn't want to hog NVMC because that's used for other stuff. In this case we only "eat up" the UICR and take NVMC as needed. Now that you mention it I think I'll add a "free -> pac::UICR" so people can get it back for the few things left until they're implemented.

nrf52-hal-common/src/uicr.rs
almindor: Not that I know off. All the PACs only list one. It also doesn't seem logical to have more than one instance.

Only nrf528** even have customer registers, although some of the others (like `APPROTECT`) seem to be everywhere.

jonas-schievink: Then it seems unnecessary to make the type generic

almindor: Made it specific

nrf52-hal-common/src/uicr.rs
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /// Release the `pac::UICR` instance back
+    pub fn free(self) -> UICR {
+        self.0
+    }
+
+    /// Erase the UICR registers. UICR registers can only be set to `0` bits, additional
+    /// overrides back to `1` can only be performed by erasing the UICR registers.
+    /// - Sets all registers to 0xFFFFu32
```
jonas-schievink:
```suggestion
    /// - Sets all registers to 0xFFFF_FFFFu32
```

nrf52-hal-common/src/uicr.rs
```diff
@@ -0,0 +1,56 @@
...
+    pub fn erase(&mut self, nvmc: &mut NVMC) {
+        assert!(nvmc.config.read().wen().is_wen() == false); // write + erase is forbidden!
+
+        nvmc.config.write(|w| w.wen().een());
+        nvmc.eraseuicr.write(|w| w.eraseuicr().erase());
+        nvmc.config.reset()
+    }
+
+    /// Store a slice of `&[u32]` values to the customer registers with given offset
+    /// - offset + slice length must be less than 32
```
jonas-schievink: This should probably mention that the new configuration only takes effect after a reset (according to the manual), otherwise that seems like an easy-to-miss footgun

almindor: Well it's a bit more nuanced than that. You can technically set say `0b01101111` for example and then override to `0b00001111` or such no problems. It's only `1` that cannot be set after a `0`. I can add more strong warnings here tho.

jonas-schievink: Ah, that's not what I meant – the manual says that the *chip* needs to be reset for the written UICR data to "take effect". Or does that only refer to the non-customer data like the reset/NFC pin configuration?

(still sounds like a good idea to mention that an erase might be needed though)

almindor: Hmm I think it's only for the "functional" registers. I did do a store and load test with my quicktime setup and it worked fine, I got the stored value back.

jonas-schievink: Okay, it would be very weird if even the customer data would require a reset, so let's not mention that in the docs.

PR #138, opened 2020-04-05
Add HAL for nRF51 series

chocol4te:
As discussed in the Matrix room, my current strategy will be to favour keeping the APIs consistent over exact/correct naming and an increase in code complexity due to many feature gates.

For example, the nRF51 does not have EasyDMA for UART, and so the peripheral is called `UART` but has been feature gated inside the nRF52 `UARTE` module. I think more discussion might be needed, but IMHO the benefit to users who want to be able to write agnostic code for any nRF5x device is greater than the confusion that will arise from the missing methods and inconsistent naming.

EDIT: Spoke too soon, there were many difference and so are now separate modules for the nRF51 versions of peripherals where needed. The API's have still been kept as similar as possible however.

## TODO
- [x] Temp
- [x] Time
- [x] Timer
- [x] Clocks
- [x] GPIO
- [x] ADC
- [x] I2C
- [x] SPI
- [x] Delay


chocol4te:
Once ready, https://github.com/nrf-rs/nrf51/pull/9 will need to be merged and a new version published so that this PR can point to version 0.9.0 of `nrf51`.

thejpster:
Ok, now https://github.com/nrf-rs/nrf52-hal/issues/113 really needs to be done :)

nrf-hal-common/Cargo.toml
```diff
@@ -30,6 +32,11 @@ version = "1.0.2"
 default-features = false
 version = "0.2.3"
 
+[dependencies.nrf51]
+optional = true
+version = "0.9.0"
+git = "https://github.com/chocol4te/nrf51"
```
jonas-schievink, line 38: This needs to be changed before merge

nrf-hal-common/Cargo.toml
```diff
@@ -52,7 +59,8 @@ version = "0.2.3"
 
 [features]
 doc = []
-default = ["52832"]
+default = ["51"]
```
jonas-schievink, line 62: We should probably not have a default feature at all – this crate's purpose is just to be reexported by the facade crates, and there's not really a reason to prefer one chip over another here

nrf-hal-common/src/adc.rs
```diff
@@ -0,0 +1,133 @@
+//! # API for the Analog to Digital converter
```
jonas-schievink, line 1:
```suggestion
//! API for the Analog to Digital converter
```

nrf-hal-common/src/delay_timer.rs
```diff
@@ -0,0 +1,56 @@
...
+use embedded_hal::blocking::delay::{DelayMs, DelayUs};
+
+use crate::timer::{self, Timer};
+pub struct Delay<T, U>(Timer<T, U>);
+
+impl<T: timer::Instance, U> Delay<T, U> {
+    /// Configures a Timer as a delay provider
+    pub fn new(timer: Timer<T, U>) -> Self {
+        Self(timer)
+    }
```
jonas-schievink, line 14: Would it be a problem to simply implement the `DelayMs` and `DelayUs` traits on `Timer<T, U>` directly?

nrf-hal-common/src/twi.rs
```diff
@@ -0,0 +1,303 @@
...
+#[derive(Debug)]
+pub enum Error {
+    Transmit,
+    Receive,
+}
+
+/// Implemented by all TWIM instances
+pub trait Instance: Deref<Target = twi0::RegisterBlock> {}
+
+impl Instance for TWI1 {}
```
jonas-schievink, line 303: No `TWI0`?

nrf51-hal/src/lib.rs
```diff
@@ -0,0 +1,20 @@
...
+pub use nrf51 as pac;
+pub use nrf_hal_common::*;
+
+pub mod prelude {
+    pub use crate::hal::prelude::*;
+    pub use nrf_hal_common::prelude::*;
+}
+
+pub use crate::clocks::Clocks;
+//pub use crate::delay::Delay;
```
jonas-schievink, line 13: Delete?

nrf-hal-common/Cargo.toml
chocol4te: Fixed, now https://github.com/nrf-rs/nrf51/pull/9 should be reviewed and merged :)

jonas-schievink: Published it! You'll need to push something since I can't request a rebuild from Travis.

chocol4te:
Okay to merge?

PR #139, opened 2020-04-10
Rename packages and repository

### LINKED ISSUE #113: Consider renaming repo name
Currently there's nrf52 and nrf9160 crates in this repo, I think it'd be better if that was split up, or if the repo name reflected the contents of the repo. Maybe a generic "nrf-hal"?

chocol4te:
Closes #113.


chocol4te:
We'll also need to rename the repository, someone with more permissions than me will need to do that :)

Also can't forget to edit the repo description.

PR #140, opened 2020-04-11
Add badges to README, and link to crates.io

thejpster:


Yatekii:
This LGTM, I guess nrf51 is not in yet, so we add that later on?

README.md
```diff
@@ -4,14 +4,19 @@
 
 [HAL]: https://crates.io/crates/embedded-hal
 
-## Documentation
+Travis Status: [![Build Status](https://travis-ci.org/nrf-rs/nrf-hal.svg?branch=master)](https://travis-ci.org/nrf-rs/nrf-hal)
```
jonas-schievink, line 7: nit: the text is not aligned with the badge

thejpster: Hmm. Any ideas how to fix this?

Yatekii: You could inline a css style an manipulate the line height. maybe padding/margin works too :) the problem is that the default html behavior is that the text line of origin aligns (which makes sense in most cases). since badges are just text with a background this looks weird now :) but you should be able to write plain hatml in markdown and it will render properly :)

thejpster: Can I propose we merge this, and then think about making it prettier? I'm very keen to get 0.9 out of the door...

jonas-schievink: Or just remove the text, the badge is enough

thejpster: Done

PR #143, opened 2020-05-05
initial ieee802.15.4 API

japaric:
what the title says

this PR adds an `ieee802154::Radio` API that wraps the `RADIO` peripheral and puts it in IEEE 802.15.4 mode.

I haven't checked the silicon erratas of (or tested) other chips (the code contains one workaround for (nRF52840 specific?) silicon errata) so for now I have put the API directly under `nrf52840-hal`. But I guess the API should also work on other nRF52 devices -- I think the nRF51 chips do not have an 802.15.4 mode? And I don't know about other chip families.

The implementation is not super IEEE compliant. It's missing a backoff delay mechanism when doing CCA (Clear Channel Assessment) before transmitting packets and it's also missing IFS (inter-frame spacing) between consecutive packets (the hardware has some built-in support for this but the PR doesn't expose it). But it is sufficient to send packets around and receive them on other nRF52s running the same implementation, though it would be good to check this against some third party implementation like the DWM1001 or the MRF24J40.

There are some TODOs here and there in terms of additional things the API could expose like changing the transmit power. I'm not sure if all those need to be added before a review though. They could be added as follow up PRs too.

TODO items
- [x] API to change TX power (the default value doesn't go very far, like tens of centimer or so in a noisy 2.4 GHz environment)
- [x] API to change the start of frame delimiter (the default value matches the IEEE spec)
- [x] API to extract the LQI (Link Quality Indicator)
- [x] verify that CRC are being checked. May need to send a malformed packet from a different device (MRF24J40?)
- [x] "The total usable payload (PSDU) is 127 octets, but when CRC is being used, this is reduced to 125 octets of usable payload." -- adjust MAX_LEN accordingly

### DIFF
--- a/nrf52840-hal/src/ieee802154.rs
+++ b/nrf52840-hal/src/ieee802154.rs
@@ -250,6 +250,10 @@ impl<'c> Radio<'c> {
     pub fn recv(&mut self, packet: &mut Packet) -> Result<u16, u16> {
         // NOTE we do NOT check the address of `packet`; see comment in `Packet::new` for details
 
+        // clear related events
+        self.radio.events_phyend.reset();
+        self.radio.events_end.reset();
+
         // go to the RXIDLE state
         self.enable_rx();
 
@@ -268,8 +272,6 @@ impl<'c> Radio<'c> {
         // wait until we have received something
         self.wait_for_event(Event::End);
         dma_end_fence();
-        // also clear the PHYEND event
-        self.radio.events_phyend.reset();
 
         let crc = self.radio.rxcrc.read().rxcrc().bits() as u16;
         if self.radio.crcstatus.read().crcstatus().bit_is_set() {
@@ -285,6 +287,10 @@ impl<'c> Radio<'c> {
     /// channel is observed to be *clear* (no transmission is currently ongoing), otherwise no
     /// packet is transmitted and the `Err` variant is returned
     pub fn try_send(&mut self, packet: &Packet) -> Result<(), ()> {
+        // clear related events
+        self.radio.events_phyend.reset();
+        self.radio.events_end.reset();
+
         // NOTE we do NOT check the address of `packet`; see comment in `Packet::new` for details
         // go to the RXIDLE state
         self.enable_rx();
@@ -337,8 +343,6 @@ impl<'c> Radio<'c> {
         // until the PHYEND event is raised
         self.wait_for_event(Event::PhyEnd);
         dma_end_fence();
-        // also clear the END event
-        self.radio.events_end.reset();
 
         Ok(())
     }
@@ -350,6 +354,10 @@ impl<'c> Radio<'c> {
     /// CCA attempts to be spec compliant
     pub fn send(&mut self, packet: &Packet) {
         // NOTE we do NOT check the address of `packet`; see comment in `Packet::new` for details
+
+        // clear related events
+        self.radio.events_phyend.reset();
+        self.radio.events_end.reset();
 
         self.radio.shorts.modify(|_, w| w.ccaidle_txen().set_bit());
 
@@ -403,8 +411,6 @@ impl<'c> Radio<'c> {
         // until the PHYEND event is raised
         self.wait_for_event(Event::PhyEnd);
         dma_end_fence();
-        // also clear the END event
-        self.radio.events_end.reset();
 
         self.radio
             .shorts
--- a/nrf52840-hal/src/ieee802154.rs
+++ b/nrf52840-hal/src/ieee802154.rs
@@ -260,7 +260,7 @@ impl<'c> Radio<'c> {
     }
 
     /// Sample the received signal power (i.e. the presence of possibly interfering signals)
-    /// within the bandwidth of the currently used channel for sample_cycles iterations.
+    /// within the bandwidth of the currently used channel for `sample_cycles` iterations.
     /// Note that one iteration has a sample time of 128μs, and that each iteration produces the
     /// average RSSI value measured during this sample time.
     ///
@@ -294,14 +294,14 @@ impl<'c> Radio<'c> {
                 self.radio.events_edend.reset();
 
                 // note that since we have increased EDCNT, the EDSAMPLE register contains the
-                // maximumrecorded value, not the average
+                // maximum recorded value, not the average
                 let read_lvl = self.radio.edsample.read().edlvl().bits();
                 return read_lvl;
             }
         }
     }
 
-    /// Recevies one radio packet and copies its contents into the given `packet` buffer
+    /// Receives one radio packet and copies its contents into the given `packet` buffer
     ///
     /// This methods returns the `Ok` variant if the CRC included the packet was successfully
     /// validated by the hardware; otherwise it returns the `Err` variant. In either case, `packet`
@@ -645,9 +645,7 @@ impl<'c> Radio<'c> {
     fn wait_for_event(&self, event: Event) {
         match event {
             Event::End => {
-                while self.radio.events_end.read().events_end().bit_is_clear() {
-                    continue;
-                }
+                while self.radio.events_end.read().events_end().bit_is_clear() {}
                 self.radio.events_end.reset();
             }
             Event::PhyEnd => {
@@ -657,9 +655,7 @@ impl<'c> Radio<'c> {
                     .read()
                     .events_phyend()
                     .bit_is_clear()
-                {
-                    continue;
-                }
+                {}
                 self.radio.events_phyend.reset();
             }
         }
@@ -667,9 +663,7 @@ impl<'c> Radio<'c> {
 
     /// Waits until the radio state matches the given `state`
     fn wait_for_state_a(&self, state: STATE_A) {
-        while self.radio.state.read().state() != state {
-            continue;
-        }
+        while self.radio.state.read().state() != state {}
     }
 }
 
--- a/nrf52840-hal/src/ieee802154.rs
+++ b/nrf52840-hal/src/ieee802154.rs
@@ -583,7 +583,7 @@ impl Packet {
     /// NOTE `src` data will be truncated to `MAX_PACKET_SIZE` bytes
     pub fn copy_from_slice(&mut self, src: &[u8]) {
         let len = cmp::min(src.len(), Self::MAX_LEN as usize) as u8;
-        self.buffer[Self::DATA][..len as usize].copy_from_slice(src);
+        self.buffer[Self::DATA][..len as usize].copy_from_slice(&src[..len.into()]);
         self.set_len(len);
     }
 
--- a/nrf52840-hal/src/ieee802154.rs
+++ b/nrf52840-hal/src/ieee802154.rs
@@ -191,10 +191,7 @@ impl<'c> Radio<'c> {
             radio.radio.crcinit.write(|w| w.crcinit().bits(0));
 
             // permanent shortcuts
-            radio
-                .radio
-                .shorts
-                .write(|w| w.ccaidle_txen().set_bit().txready_start().set_bit());
+            radio.radio.shorts.write(|w| w.txready_start().set_bit());
         }
 
         // set default settings
@@ -279,9 +276,77 @@ impl<'c> Radio<'c> {
         }
     }
 
-    /// Sends the given `data` as a single radio packet
+    /// Tries to send the given `packet`
+    ///
+    /// This method performs Clear Channel Assessment (CCA) first and sends the `packet` only if the
+    /// channel is observed to be *clear* (no transmission is currently ongoing), otherwise no
+    /// packet is transmitted and the `Err` variant is returned
+    pub fn try_send(&mut self, packet: &Packet) -> Result<(), ()> {
+        // NOTE we do NOT check the address of `packet`; see comment in `Packet::new` for details
+        // go to the RXIDLE state
+        self.enable_rx();
+
+        // start CCA
+        self.radio
+            .tasks_ccastart
+            .write(|w| w.tasks_ccastart().set_bit());
+
+        loop {
+            if self
+                .radio
+                .events_ccaidle
+                .read()
+                .events_ccaidle()
+                .bit_is_set()
+            {
+                // channel is clear
+                self.radio.events_ccaidle.reset();
+                break;
+            }
+
+            if self
+                .radio
+                .events_ccabusy
+                .read()
+                .events_ccabusy()
+                .bit_is_set()
+            {
+                // channel is busy
+                self.radio.events_ccaidle.reset();
+                return Err(());
+            }
+        }
+
+        // NOTE(unsafe) DMA transfer has not yet started
+        unsafe {
+            self.radio
+                .packetptr
+                .write(|w| w.packetptr().bits(packet.buffer.as_ptr() as u32));
+        }
+
+        // the DMA transfer will start at some point after the following write operation so we place
+        // the barrier here
+        dma_start_fence();
+        // enter TX mode
+        self.radio.tasks_txen.write(|w| w.tasks_txen().set_bit());
+
+        // due to a shortcut the transmission will start automatically so we just have to wait
+        // until the PHYEND event is raised
+        self.wait_for_event(Event::PhyEnd);
+        dma_end_fence();
+
+        Ok(())
+    }
+
+    /// Sends the given `packet`
+    ///
+    /// This is utility method that *consecutively* calls the `try_send` method until it succeeds.
+    /// Note that this approach is *not* IEEE spec compliant -- there must be delay between failed
+    /// CCA attempts to be spec compliant
     pub fn send(&mut self, packet: &Packet) {
         // NOTE we do NOT check the address of `packet`; see comment in `Packet::new` for details
+
+        self.radio.shorts.modify(|_, w| w.ccaidle_txen().set_bit());
 
         // go to the RXIDLE state
         self.enable_rx();
@@ -296,8 +361,8 @@ impl<'c> Radio<'c> {
                 .write(|w| w.packetptr().bits(packet.buffer.as_ptr() as u32));
         }
 
-        // start CCA
         'cca: loop {
+            // start CCA
             self.radio
                 .tasks_ccastart
                 .write(|w| w.tasks_ccastart().set_bit());
@@ -324,17 +389,19 @@ impl<'c> Radio<'c> {
                 {
                     // channel is busy
                     self.radio.events_ccaidle.reset();
-                    // FIXME according to the IEEE spec there should be a backoff delay before
-                    // the next CCA
                     continue 'cca;
                 }
             }
         }
 
         // due to a shortcut the transmission will start automatically so we just have to wait
-        // until the END event
+        // until the PHYEND event is raised
         self.wait_for_event(Event::PhyEnd);
         dma_end_fence();
+
+        self.radio
+            .shorts
+            .modify(|_, w| w.ccaidle_txen().clear_bit());
     }
 
     /// Moves the radio from any state to the DISABLED state
--- a/nrf52840-hal/src/ieee802154.rs
+++ b/nrf52840-hal/src/ieee802154.rs
@@ -3,6 +3,7 @@
 use core::{
     ops::{self, RangeFrom},
     sync::atomic::{self, Ordering},
+    marker::PhantomData,
 };
 
 use embedded_hal::timer::CountDown as _;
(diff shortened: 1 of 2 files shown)


thalesfragoso:
Thanks for this work.

> But I guess the API should also work on other nRF52 devices -- I think the nRF51 chips do not have an 802.15.4 mode? And I don't know about other chip families.

Both the nRF52810 and nRF52832 don't support 802.15.4 mode. 51 probably doesn't support it either.

Yatekii:
Cool stuff! I need to test this on my boards =)

The 52840, 52811 52833 and 5340 support 802.15.4.

Are you aware of the https://crates.io/crates/ieee802154 crate? I think it is higher level (MAC) layer, but could be useful to test this and not reinvent the wheel.

Thanks for putting in the work!

japaric:
> Are you aware of the https://crates.io/crates/ieee802154 crate? I think it is higher level (MAC) layer, but could be useful to test this and not reinvent the wheel.

That's the MAC frame. Yes, that's what you should put in the payload of the `Packet` defined here (which is actually the PHY layer frame) if you want to be IEEE compliant.

Also, yes that would be useful to test this against the MRF24J40, which (iirc) does do HW-level MAC address filtering. Though that may end up being more of a test for the `ieee802154` crate itself.

hannobraun:
@japaric 
> though it would be good to check this against some third party implementation like the DWM1001 or the MRF24J40

The DW1000 uses UWB (ultra-wideband) in the PHY layer, so I don't think it can communicate with the nRF radio (I might be wrong wrong; despite my work with the DW1000/DWM1001, I know very little about radios). 

> Also, yes that would be useful to test this against the MRF24J40, which (iirc) does do HW-level MAC address filtering. Though that may end up being more of a test for the `ieee802154` crate itself.

The DW1000 also does hardware-based address matching and this is used in the driver. To my knowledge, this has never caused any problems with `ieee802154`.

jonathanpallant:
Yes, the DW1000 is a UWB PHY (so around 500 MHz bandwidth at 3 to 10 GHz depending on channel) and hence incompatible with a 2.4 GHz PHY (around 80 MHz bandwidth at 2.4 GHz) based system.

I wasn't aware the DW1000 had hardware address detection, so that's good to know!

Getting back on topic, Rust 802.15.4 on the nRF52 series is excellent. Great work :)

blueluna:
Nice! Looks better than my old implementation.
https://github.com/blueluna/nrf52840-dk-experiments/tree/master/nrf52-radio-802154
I hope to try it out when I get time.

japaric:
for those not getting notifications on new commits being pushed: I have added APIs to change the TX power and the start of frame delimiter; added a method the get a packet's Link Quality Indicator; and added CRC verification to the Radio.recv method (the method now returns a Result)

nrf52840-hal/src/ieee802154.rs
```diff
@@ -0,0 +1,630 @@
...
+/// Clear Channel Assessment method
+pub enum Cca {
+    /// Carrier sense
+    CarrierSense,
+}
+
+/// IEEE 802.15.4 channels
+///
+/// NOTE these are NOT the same as WiFi 2.4 GHz channels
+// TODO it is possible to use non-standard frequencies below 2_400 MHz; should those be exposed too?
```
hubertmis, line 45: This list is correct. To be compliant with IEEE 802.15.4 other frequencies shall not be exposed. The hardware in selected nRF devices is capable of using other frequencies, but it should not be used in 802.15.4 protocol.

nrf52840-hal/src/ieee802154.rs
```diff
@@ -0,0 +1,630 @@
...
+                    .bits(8) // length = 8 bits (but highest bit is reserved and must be `0`)
+                    .s0len()
+                    .clear_bit() // no S0
+                    .s1len()
+                    .bits(0) // no S1
+            });
+
+            radio.radio.pcnf1.write(|w| {
+                w.maxlen()
+                    .bits(Packet::MAX_LEN + 2 /* CRC */) // payload length
```
hubertmis, line 174: This line is a little confusing. When we configure hardware, we usually work on PHY layer. Max PSDU length is 127 and it includes CRC. I would change this constant to `MAX_PSDU_LEN = 127` and use it here without `+ 2`. It would be unambiguous.

nrf52840-hal/src/ieee802154.rs
```diff
@@ -0,0 +1,630 @@
...
+        radio.set_txpower(DEFAULT_TXPOWER);
+
+        radio
+    }
+
+    /// Changes the radio channel
+    pub fn set_channel(&mut self, channel: Channel) {
+        self.disable();
+
+        // NOTE(unsafe) radio is currently disabled
```
hubertmis, line 210: It is OK to change frequency while radio is enabled. The FREQUENCY register is double buffered and it is read on RXEN or TXEN tasks.
There is no need to call `self.disable();` before.

nrf52840-hal/src/ieee802154.rs
```diff
@@ -0,0 +1,630 @@
...
+        unsafe {
+            self.radio
+                .frequency
+                .write(|w| w.map().clear_bit().frequency().bits(channel as u8))
+        }
+    }
+
+    /// Changes the Clear Channel Assessment method
+    pub fn set_cca(&mut self, cca: Cca) {
+        self.disable();
```
hubertmis, line 220: Like with frequency it is OK to change CCA mode while radio is enabled.

nrf52840-hal/src/ieee802154.rs
```diff
@@ -0,0 +1,630 @@
...
+
+        match cca {
+            Cca::CarrierSense => self.radio.ccactrl.write(|w| w.ccamode().carrier_mode()),
+        }
+    }
+
+    /// Changes the Start of Frame Delimiter
+    pub fn set_sfd(&mut self, sfd: u8) {
+        // FIXME don't completely turn off the radio; RXIDLE or TXIDLE are probably OK
```
hubertmis, line 229: I expect that every state is OK, I think this register is also double buffered, but I'm not sure about it, because we have never modified SFD value.

nrf52840-hal/src/ieee802154.rs
```diff
@@ -0,0 +1,630 @@
...
+                }
+            } else {
+                // STATE register is in an invalid state
+                unreachable!()
+            }
+        }
+    }
+
+    /// Moves the radio from any state to the RXEN state
+    fn enable_rx(&mut self) {
```
hubertmis, line 448: There is a possibility that with RADIO SHORTS enabled this function would work in an unexpected way. I would add at least assertion here to verify that selected shorts are disabled.

nrf52840-hal/src/ieee802154.rs
```diff
@@ -0,0 +1,630 @@
...
+        // set up RX buffer
+        unsafe {
+            self.radio
+                .packetptr
+                .write(|w| w.packetptr().bits(packet.buffer.as_mut_ptr() as u32));
+        }
+
+        // start transfer
+        dma_start_fence();
+        self.radio.tasks_start.write(|w| w.tasks_start().set_bit());
```
hubertmis, line 266: That's exacly the place where I would clear `Event::End` flag, to be sure that it is not unintentionally set by some previously executed operation.

nrf52840-hal/src/ieee802154.rs
```diff
@@ -0,0 +1,630 @@
...
+
+        // wait until we have received something
+        self.wait_for_event(Event::End);
```
hubertmis, line 269: I'm not sure what is the design of this driver and whether it is OK, but please note that this is blocking and it can be blocked for a very long time (infinite in theory, if this is the only device in the network).

nrf52840-hal/src/ieee802154.rs
```diff
@@ -0,0 +1,630 @@
...
+
+            if self
+                .radio
+                .events_ccabusy
+                .read()
+                .events_ccabusy()
+                .bit_is_set()
+            {
+                // channel is busy
+                self.radio.events_ccaidle.reset();
```
hubertmis, line 315: Typo: `events_ccaidle` -> `events_ccabusy`

nrf52840-hal/src/ieee802154.rs
```diff
@@ -0,0 +1,630 @@
...
+            Err(())
+        }
+    }
+
+    /// Tries to send the given `packet`
+    ///
+    /// This method performs Clear Channel Assessment (CCA) first and sends the `packet` only if the
+    /// channel is observed to be *clear* (no transmission is currently ongoing), otherwise no
+    /// packet is transmitted and the `Err` variant is returned
+    pub fn try_send(&mut self, packet: &Packet) -> Result<(), ()> {
```
hubertmis, line 284: It is good to use `CCAIDLE_TXEN` short for this operation to make sure that the transmission takes place right after CCA. With the design of this function, there is unknown software delay between CCAIDLE event and TXEN task.

nrf52840-hal/src/ieee802154.rs
```diff
@@ -0,0 +1,630 @@
...
+        if self.radio.crcstatus.read().crcstatus().bit_is_set() {
+            Ok(())
+        } else {
+            Err(())
+        }
+    }
+
+    /// Tries to send the given `packet`
+    ///
+    /// This method performs Clear Channel Assessment (CCA) first and sends the `packet` only if the
```
hubertmis, line 281: Please note, that this method shall not be used for ACK TX, because you shall not perform CCA when transmitting ACKs

nrf52840-hal/src/ieee802154.rs
hubertmis: I see it is applied in `send`. Is there any reason here it is handled by software?

nrf52840-hal/src/ieee802154.rs
```diff
@@ -0,0 +1,630 @@
...
+        'cca: loop {
+            // start CCA
+            self.radio
+                .tasks_ccastart
+                .write(|w| w.tasks_ccastart().set_bit());
+
+            loop {
+                if self
+                    .radio
+                    .events_ccaidle
```
hubertmis, line 373: You could wait for PHYEND instead. There is no need to wait here for CCAIDLE and for PHYEND later

nrf52840-hal/src/ieee802154.rs
```diff
@@ -0,0 +1,630 @@
...
+
+    /// Moves the radio from any state to the DISABLED state
+    fn disable(&mut self) {
+        // See figure 110 in nRF52840-PS
+        loop {
+            if let Variant::Val(state) = self.radio.state.read().state().variant() {
+                match state {
+                    STATE_A::DISABLED => return,
+
+                    STATE_A::RXRU | STATE_A::RXIDLE | STATE_A::TXRU | STATE_A::TXIDLE => {
```
hubertmis, line 415: `RXIDLE` state may indicate ongoing CCA procedure that may result in TX with radio shorts. To be fully safe, following set of tasks should be triggered:
`CCASTOP`
`STOP`
`DISABLE`

nrf52840-hal/src/ieee802154.rs
```diff
@@ -0,0 +1,630 @@
...
+        self.wait_for_event(Event::PhyEnd);
+        dma_end_fence();
+
+        self.radio
+            .shorts
+            .modify(|_, w| w.ccaidle_txen().clear_bit());
+    }
+
+    /// Moves the radio from any state to the DISABLED state
+    fn disable(&mut self) {
```
hubertmis, line 408: There is a possibility that with RADIO SHORTS enabled this function would work in an unexpected way. I would add at least assertion here to verify that selected shorts are disabled.

nrf52840-hal/src/ieee802154.rs
```diff
@@ -0,0 +1,630 @@
...
+}
+
+enum Event {
+    End,
+    PhyEnd,
+}
+
+/// An IEEE 802.15.4 packet
+///
+/// This `Packet` is closest to the PPDU (PHY Protocol Data Unit) defined in the IEEE spec. The API
```
hubertmis, line 546: `closest to the PPDU` is ambiguous. I would state here clearly that this packet is PHR concatenated with PSDU

nrf52840-hal/src/ieee802154.rs
```diff
@@ -0,0 +1,630 @@
...
+        };
+        packet.set_len(0);
+        packet
+    }
+
+    /// Fills the packet with given `src` data
+    ///
+    /// NOTE `src` data will be truncated to `MAX_PACKET_SIZE` bytes
+    pub fn copy_from_slice(&mut self, src: &[u8]) {
+        let len = cmp::min(src.len(), Self::MAX_LEN as usize) as u8;
```
hubertmis, line 585: I don't think it is a good idea to silently limit the given slice size. When user of this function wants to transmit 200 bytes packet and calls this function, would get only 125 bytes transmitted and would not be notified that there was a problem. I would change function signature to return `Result` or add assertion.

nrf52840-hal/src/ieee802154.rs
```diff
@@ -0,0 +1,630 @@
...
+    /// Returns the size of this packet's payload
+    pub fn len(&self) -> u8 {
+        self.buffer[Self::PHY_HDR] - Self::CRC
+    }
+
+    /// Changes the size of the packet's payload
+    ///
+    /// NOTE `len` will be truncated to `MAX_LEN` bytes
+    pub fn set_len(&mut self, len: u8) {
+        let len = cmp::min(len, Self::MAX_LEN);
```
hubertmis, line 599: Again silently truncated. Using such API could be a source of hard to find bugs in the higher layers.

japaric:
@hubertmis thanks for the great feedback! (I have seen it just now). I'll incorporate it in the PR in the coming days.

nrf52840-hal/src/ieee802154.rs
```diff
@@ -0,0 +1,662 @@
...
+// See figure 124 in nRF52840-PS
+impl Packet {
+    // for indexing purposes
+    const PHY_HDR: usize = 0;
+    const DATA: RangeFrom<usize> = 1..;
+
+    /// Maximum amount of usable payload (CRC excluded) a single packet can contain, in bytes
+    pub const CAPACITY: u8 = 125;
+    const CRC: u8 = 2; // size of the CRC, which is *never* copied to / from RAM
+    const MAX_PSDU_LEN: u8 = Self::CAPACITY + 2 /* Self::CRC */;
```
hubertmis, line 593:
```suggestion
    const MAX_PSDU_LEN: u8 = Self::CAPACITY + Self::CRC;
```
I'm not rust expert yet and I'm not sure if this syntax would cause any problems, but it looks much simpler to replace a magic number and a comment with the constant mentioned by the comment.

hubertmis:
@japaric , to verify CRC of generated frames, you can easily use https://github.com/NordicSemiconductor/nRF-Sniffer-for-802.15.4 . Firmware used in this project reports only frames with valid CRC.

nrf52840-hal/src/ieee802154.rs
japaric: ah, yes. I meant to use the constant here. Thanks for catching this.

hubertmis:
I was thinking about timing out while receiving. Current implementation blocks execution of the calling thread until a frame is received or a timeout fires. It's not good for the system performance.

In drivers in C usually asynchronous approach is implemented: function like `recv` changes the driver state to 'receive' and stops execution. If the driver receives a frame, it calls a callback asynchronously. But it looks that this asynchronous callback approach is not perfect fit for Rust, because I think it does not go well with futures or programs using async/await.

Recently I've spotted https://github.com/rust-embedded/nb and it looks like a solution to this problem. The user of the driver could call `recv()` just to enter receive state. Later the user could call a new function like `read()` that would implement this nb scheme and return a frame if it was received before the function was called, or `WouldBlock` if the RX buffer is empty. With such API, the calling thread would not be blocked while waiting for frames and futures, async/await, or other models could be used.

What do you think about it?

nrf52840-hal/src/ieee802154.rs
```diff
@@ -0,0 +1,804 @@
...
+
+                // note that since we have increased EDCNT, the EDSAMPLE register contains the
+                // maximumrecorded value, not the average
+                let read_lvl = self.radio.edsample.read().edlvl().bits();
+                return read_lvl;
+            }
+        }
+    }
+
+    /// Recevies one radio packet and copies its contents into the given `packet` buffer
```
jonas-schievink, line 304:
```suggestion
    /// Receives one radio packet and copies its contents into the given `packet` buffer
```

nrf52840-hal/src/ieee802154.rs
```diff
@@ -0,0 +1,804 @@
...
+    /// Changes the TX power
+    pub fn set_txpower(&mut self, power: TxPower) {
+        self.needs_enable = true;
+        self.radio
+            .txpower
+            .write(|w| w.txpower().variant(power._into()));
+    }
+
+    /// Sample the received signal power (i.e. the presence of possibly interfering signals)
+    /// within the bandwidth of the currently used channel for sample_cycles iterations.
```
jonas-schievink, line 263:
```suggestion
    /// within the bandwidth of the currently used channel for `sample_cycles` iterations.
```

nrf52840-hal/src/ieee802154.rs
```diff
@@ -0,0 +1,804 @@
...
+            .tasks_edstart
+            .write(|w| w.tasks_edstart().set_bit());
+
+        loop {
+            if self.radio.events_edend.read().events_edend().bit_is_set() {
+                // sampling period is over; collect value
+                self.radio.events_edend.reset();
+
+                // note that since we have increased EDCNT, the EDSAMPLE register contains the
+                // maximumrecorded value, not the average
```
jonas-schievink, line 297:
```suggestion
                // maximum recorded value, not the average
```

nrf52840-hal/src/ieee802154.rs
```diff
@@ -0,0 +1,804 @@
...
+    /// Maximum amount of usable payload (CRC excluded) a single packet can contain, in bytes
+    pub const CAPACITY: u8 = 125;
+    const CRC: u8 = 2; // size of the CRC, which is *never* copied to / from RAM
+    const MAX_PSDU_LEN: u8 = Self::CAPACITY + Self::CRC;
+    const SIZE: usize = 1 /* PHR */ + Self::MAX_PSDU_LEN as usize;
+
+    /// Returns an empty packet (length = 0)
+    // XXX I believe that be making this *not* `const` it is not possible to place a `Packet` in
+    // `.rodata` (modulo unsafe `#[link_section]` shenanigans) thus it is not necessary to check the
+    // address of packet in `Radio.{send,recv}` (EasyDMA can only operate on RAM addresses)
```
jonas-schievink, line 741: Hmm, I think I'd prefer not to rely on this. Instead, `recv` and `send` could both take a `&mut Packet`, which ensures that it's in writeable RAM (other modules in the HAL already skip the address check when they deal with mutable data).

nrf52840-hal/src/ieee802154.rs
```diff
@@ -0,0 +1,804 @@
...
+    needs_enable: bool,
+    // used to freeze `Clocks`
+    _clocks: &'c (),
+}
+
+/// Default Clear Channel Assessment method = Carrier sense
+pub const DEFAULT_CCA: Cca = Cca::CarrierSense;
+
+/// Default radio channel = Channel 20 (`2_450` MHz)
+pub const DEFAULT_CHANNEL: Channel = Channel::_18; // _17 todo undo; set this to interfere with my wifi
```
jonas-schievink, line 33: It looks like this needs to be reset to channel 17.

Are the defaults specified somewhere or were they picked arbitrarily?

nrf52840-hal/src/ieee802154.rs
```diff
@@ -0,0 +1,804 @@
...
+    timer::{self, Timer},
+};
+
+/// IEEE 802.15.4 radio
+pub struct Radio<'c> {
+    radio: RADIO,
+    // RADIO needs to be (re-)enabled to pick up new settings
+    needs_enable: bool,
+    // used to freeze `Clocks`
+    _clocks: &'c (),
```
jonas-schievink, line 26: This could be `PhantomData` to shrink `Radio` by a pointer.

nrf52840-hal/src/ieee802154.rs
japaric: the FREQUENCY register reset value is `2`, which doesn't match to any of the `Channel` values because they are all multiples of 5.
The channel values (the frequencies) are the standard IEEE frequencies.

japaric: changed the default channel to 11 (2405 MHz) which is closer to the reset value of the FREQUENCY register (2402 MHz)

jonas-schievink:
Okay, I think this is ready to go. The remaining notes can be addressed later, and an async/non-blocking API can also be added then.

bors r+

jonas-schievink:
I only just noticed that the module was added directly to the nrf52840-hal instead of nrf-hal-common. I've opened https://github.com/nrf-rs/nrf-hal/pull/299 to move it (and enable it for the nRF52833 in the process).

PR #144, opened 2020-05-11
USB support

japaric:
this PR builds on top of @jonas-schievink  PR (#133)

I have managed to get a CDC ACM (using the `usbd-serial` crate) device enumerated and got some output out from it using the following program (snippet):

``` rust
    loop {
        if !usb_dev.poll(&mut [&mut serial]) {
            continue;
        }

        if usb_dev.state() == UsbDeviceState::Configured {
            if once {
                serial.write(b"Hello, world!\n").unwrap();
                serial.flush().unwrap();

                once = false;
            }
        }
    }
```

``` console
$ cat /dev/ttyACM1
Hello, world!
```

(don't forget to send a newline or `cat` won't print anything)

However, I'm not getting reliable enumeration (*). I still need to dig deeper into that issue but would appreciate other people testing out this branch.

(*) though this may be mainly caused by *how* I'm terminating my embedded application: I'm halting the device without first disconnecting it from the USB bus; this seems to make the host USB controller very sad (it power cycles the USB bus a few times and never gets an answer from the device (because it's halted))

P.S. this branch depends on this Jonas' fork of usbd-device (which handles the nrf52 quirk of handling SET_ADDRESS requests entirely in hardware). Use the following patch before you try out this branch:

``` toml
[patch.crates-io.usb-device]
git = "https://github.com/jonas-schievink/usb-device.git"
branch = "inhibit-setaddr-resp"
```

### DIFF
--- a/nrf-hal-common/src/usbd/mod.rs
+++ b/nrf-hal-common/src/usbd/mod.rs
@@ -95,11 +95,29 @@ impl Usbd {
         unsafe { &*USBD::ptr() }.usbaddr.read().addr().bits()
     }
 
-    fn alloc_ep_buf(&mut self, size: u16) -> usb_device::Result<&'static mut [u8]> {
-        assert!(size <= 64);
+    fn alloc_ep_buf(
+        &mut self,
+        ep_type: EndpointType,
+        mut size: u16,
+    ) -> usb_device::Result<&'static mut [u8]> {
         if self.unalloc_buffers.len() < usize::from(size) {
             Err(UsbError::EndpointMemoryOverflow)
         } else {
+            if ep_type == EndpointType::Bulk || ep_type == EndpointType::Interrupt {
+                // datasheet: buffer must be 4-byte aligned and its size must be a multiple of 4
+                let rem = self.unalloc_buffers.as_mut_ptr() as usize % 4;
+                if rem != 0 {
+                    let (_padding, remaining) =
+                        mem::replace(&mut self.unalloc_buffers, &mut []).split_at_mut(4 - rem);
+                    self.unalloc_buffers = remaining;
+                }
+
+                let rem = size % 4;
+                if rem != 0 {
+                    size = size + 4 - rem;
+                }
+            }
+            assert!(size <= 64);
             let (alloc, remaining) =
                 mem::replace(&mut self.unalloc_buffers, &mut []).split_at_mut(size.into());
             self.unalloc_buffers = remaining;
@@ -183,7 +201,7 @@ impl UsbBus for Usbd {
 
         // Endpoint directions are allocated individually.
 
-        let buf = self.alloc_ep_buf(max_packet_size)?;
+        let buf = self.alloc_ep_buf(ep_type, max_packet_size)?;
 
         if false {
             unimplemented!(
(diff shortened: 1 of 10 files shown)


Disasm:
We have a test repo with a few tests that you can port and run on your hardware: https://github.com/Disasm/stm32-usbd-tests

Disasm:
For some reason, this driver doesn't work with the default EP0 size (8 bytes). Also a serial loopback example (based on https://github.com/Disasm/usb-otg-workspace/blob/master/example-f429zi-board/examples/serial.rs#L63-L85) fails to pass stress-testing.

myfreeweb:
Rebasing the `usb-device` quirk onto the current (0.2.7) version, looks like https://github.com/mvirkkunen/usb-device/commit/7509c776cf5d33e6aa8b8096e04e94830fb06034 breaks device enumeration, I've had to revert that commit to make it work.

nrf-hal-common/src/usbd/mod.rs
```diff
@@ -0,0 +1,746 @@
...
+        })
+    }
+
+    fn set_stalled(&self, ep_addr: EndpointAddress, stalled: bool) {
+        interrupt::free(|cs| {
+            let regs = self.periph.borrow(cs);
+
+            unsafe {
+                if ep_addr.index() == 0 {
+                    regs.tasks_ep0stall.write(|w| w.tasks_ep0stall().set_bit());
```
myfreeweb, line 575: Oh.

```rust
                    regs.tasks_ep0stall.write(|w| if stalled { w.tasks_ep0stall().set_bit() } else { w.tasks_ep0stall().clear_bit() });
```

wez:
@nrf-rs/nrf52 I've rebased this and plan to push this PR + my local fixup diff to close out this PR.

I have a few "Particle Xenon" boards with 52840's on them, as well as that weird MDK thing that I previously added a bsp for in this org, so I'm able to test this and confirmed that the device enumerates and responds:

```
[72570.386287] usb 3-3.4.4.2: New USB device strings: Mfr=0, Product=2, SerialNumber=0
[72570.386289] usb 3-3.4.4.2: Product: nRF52840 Serial Port Demo
[72570.472123] cdc_acm 3-3.4.4.2:1.0: ttyACM0: USB ACM device
[72570.472567] usbcore: registered new interface driver cdc_acm
[72570.472569] cdc_acm: USB Abstract Control Model driver for USB modems and ISDN adapters
```

```
$ sudo cat /dev/ttyACM0
Hello, world!
^C
```

jonas-schievink:
Awesome! Happy to take a look.

wez:
My fixup diff on top of this is; I'll just push it and if you spot something I'm around to act on it!

nrf-hal-common/src/usbd/mod.rs
wez: I went with:

```rust
  regs.tasks_ep0stall.write(|w| w.tasks_ep0stall().bit(stalled));
```

jonas-schievink:
Please wait for CI to pass before pushing to master

wez:
[Too late](https://github.com/nrf-rs/nrf-hal/runs/1899200248?check_suite_focus=true)... but I'm running those tests locally and fixing up now :)

Disasm:
I'm going to run test_class as well as the serial benchmark soon

Disasm:
Tested, the behavior is the same as mentioned in https://github.com/nrf-rs/nrf-hal/pull/144#issuecomment-668598080

PR #145, opened 2020-05-16
AES electronic codebook implementation and example

thalesfragoso:
Pretty straight forward implementation.

There's just one thing that I want a confirmation, I'm not checking if the data structure is in RAM since neither the method for creating `EcbData` nor `Ecb` are `const`, am I wrong in assuming that `EcbData` will always be in RAM ?


thalesfragoso:
No fmt on nightly :/

Yatekii:
> There's just one thing that I want a confirmation, I'm not checking if the data structure is in RAM since neither the method for creating `EcbData` nor `Ecb` are `const`, am I wrong in assuming that `EcbData` will always be in RAM ?

Well, it will always be moved to RAM during runtime anyways, because access in flash surely is slower, no?

jonas-schievink:
Also see: https://github.com/nrf-rs/nrf51-hal/blob/master/src/ecb.rs

(looks like it was missed by the port)

thalesfragoso:
> Well, it will always be moved to RAM during runtime anyways, because access in flash surely is slower, no?

It has to be in RAM because of the EasyDMA, otherwise it can cause hardfaults or memory corruptions.

@jonas-schievink Oh, I didn't know about that. Do you prefer that API ? It seems more straight forward than mine, but with more copying (possibly). Anyway, I should also add a method to change the key.

jonas-schievink:
The nRF51 API has no persistent memory usage, that part as least seems preferable to me.

thalesfragoso:
> The nRF51 API has no persistent memory usage, that part as least seems preferable to me.

I changed the API, let me know what you think.

nrf-hal-common/src/ecb.rs
```diff
@@ -5,76 +5,73 @@
...
-        let data_ptr = &mut self.data as *mut _ as u32;
+    /// Encrypts a `block` with `key`.
+    ///
+    /// # Errors
+    ///
+    /// An error will be returned when the AES hardware raises an `ERRORECB`
+    /// event. This can happen when an operation is started that shares the AES
+    /// hardware resources with the AES ECB peripheral while an encryption
+    /// operation is running.
+    pub fn crypt_block(
```
jonas-schievink, line 51:
```suggestion
    pub fn encrypt_block(
```

thalesfragoso: Done

examples/ecb-demo/README.md
```diff
@@ -0,0 +1,19 @@
+# AES electronic codebook mode encryption demo
+
+Chose the microcontroller with one of the following features:
```
jonas-schievink, line 3:
```suggestion
Choose the microcontroller with one of the following features:
```

examples/ecb-demo/Embed.toml
```diff
@@ -0,0 +1,46 @@
...
+timeout = 3000
+# Whether timestamps in the RTTUI are enabled
+show_timestamps = true
+
+[gdb]
+# Whether or not a GDB server should be opened after flashing.
+# This is exclusive and cannot be used with RTT at the moment.
+enabled = false
+# The connection string in host:port format wher the GDB server will open a socket.
+# gdb_connection_string
```
jonas-schievink, line 46: Is it normal to check this file into git?

examples/ecb-demo/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+[[bin]]
+name = "ecb-demo"
+doc = false
+test = false
+
+[features]
+51 = ["nrf51-hal"]
+52810 = ["nrf52810-hal"]
+52832 = ["nrf52832-hal"]
+52840 = ["nrf52840-hal"]
```
jonas-schievink, line 26: We now have support for the 52833 as well

jonas-schievink: Also the new HAL needs to reexport this

nrf-hal-common/src/ecb.rs
```diff
@@ -0,0 +1,98 @@
+//! HAL interface to the AES electronic codebook mode encryption
+//!
+//! The ECB encryption block supports 128 bit AES encryption (encryption only, not decryption).
+
+use crate::target::ECB;
+use core::sync::atomic::{compiler_fence, Ordering};
+
+/// Error type to represent a sharing conflict during encryption
```
jonas-schievink, line 8:
```suggestion
/// Error type to represent a sharing conflict during encryption.
```

nrf-hal-common/src/ecb.rs
```diff
@@ -0,0 +1,98 @@
+//! HAL interface to the AES electronic codebook mode encryption
```
jonas-schievink, line 1:
```suggestion
//! HAL interface to the AES electronic codebook mode encryption.
```

examples/ecb-demo/Embed.toml
Yatekii: Yes, this file is meant to be checked into git, assuming you/your users use cargo embed :)

PR #147, opened 2020-05-19
Add Thingy:91 BSP

jonathanpallant:
I know there was talk of moving the BSPs somewhere else, but in the mean time, here's one for the Thingy:91 board.


Yatekii:
Nice! Need to try this out :)

I would prefer this to be in a separate repo in the org straightaway and make a separate repo for all the other crates too.

If we don't start now, we wont do it (maybe not even if we start now :D)

jonathanpallant:
OK, see nrf-bsp-template.

Yatekii:
Great work, thanks! Looking good. I'll try and port my BSP to a new repo following the readme template, thanks!

PR #148, opened 2020-05-23
Adding support for Nordic nRF52833

blueluna:
This PR adds support for the Nordic nRF52833. This chip is similar to the nRF52840 with less flash and memory, no CryptoCell, some radio differences and probably som other smaller differences.

I have tested using clocks, gpio, timer and uarte in a RTFM program running on a Nordic nRF52833-DK.


nrf-hal-common/Cargo.toml
```diff
@@ -44,6 +44,12 @@ version = "0.9.0"
 optional = true
 version = "0.9.0"
 
+
+[dependencies.nrf52833-pac]
+optional = true
+version = "0.9.0"
+git = "https://github.com/nrf-rs/nrf52833-pac.git"
```
jonas-schievink, line 51: This should be published to crates.io before merge

chocol4te:
Is there a good pattern for solving `#[cfg]` hell?

jonas-schievink:
I could imagine using a build.rs script to set various `cfg`s based on the enabled features. Other than that and `cfg_if`, I don't think there are any good solutions.

Yatekii:
When you're talking about cfg-hell you mean from a developer perspective, right? Because from a user perspective I like it quite much :)

nrf-hal-common/Cargo.toml
blueluna: I guessed that was true, decided to get this out there to get comments.

blueluna:
With regards to cfg-hell. I experimented a bit with adding a build.rs that generated features from the features defined in Cargo.toml. I am on the fence on if the result is better or not. Maybe open a issue for cfg opportunity for improvement?

```
impl Instance for SPIM0 {}

#[cfg(feature = "peripheral_spim1_enable")]
impl Instance for SPIM1 {}

#[cfg(feature = "peripheral_spim2_enable")]
impl Instance for SPIM2 {}

#[cfg(feature = "peripheral_spim3_enable")]
impl Instance for SPIM3 {}
```

build.rs
```
use std::env;

fn is_feature_enabled(name: &str) -> bool {
    match env::var(format!("CARGO_FEATURE_{}", name.to_uppercase())) {
        Ok(val) => {
            if val == "1" {
                true
            } else {
                false
            }
        }
        Err(_) => false,
    }
}

fn get_chip_code() -> u32 {
    if is_feature_enabled("51") {
        51
    } else if is_feature_enabled("52810") {
        52810
    } else if is_feature_enabled("52811") {
        52811
    } else if is_feature_enabled("52820") {
        52820
    } else if is_feature_enabled("52832") {
        52832
    } else if is_feature_enabled("52833") {
        52833
    } else if is_feature_enabled("52840") {
        52840
    } else if is_feature_enabled("5340") {
        5340
    } else if is_feature_enabled("9160") {
        9160
    } else {
        0
    }
}

fn spim_count(chip_code: u32) -> u32 {
    match chip_code {
        52833 | 52840 => 4,
        52832 => 3,
        _ => 1,
    }
}

fn emit_feature(value: &str) {
    println!("cargo:rustc-cfg=feature=\"{}\"", value);
}

fn main() {
    let chip_code = get_chip_code();
    for n in 0..spim_count(chip_code) {
        emit_feature(&format!("peripheral_spim{}_enable", n));
    }
}
```

PR #153, opened 2020-05-26
Set up GitHub Actions

jonas-schievink:


scripts/build.sh
```diff
@@ -41,5 +41,3 @@ echo Building examples/twi-ssd1306...
 cargo build --manifest-path examples/twi-ssd1306/Cargo.toml
 echo Building examples/twi-ssd1306...
 cargo build --manifest-path examples/twi-ssd1306/Cargo.toml --no-default-features --features="52840" --target thumbv7em-none-eabi
-echo Checking source code formatting...
-cargo fmt -- --check
```
hannobraun, line 45: I think it would be better to leave this. Otherwise a contributor is less likely to catch formatting problems before submitting the PR, and after submitting the PR, they won't immediately be notified either (since CI takes a while). That would increase the back-and-forth required to get a pull request ready, which is more work for the contributor and increases the chance of abandoned pull requests.

jonas-schievink: True, but we must only check this on stable, not the other channels. Is there a simple way to implement that? I guess I can pass an argument to the script or something...

jonathanpallant: `cargo +stable fmt` ?

hannobraun: > I guess I can pass an argument to the script or something...

We're doing this check in `lpc8xx-hal`, but that's still using Travis: https://github.com/lpc-rs/lpc8xx-hal/blob/master/scripts/build.sh#L15-L18

But I like @jonathanpallant's solution. Hopefully that works.

PR #155, opened 2020-06-06
Clean up gpio.rs

### LINKED ISSUE #146: `Pin`'s public fields let you break `Pin`'s singleton property
For example:

``` rust
let p0 = p0::Parts(p.P0);
let a = p0.p0_00.degrade();
let mut b = p0.p0_01.degrade();
b.pin = 0;
```

Lets you create 2 P0.00 pins. I don't think this operation is UB per se but it does break the intended (singleton) semantics and can result in IO misconfiguration, e.g. configuring P0.00 both as serial Tx and Rx if you pass the `a` and `b` from above to the `Uarte` abstraction.

Setting the public `pin` field to value equal or greater than `32` may result in out of bounds array access though (either a panic or UB if the HAL is using unchecked bounds access).

I think making the fields private and instead providing getter methods for `pin` and `port` would be a better fit here.

jonas-schievink:
Fixes #146

This compacts the generic `Pin` struct to a single byte on all MCUs, encoding the port in the upper bit.

Breaking change.

### DIFF
--- a/nrf-hal-common/src/gpio.rs
+++ b/nrf-hal-common/src/gpio.rs
@@ -1,7 +1,3 @@
-// TODO - clean these up
-#![allow(unused_imports)]
-#![allow(non_camel_case_types)]
-
 use core::marker::PhantomData;
 
 /// Input mode (type state)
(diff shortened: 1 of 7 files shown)


jamesmunns:
If we're doing a breaking change, we might want to consider @hannobraun's suggestion to reduce the macro surface area, instead moving to a more `Instance` style way of doing things. We can do that in another PR though.

Yatekii:
How is this performance wise? When I initially implemented this (I did the initial 840 impl) I thought that it rather save a few CPU cycles which are more important than the bit of memory we can save.
Do you think the other way round or do you think this has no perf implications anyways?

jonas-schievink:
This should have no performance implication on any chip with only one GPIO bank, and on the others most use cases should only be affected if the inliner fails. I'm happy to remove that change though, if anyone feels strongly.

Yatekii:
I have no strong feelings. I was just curious what the actual implications might be. I had to decide what to do initially and picked the one I thought was a better fit, without any proof or strong oppinion :)

This change definitely seems ok to me.

hannobraun:
> If we're doing a breaking change, we might want to consider @hannobraun's suggestion to reduce the macro surface area, instead moving to a more `Instance` style way of doing things. We can do that in another PR though.

For reference, James is referring to #8, I believe.

PR #157, opened 2020-06-13
Release the pins held by comm drivers

### LINKED ISSUE #94: Uarte::free() doesn't give back pins
Right now, it is not possible to obtain back the pins used to create a Uarte.

bugadani:
As an alternative, if this causes excess amount of memory usage, the peripherials hold the pin numbers that could be used to conjure up the pins on release.

This is a breaking change for anyone who uses `free()`

Closes #94

### DIFF
--- a/nrf-hal-common/src/spi.rs
+++ b/nrf-hal-common/src/spi.rs
@@ -13,7 +13,7 @@ pub use embedded_hal::{
 pub use spi0::frequency::FREQUENCY_A as Frequency;
 
 /// Interface to a SPI instance
-pub struct Spi<T>(T);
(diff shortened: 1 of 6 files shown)


jonas-schievink:
Since this is a somewhat rarely used feature, I'd prefer to not incur this size increase by default. We can probably solve this by making peripherals take another generic argument instead, which defaults to some ZST, but can be set to `Pins` if the user wants to reuse the pins.

bugadani:
Fair enough :)

jonas-schievink:
Yeah, we have those too, but it is somewhat easier for peripherals to deal with the erased pin type. I'm open to changing that though.

PR #159, opened 2020-06-13
Fix incorrect logic in transfer_split_uneven

### LINKED ISSUE #156: SPIM write sends only logic low
SPIM::transfer_split_uneven seems to send only logic low when the rx_buffer is a Zero Sized Slice. Could it be because the Rust compiler is optimizing out operations on the zero sized type? I'm still new to this but replicating the first part of the transfer_split_uneven function on my desktop compiled to no output.

Full code below (running on a nRF52-DK). If i substitute read() with a zero sized array, for the write() I get the same result:

```rust
#![no_std]
#![no_main]

extern crate panic_halt;
extern crate nrf52832_hal;
extern crate embedded_hal_spy;

use cortex_m_rt::entry;
use nrf52832_hal::gpio::{self, *,  p0::*,  Level};
use nrf52832_hal::spim::Spim;


#[entry]
fn main() -> ! {

    let p = nrf52832_hal::pac::Peripherals::take().unwrap();
    let port0 = p0::Parts::new(p.P0);
    
    let cs: P0_12<gpio::Output<PushPull>> = port0.p0_12.into_push_pull_output(Level::High);
    let spiclk = port0.p0_24.into_push_pull_output(Level::Low).degrade();
    let spimosi = port0.p0_23.into_push_pull_output(Level::Low).degrade();
    let spimiso = port0.p0_22.into_floating_input().degrade();

    let pins = nrf52832_hal::spim::Pins {
        sck: spiclk,
        miso: Some(spimiso),
        mosi: Some(spimosi),
    };
    let mut spi = Spim::new(
        p.SPIM1,
        pins,
        nrf52832_hal::spim::Frequency::K500,
        nrf52832_hal::spim::MODE_0,
        0,
    );

    let data: [u8; 1] = [0b1101_0010];
    //let mut readbuf: [u8; 0];
    //spi.write(&mut cs.degrade(), &data).unwrap();
    spi.read(&mut cs.degrade(), &data, &mut [0u8; 0]).unwrap();

    loop {}
}
```

bugadani:
This seems to be a classic case of comment vs code. Both iterators are finished if both of them return None, not one of them.

Maybe Closes #156

### DIFF
--- a/nrf-hal-common/src/spim.rs
+++ b/nrf-hal-common/src/spim.rs
@@ -340,7 +340,7 @@ where
         // Don't return early, as we must reset the CS pin
         let res = txi
             .zip(rxi)
-            .take_while(|(t, r)| t.is_some() && r.is_some())
+            .take_while(|(t, r)| t.is_some() || r.is_some())
             // We also turn the slices into either a DmaSlice (if there was data), or a null
             // DmaSlice (if there is no data)
             .map(|(t, r)| {


jonas-schievink:
Thanks! Looks reasonable to me. @bartekkowalski can you check if this fixes your issue in #156?

bartekkowalski:
@jonas-schievink  Works great, thanks guys!

Issue #150, opened 2020-05-25
Make sure all embedded-hal interfaces can implicitly copy


jamesmunns:
For compat reasons, we should support implicit copy (if the user passes in a slice from `.text`) of chunks on all interfaces. We currently do this for UARTE, but not for SPIM/TWIM.

kalkyl:
Instead of returning Error::DMABufferNotInDataMemory?

Yatekii:
I think the majority vote was in favor of returning an Error on the main interface but providing a call which does implicit copy if you ask it to.
But the interfaces should all have the same behavior/look :)

jamesmunns:
The goal was:

1. When using `embedded-hal` traits, implicitly copy if necessary (when we would otherwise return `DMABufferNotInDataMemory`)
2. When using `nrf-hal` direct interfaces, do not implicitly copy, instead return the error

This is because some (non nrf specific) drivers send constant strings, which will end up in `.text`.

kalkyl:
I had a got at it in PR #163 for spim transfer, if i got it right :)

kalkyl:
Oups, i messed up my first PR and accidently pushed the wrong branch, still getting used to the workflow of contributing, haha! Anyways, i now submitted the one i was supposted to (PR #164) for spim transfer :)

kalkyl:
Added Twim write/write_read aswell and made a new PR #165 with both Spim and Twim instead, sorry for the PR mess, i'm new here... :)

(Not sure why the githud build failed (##[error]Process completed with exit code 1), is that a common issue or is it me? It builds just fine on my machine...)

PR #165, opened 2020-06-24
Twim: Implicitly copy buffer into RAM if needed when using embedded hal traits

kalkyl:
Issue #150

### DIFF
--- a/nrf-hal-common/src/twim.rs
+++ b/nrf-hal-common/src/twim.rs
@@ -380,9 +380,9 @@ where
             unsafe { w.maxcnt().bits(rx_buffer.len() as _) });
 
         // Chunk write data
-        for chunk in tx_buffer.chunks(FORCE_COPY_BUFFER_SIZE) {
+        let wr_buffer = &mut [0; FORCE_COPY_BUFFER_SIZE][..];
+        for chunk in tx_buffer.chunks(FORCE_COPY_BUFFER_SIZE.min(EASY_DMA_SIZE)) {
             // Copy chunk into RAM
-            let wr_buffer = &mut [0; FORCE_COPY_BUFFER_SIZE][..];
             wr_buffer[..chunk.len()].copy_from_slice(chunk);
 
             // Set up the DMA write
--- a/nrf-hal-common/src/twim.rs
+++ b/nrf-hal-common/src/twim.rs
@@ -18,8 +18,7 @@ use crate::target::TWIM1;
 
 use crate::{
     gpio::{Floating, Input, Pin},
-    slice_in_ram_or,
-    slice_in_ram,
+    slice_in_ram, slice_in_ram_or,
     target_constants::{EASY_DMA_SIZE, FORCE_COPY_BUFFER_SIZE},
 };
 
@@ -349,10 +348,9 @@ where
     type Error = Error;
 
     fn write<'w>(&mut self, addr: u8, bytes: &'w [u8]) -> Result<(), Error> {
-        if slice_in_ram(bytes){
+        if slice_in_ram(bytes) {
             self.write(addr, bytes)
-        }
-        else {
+        } else {
             let buf = &mut [0; FORCE_COPY_BUFFER_SIZE][..];
             for chunk in bytes.chunks(FORCE_COPY_BUFFER_SIZE) {
                 buf[..chunk.len()].copy_from_slice(chunk);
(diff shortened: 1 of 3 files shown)


Yatekii:
Could you be so kind and run `cargo fmt` such that the CI is happy?

nrf-hal-common/src/twim.rs
```diff
@@ -333,6 +334,122 @@ where
...
+            // is also fine.
+            //
+            // Note that that nrf52840 maxcnt is a wider
+            // type than a u8, so we use a `_` cast rather than a `u8` cast.
+            // The MAXCNT field is thus at least 8 bits wide and accepts the
+            // full range of values that fit in a `u8`.
+            unsafe { w.maxcnt().bits(rx_buffer.len() as _) });
+
+        // Chunk write data
+        let wr_buffer = &mut [0; FORCE_COPY_BUFFER_SIZE][..];
```
Yatekii, line 383:
```suggestion
        let wr_buffer = &mut [0; FORCE_COPY_BUFFER_SIZE.min(EASY_DMA_SIZE)][..];
```

I think we should take the minimum here too? Or ensure elseplace that `FORCE_COPY_BUFFER_SIZE` is always <= `EASY_DMA_SIZE`. Maybe we should do that where we define `FORCE_COPY_BUFFER_SIZE` even?

nrf-hal-common/src/twim.rs
```diff
@@ -333,6 +334,122 @@ where
...
+        self.0.tasks_startrx.write(|w|
+            // `1` is a valid value to write to task registers.
+            unsafe { w.bits(1) });
+
+        // Wait until read operation is about to end
+        while self.0.events_lastrx.read().bits() == 0 {}
+        self.0.events_lastrx.write(|w| w); // reset event
+
+        // Stop read operation
+        self.0.tasks_stop.write(|w|
```
Yatekii, line 432: Doesn't it stop automatically when the requested data size was received?

nrf-hal-common/src/twim.rs
kalkyl: Yes, i don't know how to do that min check in const though?
I guess ensuring `FORCE_COPY_BUFFER_SIZE <= EASY_DMA_SIZE` where we define `FORCE_COPY_BUFFER_SIZE` is the best solution, then we can just use `FORCE_COPY_BUFFER_SIZE` for our buffer and chunk sizes?

nrf-hal-common/src/twim.rs
kalkyl: No, according to the manual:

> "Note that the TWI master does not stop by itself when the RAM buffer is full, or when an error occurs. The STOP task must be issued, through the use of a local or PPI shortcut, or in software as part of the error handler."

So i guess an alternative could be to add a PPI shortcut instead, like:
`self.0.shorts.modify(|_r, w| w.lastrx_stop().enabled());`

What do you prefer?
Thanks for your comments :)

Yatekii: ok, this seems odd to me tbh, but I guess if the manual states this, it must be true :)

I think the short is cleaner but it also occupies a short slot which could be used by the user for another job. So I think we should leave it as is.

nrf-hal-common/src/twim.rs
Yatekii: Ok, I think we should use .min() at both places or at no place in this code then :) Otherwise it's just confusing I think.

nrf-hal-common/src/twim.rs
kalkyl: Cool! :)

nrf-hal-common/src/twim.rs
kalkyl: Totally agree, so i will use `FORCE_COPY_BUFFER_SIZE` in both places then :)

How can i assert `FORCE_COPY_BUFFER_SIZE <= EASY_DMA_SIZE` when they are defined in "top-level"?

Yatekii: I am not sure we can ensure this at compile time other than manually making sure of this and maybe add a comment.

kalkyl: The only hacky way i can think of right now is not pretty and not zero cost, something like:
`const _CHECK_FORCE_COPY_BUFFER_SIZE: usize = EASY_DMA_SIZE - FORCE_COPY_BUFFER_SIZE;`
`// ERROR: FORCE_COPY_BUFFER_SIZE must be <= EASY_DMA_SIZE`
..that would fail with "attempt to subtract with overflow" :)
But i guess just a comment should be enough right?

Yatekii: Wont it just optimize that out and thus be zero cost?

kalkyl: Ahh yes true!

PR #167, opened 2020-07-02
Add Gpiote module (open for feedback)

kalkyl:
Hi nRF team!

Here's a starting point of a Gpiote module, based on how i've been using it in my projects.
Let's discuss a suitable api, at the moment it looks like this:

`let gpiote = gpiote::Gpiote::new(p.GPIOTE);`

Example how to configure an input pin as an event source for channel 0, triggering on hi-to-low transition and firing the GPIO interrupt:
`let btn1 = p0.p0_11.into_pullup_input().degrade();`
`gpiote.channel0().input_pin(&btn1).hi_to_lo().enable_interrupt();`

Or using port event:
`gpiote.port().input_pin(&btn1).low();`
`gpiote.port().input_pin(&btn3).low();`
`gpiote.port().enable_interrupt();`

In the GPIOTE ISR, check for triggered events like:
`gpiote.channel0().is_event_triggered()`
 Reset channel events like:
`gpiote.channel0().reset_events();`
Reset port events like
`gpiote.port().reset_events();`
Or globally reset all events like:
`gpiote.reset_events();`

Task output pins are configured like:
`let led1 = p0.p0_13.into_push_pull_output(Level::High).degrade();`
`gpiote.channel1().output_pin(led1).init_high();`

The assigned output pin is consumed because when an output pin is in task mode, it can no longer be used with the normal gpio api, instead you would use the provided task functions `set()`, `clear()` and `out()`.

All configuations available in the registers are implemented aswell as far as i can tell.

Here's a demo using interrupts and PPI with the onboard buttons and leds of the nrf52840-DK:
https://github.com/kalkyl/nrf-hal/blob/gpiote/examples/gpiote-demo/src/main.rs

Cheers!

EDIT: Updated with the latest version of the api

### DIFF
--- a/nrf-hal-common/src/gpiote.rs
+++ b/nrf-hal-common/src/gpiote.rs
@@ -98,7 +98,7 @@ impl<'a> GpioteChannel<'_> {
         }
     }
 
-    pub fn output_pin<P: GpioteOutputPin>(&'a self, pin: &'a P) -> GpioteTask<'a, P> {
+    pub fn output_pin<P: GpioteOutputPin>(&'a self, pin: P) -> GpioteTask<'a, P> {
         GpioteTask {
             gpiote: &self.gpiote,
             pin: pin,
@@ -274,7 +274,7 @@ fn config_port_event_pin<P: GpioteInputPin>(pin: &P, sense: PortEventSense) {
 
 pub struct GpioteTask<'a, P: GpioteOutputPin> {
     gpiote: &'a GPIOTE,
-    pin: &'a P,
+    pin: P,
     channel: usize,
     task_out_polarity: TaskOutPolarity,
 }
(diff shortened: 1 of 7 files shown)


kalkyl:
Added an example for the nRF52840-DK with both interrupt- and PPI usage with the onboard buttons and leds:
https://github.com/kalkyl/nrf-hal/blob/gpiote/examples/gpiote-demo/src/main.rs

nrf-hal-common/src/gpiote.rs
```diff
@@ -153,41 +188,44 @@ pub struct GpioteChannelEvent<'a, P: GpioteInputPin> {
...
         unsafe {
             self.gpiote
                 .intenset
                 .modify(|r, w| w.bits(r.bits() | self.pin.pin() as u32))
         }
+        self
+    }
+
+    pub fn disable_interrupt(&self) -> &Self {
+        // Disable interrupt for pin
```
jonas-schievink, line 222: This can just be promoted to a doc comment (and above as well)

kalkyl: Thanks, i will make sure to go through and add doc comments all over :)

jonas-schievink:
Looks good, thanks! I'll leave this open for a while to see if anyone else has opinions.

PR #168, opened 2020-07-10
Add methods for task clear and trigger overflow to the RTC

mariusknaust:
Expose missing tasks of the RTC, according to: https://infocenter.nordicsemi.com/index.jsp?topic=%2Fcom.nordic.infocenter.nrf52832.ps.v1.1%2Frtc.html

### DIFF
--- a/nrf-hal-common/src/rtc.rs
+++ b/nrf-hal-common/src/rtc.rs
@@ -203,6 +203,20 @@ where
         self.periph.counter.read().bits()
     }
 
+	/// Clear the Real Time Counter
+    pub fn clear_counter(&self) {
+        unsafe {
+            self.periph.tasks_clear.write(|w| w.bits(1));
+        }
+    }
+
+    /// Trigger the Real Time Counter to overflow, mainly for testing purposes
+    pub fn trigger_overflow(&self) {
+        unsafe {
+            self.periph.tasks_trigovrflw.write(|w| w.bits(1));
+        }
+    }
+
     /// Destructure the high level interface. Does not reset any configuration made
     /// to the given RTC peripheral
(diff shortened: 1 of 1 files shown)


nrf-hal-common/src/rtc.rs
```diff
@@ -203,6 +203,20 @@ where
...
 
+	/// Clear the Real Time Counter
+    pub fn clear_counter(&self) {
+        unsafe {
+            self.periph.tasks_clear.write(|w| w.bits(1));
+        }
+    }
+
+    /// Trigger the Real Time Counter to overflow, mainly for testing purposes
+    pub fn trigger_overflow(&self) {
```
jonas-schievink, line 214: What happens if this is called while the timer is stopped? According to the infocenter article the counter would be set to 0xFFFFF0 (*not* 0xFFFFFF), so that wouldn't result in overflow - and even when it's running it wouldn't overflow immediately.

Maybe this should just document that it just sets the counter to 0xFFFFF0?

mariusknaust: Agreed. It makes the overflow occur sooner, for testing purposes. I'll change the documentation, should we stick with the name, as the task is named the same way?

jonas-schievink: Yeah that should be fine

nrf-hal-common/src/rtc.rs
```diff
@@ -203,6 +203,21 @@ where
         self.periph.counter.read().bits()
     }
 
+	/// Clear the Real Time Counter
```
jonas-schievink, line 206:
```suggestion
    /// Clear the Real Time Counter
```

PR #172, opened 2020-07-10
Stop reexporting the PAC under `target`

jonas-schievink:
Builds on https://github.com/nrf-rs/nrf-hal/pull/171

Now we only have a single reexport of the PAC, called `pac`.


Yatekii:
Is there a reasoning why it can't be called target anymore? I have neither preference but I'm curious. Also maybe we should include an alias to target still and add a deprecation message, such that it doesn't just explode for users.

jonas-schievink:
Mainly because `pac` does the same thing as `target` but has a much clearer name. I'm planning to write a changelog so users can be informed of the change there. Since migrating is very mechanical we shouldn't need a migration period.

Yatekii:
Ok, this is ok for me then :) I guess it's really much clearer :)

PR #175, opened 2020-07-12
Watchdog

jamesmunns:
Implement basic watchdog timer functionality.

~~WIP until I test this on hardware and add an example.~~ Ready to review.

### DIFF
--- a/examples/wdt-demo/Cargo.toml
+++ b/examples/wdt-demo/Cargo.toml
@@ -11,7 +11,7 @@ publish = false
 cortex-m = "0.6.2"
 cortex-m-rtic = "0.5.3"
 rtt-target = {version = "0.2.0", features = ["cortex-m"] }
-nrf52840-hal = { version = "0.10", features = ["rt"], path = "../../nrf52840-hal" }
+nrf52840-hal = { version = "0.11", features = ["rt"], path = "../../nrf52840-hal" }
 
 [dependencies.embedded-hal]
 version = "0.2.3"
--- a/examples/wdt-demo/Cargo.toml
+++ b/examples/wdt-demo/Cargo.toml
@@ -3,6 +3,7 @@ name = "wdt-demo"
 version = "0.1.0"
 authors = ["James Munns <james.munns@ferrous-systems.com>"]
 edition = "2018"
+publish = false
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
--- a/examples/wdt-demo/Cargo.toml
+++ b/examples/wdt-demo/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "wdt-demo"
+version = "0.1.0"
+authors = ["James Munns <james.munns@ferrous-systems.com>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+cortex-m = "0.6.2"
+cortex-m-rtic = "0.5.3"
+rtt-target = {version = "0.2.0", features = ["cortex-m"] }
+nrf52840-hal = { version = "0.10", features = ["rt"], path = "../../nrf52840-hal" }
+
+[dependencies.embedded-hal]
+version = "0.2.3"
+features = ["unproven"]
--- a/nrf-hal-common/src/wdt.rs
+++ b/nrf-hal-common/src/wdt.rs
@@ -18,7 +18,7 @@ pub struct Watchdog<T: sealed::WdMode> {
 }
 
 /// An interface to feed the Watchdog
-pub struct WatchdogHandle<T:sealed::HandleId>(T);
+pub struct WatchdogHandle<T: sealed::HandleId>(T);
 
 impl<T> WatchdogHandle<T>
 where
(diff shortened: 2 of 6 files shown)


nrf-hal-common/src/wdt.rs
```diff
@@ -0,0 +1,314 @@
...
+}
+
+impl<T> Watchdog<T>
+where
+    T: sealed::WdMode,
+{
+    /// Is the watchdog active?
+    #[inline(always)]
+    pub fn is_running(&self) -> bool {
+        // TODO: Should we just believe the type state?
```
jonas-schievink, line 254: Since the watchdog can apparently be running after reset, it seems like the typestate is actually incorrect, so why not remove it?

jamesmunns: The typestate is mainly there to prevent trying to configure the device after activation (which will have no effect). The watchdog will only be running after a start if there was some kind of soft reset (I'm not sure what kind of resets will or won't clear the watchdog state).

I thought about checking if the watchdog is already active in `Watchdog::new()`, though I'm not sure what the right response there is, other than to panic, or just block until it expires, but that could be bad if the watchdog is set to something huge, like 1.5 days.

jamesmunns:
> Do you think the typestate in the `WatchdogHandle` pulls its weight? It looks like that makes the API somewhat less ergonomic than it could be

Do you have any suggestions for what else to do with this? I suppose we could always use the `degraded` mode, at the const of a u8 for each handle.

examples/wdt-demo/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "wdt-demo"
+version = "0.1.0"
+authors = ["James Munns <james.munns@ferrous-systems.com>"]
+edition = "2018"
+
```
thalesfragoso, line 6:
```suggestion
publish = false

```

nrf-hal-common/src/wdt.rs
```diff
@@ -0,0 +1,450 @@
...
+impl Watchdog<Inactive> {
+    /// Try to create a new watchdog instance from the peripheral
+    ///
+    /// This function will return an error if the watchdog has already
+    /// been activated, which may happen on a (non-watchdog) soft reset.
+    /// In this case, it may be possible to still obtain the handles with
+    /// the `Watchdog::try_recover()` method.
+    ///
+    /// If the watchdog has already started, configuration is no longer possible.
+    pub fn try_new(wdt: WDT) -> Result<Watchdog<Inactive>, WDT> {
```
jonas-schievink, line 75:
```suggestion
    #[inline]
    pub fn try_new(wdt: WDT) -> Result<Watchdog<Inactive>, WDT> {
```

nrf-hal-common/src/wdt.rs
```diff
@@ -0,0 +1,450 @@
...
+            Err(watchdog.wdt)
+        } else {
+            Ok(watchdog)
+        }
+    }
+
+    /// Release the peripheral
+    ///
+    /// Note: The peripheral cannot be released after activation
+    pub fn release(self) -> WDT {
```
jonas-schievink, line 91:
```suggestion
    #[inline]
    pub fn release(self) -> WDT {
```

nrf-hal-common/src/wdt.rs
```diff
@@ -0,0 +1,450 @@
...
+    pub fn set_lfosc_ticks(&mut self, ticks: u32) {
+        self.wdt
+            .crv
+            .write(|w| unsafe { w.bits(ticks.max(0x0000_000F)) });
+    }
+
+    /// Should the watchdog continue to count during sleep modes?
+    ///
+    /// This value defaults to ENABLED on reset.
+    pub fn run_during_sleep(&self, setting: bool) {
```
jonas-schievink, line 148:
```suggestion
    #[inline]
    pub fn run_during_sleep(&self, setting: bool) {
```

nrf-hal-common/src/wdt.rs
```diff
@@ -0,0 +1,450 @@
...
+        self.wdt.config.modify(|_r, w| w.sleep().bit(setting));
+    }
+
+    /// Should the watchdog continue to count when the CPU is halted for debug?
+    ///
+    /// This value defaults to DISABLED on reset.
+    pub fn run_during_debug_halt(&self, setting: bool) {
```
jonas-schievink, line 155:
```suggestion
    #[inline]
    pub fn run_during_debug_halt(&self, setting: bool) {
```

nrf-hal-common/src/wdt.rs
```diff
@@ -0,0 +1,450 @@
...
+                WatchdogHandle(Hdl1),
+                WatchdogHandle(Hdl2),
+                WatchdogHandle(Hdl3),
+                WatchdogHandle(Hdl4),
+                WatchdogHandle(Hdl5),
+                WatchdogHandle(Hdl6),
+                WatchdogHandle(Hdl7),
+            )
+        }
+    }
```
jonas-schievink, line 449: Do you think you could use a macro to generate these impls? Might make the code a bit shorter.

jamesmunns: I'm not against it, but I've already typed it out, and I have very little lust auf writing a macro at the moment.

Issue #141, opened 2020-04-21
Move board support crates out of this repo


jonas-schievink:
This repository should focus on HALs for the microcontrollers, not on providing support for every conceivable devboard that contributors might not have. Board specific crates should be moved to their own repository.

jonathanpallant:
That seems reasonable.

jonathanpallant:
I think we could do this without losing history using git-subtree?

https://dharmitshah.com/2017/10/git-subtree-new-repo/

PR #176, opened 2020-07-12
Delete board support crates

jonas-schievink:
I've moved the remaining ones into their own repos in this org.

Fixes https://github.com/nrf-rs/nrf-hal/issues/141


Yatekii:
LGTM! Thanks a ton for doing the work I was too lazy to do!

PR #178, opened 2020-07-14
Move to 'fixed' crate

diondokter:
This PR moves the fixed point library to one that is stable, maintained and has better functionality.

This is a breaking change, but IMO a minor one.

The only blocker I can think of is that since the 1.0 release of the fixed crate, it requires rustc 1.44 or higher. I don't see a minimum version mentioned here in the HAL though.

### DIFF
--- a/nrf-hal-common/Cargo.toml
+++ b/nrf-hal-common/Cargo.toml
@@ -21,7 +21,7 @@ edition = "2018"
 [dependencies]
 cortex-m = "0.6.2"
 nb = "0.1.2"
-fpa = "0.1.0"
+fixed = "1.0.0"
 rand_core = "0.5.1"
 cfg-if = "0.1.10"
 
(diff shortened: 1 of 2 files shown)


jonas-schievink:
Thanks! Any MSRV bump should be fine, I don't personally have the capacity to make any guarantees here beyond "builds on stable".

diondokter:
Just for everyone's sanity:

The nRF temp sensor takes 0.25 degree steps: [docs](https://infocenter.nordicsemi.com/index.jsp?topic=%2Fcom.nordic.infocenter.nrf52832.ps.v1.1%2Ftemp.html)
These steps are in the two lowest bits.

The fixed crate also has the fractional bits in the lowest bits: [docs (look at the example)](https://docs.rs/fixed/1.0.0/fixed/struct.FixedI32.html#method.from_bits)

PR #179, opened 2020-07-14
Updated dependencies

diondokter:
Updated the nb crate to the 1.0.0 release.
Updated the patch level for other crates as well (this has no effect for the crate though).

### DIFF
--- a/nrf-hal-common/Cargo.toml
+++ b/nrf-hal-common/Cargo.toml
@@ -20,7 +20,7 @@ edition = "2018"
 
 [dependencies]
 cortex-m = "0.6.2"
-nb = "0.1.2"
+nb = "1.0.0"
 fpa = "0.1.0"
 rand_core = "0.5.1"
 cfg-if = "0.1.10"
@@ -55,11 +55,11 @@ version = "0.9.0"
 
 [dependencies.nrf9160-pac]
 optional = true
-version = "0.2.0"
+version = "0.2.1"
 
 [dependencies.embedded-hal]
 features = ["unproven"]
-version = "0.2.3"
+version = "0.2.4"
 
 [features]
 doc = []


hannobraun:
Thank you, looks good to me!

jonas-schievink:
This duplicates the nb crate since embedded-hal depends on an older version of it

hannobraun:
Isn't `nb` 1.0 brought in through the semver trick anyway?

jonas-schievink:
Ah, right. I forgot we did that.

nebelgrau77:
I just stumbled upon a problem that seems to be related to this. I'm using the master branch, and here's what happens:

Bottomline: nrf-hal-common won't compile. The reason is this:

```rust
 = note: expected fn pointer `fn(&mut saadc::Saadc, &mut PIN) -> core::result::Result<_, nb::Error<()>>` (enum `nb::Error`)
               found fn pointer `fn(&mut saadc::Saadc, &mut PIN) -> core::result::Result<_, nb::Error<()>>` (enum `nb::Error`)
 = note: perhaps two different versions of crate `nb` are being used?
```

This affects `saadc` and `timer` modules. 
In two cases the "expected" and "found" look identical as seen above, but then they aren't:

```rust
    = note: expected enum `nb::Error<void::Void>`
               found enum `nb::Error<_>`
    = note: perhaps two different versions of crate `nb` are being used?
```

If I run `cargo tree` it does confirm that two different versions of nb are used: v1.0.0 in the nrf-common-hal, and v0.1.2 in embedded-hal. 

So I checked something that works for me: https://github.com/nebelgrau77/ItsyBitsy_nRF52840_ADC_serialTX - this uses the v0.11.0 of the nrf-common-hal, pulled from crates.io, so the nb version was still v0.1.2. 

In fact downgrading it back to 0.1.2 in a local fork of the master fixed it. 

There's a big fat chance I'm missing something, though! :)

hannobraun:
Please try running `cargo update` instead. That should pick up nb 0.1.3, which provides full compatibility with 1.0 using the semver trick.

PR #188, opened 2020-07-29
Add QDEC module

kalkyl:
Added a HAL module for the quadrature decoder peripheral.
Tested on NRF52840-DK with a Bourns PEC11R-4215F-S0024 rotary encoder

Demo of usage with rotary encoder triggering interrupt (nRF52840):
https://github.com/kalkyl/nrf-hal/blob/qdec/examples/qdec-demo/src/main.rs

### DIFF
--- a/nrf-hal-common/src/qdec.rs
+++ b/nrf-hal-common/src/qdec.rs
@@ -47,7 +47,7 @@ impl Qdec {
         if let Some(p) = &pin_led {
             qdec.psel.led.write(|w| {
                 #[cfg(any(feature = "52833", feature = "52840"))]
-                w.port().bit(match pin_a.port() {
+                w.port().bit(match p.port() {
                     Port::Port0 => false,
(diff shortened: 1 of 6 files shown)


nrf-hal-common/src/qdec.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    /// Returns the accumulated change since last read
+    #[inline(always)]
+    pub fn read(&self) -> i16 {
+        self.qdec.tasks_readclracc.write(|w| unsafe { w.bits(1) });
+        self.qdec.accread.read().bits() as i16
+    }
+
+    /// Consumes `self` and returns back the raw `QDEC` peripheral.
+    pub fn free(
```
jonas-schievink, line 156:
```suggestion
    #[inline]
    pub fn free(
```

kalkyl:
Yes great, thanks for your comments @jonas-schievink !

PR #189, opened 2020-08-01
Add COMP module

kalkyl:
Added a HAL module for the `COMP` analog comparator peripheral.
The comparator compares an input voltage (Vin) against a second input voltage (Vref).
Vin can be derived from an analog input pin (AIN0-AIN7).
Vref can be derived from multiple sources depending on the operation mode of the comparator.
Tested on NRF52840-DK

Demo (nRF52840):
https://github.com/kalkyl/nrf-hal/blob/comp/examples/comp-demo/src/main.rs

### DIFF
--- a/examples/comp-demo/Embed.toml
+++ b/examples/comp-demo/Embed.toml
@@ -1,0 +1,46 @@
+[probe]
+# The index of the probe in the connected probe list.
+# probe_index = 0
+# The protocol to be used for communicating with the target.
+protocol = "Swd"
+# The speed in kHz of the data link to the target.
+# speed = 1337
+
+[flashing]
+# Whether or not the target should be flashed.
+enabled = true
+# Whether or not the target should be halted after flashing.
+halt_afterwards = false
+# Whether or not bytes erased but not rewritten with data from the ELF
+# should be restored with their contents before erasing.
+restore_unwritten_bytes = false
+# The path where an SVG of the assembled flash layout should be written to.
+# flash_layout_output_path = "out.svg"
+
+[general]
+# The chip name of the chip to be debugged.
+chip = "nRF52840"
+# A list of chip descriptions to be loaded during runtime.
+chip_descriptions = []
+# The default log level to be used.
+log_level = "Warn"
+
+[rtt]
+# Whether or not an RTTUI should be opened after flashing.
+# This is exclusive and cannot be used with GDB at the moment.
+enabled = true
+# A list of channel associations to be displayed. If left empty, all channels are displayed.
+channels = [
+    # { up = 0, down = 0, name = "name" }
+]
+# The duration in ms for which the logger should retry to attach to RTT.
+timeout = 3000
+# Whether timestamps in the RTTUI are enabled
+show_timestamps = true
+
+[gdb]
+# Whether or not a GDB server should be opened after flashing.
+# This is exclusive and cannot be used with RTT at the moment.
+enabled = false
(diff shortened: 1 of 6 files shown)


nrf-hal-common/src/comp.rs
```diff
@@ -0,0 +1,277 @@
...
+
+    /// Returns reference to `Down` transition event endpoint for PPI.
+    pub fn event_down(&self) -> &Reg<u32, _EVENTS_DOWN> {
+        &self.comp.events_down
+    }
+
+    /// Returns reference to `Cross` transition event endpoint for PPI.
+    pub fn event_cross(&self) -> &Reg<u32, _EVENTS_CROSS> {
+        &self.comp.events_cross
+    }
```
jonas-schievink, line 183: Seems like these event functions could also use an `#[inline]`

kalkyl: Totally!

examples/comp-demo/Embed.toml
```diff
@@ -0,0 +1,46 @@
+[probe]
```
jamesmunns, line 1: You might want to update to cargo-embed v0.8 and change these to [default.probe] etc.

nrf-hal-common/src/comp.rs
```diff
@@ -0,0 +1,281 @@
...
+        self.comp.extrefsel.write(|w| match ref_pin.ain() {
+            0 => w.extrefsel().analog_reference0(),
+            1 => w.extrefsel().analog_reference1(),
+            2 => w.extrefsel().analog_reference2(),
+            3 => w.extrefsel().analog_reference3(),
+            4 => w.extrefsel().analog_reference4(),
+            5 => w.extrefsel().analog_reference5(),
+            6 => w.extrefsel().analog_reference6(),
+            7 => w.extrefsel().analog_reference7(),
+            _ => unreachable!(),
```
jamesmunns, line 75: Maybe use an enum instead of an index for this? You can probably return an `EXTREFSEL_A` from your `ain()` function and just use do `w.extrefsel().variant(ref_pin.ain())`.

 https://docs.rs/nrf52840-pac/0.9.0/nrf52840_pac/comp/extrefsel/enum.EXTREFSEL_A.html

nrf-hal-common/src/comp.rs
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    /// Consumes `self` and returns back the raw `COMP` peripheral.
+    #[inline(always)]
+    pub fn free(self) -> COMP {
+        self.comp
+    }
+}
+
+pub enum OperationMode {
```
jamesmunns, line 224: Probably want to hit all of these enums with derives for Debug, Eq, PartialEq, Clone, and Copy.

nrf-hal-common/src/comp.rs
```diff
@@ -0,0 +1,281 @@
...
+    Int1V8,
+    Int2V4,
+    Vdd,
+    ARef,
+}
+
+/// Trait to represent analog input pins.
+pub trait AnalogPin {
+    /// Returns `AIN` id.
+    fn ain(&self) -> u8;
```
jamesmunns, line 257: See note above about returning EXTREFSEL_A

nrf-hal-common/src/comp.rs
kalkyl: Good idea! Did the same with PSEL_A for input

jamesmunns:
Merging manually due to false CI failures

PR #190, opened 2020-08-02
Remove `Spi::read`

jonas-schievink:
It was deprecated a while ago. Breaking change.

### DIFF
--- a/examples/spi-demo/src/main.rs
+++ b/examples/spi-demo/src/main.rs
@@ -63,11 +63,11 @@ fn main() -> ! {
 
     // This will write 8 bytes, then shift out ORC
 
-    // Note :     spi.read( &mut cs.degrade(), reference_data, &mut readbuf )
-    //            will fail because reference data is in flash, the copy to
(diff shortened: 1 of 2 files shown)


examples/spi-demo/src/main.rs
```diff
@@ -63,11 +63,11 @@ fn main() -> ! {
 
     // This will write 8 bytes, then shift out ORC
 
-    // Note :     spi.read( &mut cs.degrade(), reference_data, &mut readbuf )
-    //            will fail because reference data is in flash, the copy to
-    //            an array will move it to RAM.
+    // Note: `spi.transfer_split_uneven(&mut cs.degrade(), reference_data, &mut readbuf)`
+    //       will fail because reference data is in flash, the copy to an array
+    //       will move it to RAM.
```
simonsso, line 68: Is this limitation and comment still valid?

jonas-schievink: Should be, yes

Issue #191, opened 2020-08-02
Re-Enable RTIC tests


jamesmunns:

jonas-schievink:
CI is working again

PR #192, opened 2020-08-02
Disable RTIC CI for now

jamesmunns:
Due to https://github.com/rust-lang/rust/issues/75074, RTIC examples seem to hang the nightly build forever. Disable these tests for now.

Opened https://github.com/nrf-rs/nrf-hal/issues/191 to remind us to re-enable


jonas-schievink:
We should mark nightly as an allowed failure, if that's possible on GHA

jonas-schievink:
CI is working again

PR #195, opened 2020-08-02
Add LPCOMP module

kalkyl:
Added a HAL module for the LPCOMP low power comparator peripheral.

In System ON, the LPCOMP can generate separate events on rising and falling edges of a signal, or sample the current state of the pin as being above or below the selected reference. The block can be configured to use any of the analog inputs on the device.

Additionally, the low power comparator can be used as an analog wakeup source from System OFF or System ON. The comparator threshold can be programmed to a range of fractions of the supply voltage or to use an external analog reference input pin.

Tested on nRF52840-DK

Demo showing "analog power up" usage and interrupt triggering/reading comparator state:
https://github.com/kalkyl/nrf-hal/blob/lpcomp/examples/lpcomp-demo/src/main.rs

### DIFF
--- a/nrf-hal-common/src/lpcomp.rs
+++ b/nrf-hal-common/src/lpcomp.rs
@@ -147,9 +147,9 @@ impl LpComp {
     #[inline(always)]
     pub fn reset_event(&self, event: Transition) {
         match event {
-            Transition::Cross => self.lpcomp.events_cross.write(|w| w),
-            Transition::Down => self.lpcomp.events_down.write(|w| w),
-            Transition::Up => self.lpcomp.events_up.write(|w| w),
+            Transition::Cross => self.lpcomp.events_cross.reset(),
+            Transition::Down => self.lpcomp.events_down.reset(),
+            Transition::Up => self.lpcomp.events_up.reset(),
         }
     }
 
--- a/nrf-hal-common/src/lpcomp.rs
+++ b/nrf-hal-common/src/lpcomp.rs
@@ -1,8 +1,11 @@
 //! HAL interface for the LPCOMP peripheral.
 //!
-//! The comparator (LPCOMP) compares an input voltage (Vin) against a second input voltage (Vref).
-//! Vin can be derived from an analog input pin (AIN0-AIN7).
-//! Vref can be derived from multiple sources depending on the operation mode of the comparator.
+//! In System ON, the LPCOMP can generate separate events on rising and falling edges of a signal,
+//! or sample the current state of the pin as being above or below the selected reference.
+//! The block can be configured to use any of the analog inputs on the device.
+//! Additionally, the low power comparator can be used as an analog wakeup source from System OFF.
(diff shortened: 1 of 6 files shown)


nrf-hal-common/src/lpcomp.rs
```diff
@@ -0,0 +1,295 @@
...
+            Transition::Up => self.lpcomp.events_up.write(|w| w),
+        }
+    }
+
+    /// Marks all events as handled.
+    #[inline(always)]
+    pub fn reset_events(&self) {
+        self.lpcomp.events_cross.write(|w| w);
+        self.lpcomp.events_down.write(|w| w);
+        self.lpcomp.events_up.write(|w| w);
```
jonas-schievink, line 161:
```suggestion
        self.lpcomp.events_cross.reset();
        self.lpcomp.events_down.reset();
        self.lpcomp.events_up.reset();
```

nrf-hal-common/src/lpcomp.rs
```diff
@@ -0,0 +1,295 @@
...
+        &self.lpcomp.events_cross
+    }
+
+    /// Marks event as handled.
+    #[inline(always)]
+    pub fn reset_event(&self, event: Transition) {
+        match event {
+            Transition::Cross => self.lpcomp.events_cross.write(|w| w),
+            Transition::Down => self.lpcomp.events_down.write(|w| w),
+            Transition::Up => self.lpcomp.events_up.write(|w| w),
```
jonas-schievink, line 152:
```suggestion
            Transition::Cross => self.lpcomp.events_cross.reset(),
            Transition::Down => self.lpcomp.events_down.reset(),
            Transition::Up => self.lpcomp.events_up.reset(),
```

nrf-hal-common/src/lpcomp.rs
```diff
@@ -0,0 +1,295 @@
...
+    lpcomp: LPCOMP,
+}
+
+impl LpComp {
+    /// Takes ownership of the `LPCOMP` peripheral, returning a safe wrapper.
+    pub fn new<P: LpCompInputPin>(lpcomp: LPCOMP, input_pin: &P) -> Self {
+        lpcomp.psel.write(|w| w.psel().variant(input_pin.ain()));
+        lpcomp
+            .refsel
+            .write(|w| w.refsel().bits(VRef::_4_8Vdd.into()));
```
jonas-schievink, line 30: Why select a reference here? Is this just resetting the value of refsel? If so, `lpcomp.refsel.reset();` should work too.

kalkyl: I just chose an arbitrary reasonable in-the-middle default value, to have an "expected" default behaviour... Could be omitted, i just thought it was a good start value :)

jonas-schievink: Okay, that seems fine. The function docs should mention it though.

PR #196, opened 2020-08-03
Add TWIS module

kalkyl:
Added a HAL module for TWIS (I2C compatible two-wire interface peripheral, formerly referred to as an "I2C slave")

Here's a demo showing read and write operations using interrupts and multiple I2C addresses (ItsyBitsy nRF52840):
https://github.com/kalkyl/nrf-hal/blob/twis/examples/twis-demo/src/main.rs

Also added a TWIM demo that can also be used together with the above TWIS demo, to show the corresponding controller code (nRF52840-DK):
https://github.com/kalkyl/nrf-hal/blob/twis/examples/twim-demo/src/main.rs

### DIFF
--- a/examples/twim-demo/src/main.rs
+++ b/examples/twim-demo/src/main.rs
@@ -1,0 +1,127 @@
+#![no_std]
+#![no_main]
+
+use embedded_hal::digital::v2::InputPin;
+use {
+    core::{
+        panic::PanicInfo,
+        sync::atomic::{compiler_fence, Ordering},
+    },
+    hal::{
+        gpio::{p0::Parts, Input, Pin, PullUp},
+        gpiote::Gpiote,
+        pac::TWIM0,
+        twim::*,
+    },
+    nrf52840_hal as hal,
+    rtic::cyccnt::U32Ext,
+    rtt_target::{rprintln, rtt_init_print},
+};
+
+#[rtic::app(device = crate::hal::pac, peripherals = true,  monotonic = rtic::cyccnt::CYCCNT)]
+const APP: () = {
+    struct Resources {
+        twim: Twim<TWIM0>,
+        gpiote: Gpiote,
+        btn1: Pin<Input<PullUp>>,
+        btn2: Pin<Input<PullUp>>,
+        btn3: Pin<Input<PullUp>>,
+        btn4: Pin<Input<PullUp>>,
+    }
+
+    #[init]
+    fn init(mut ctx: init::Context) -> init::LateResources {
+        let _clocks = hal::clocks::Clocks::new(ctx.device.CLOCK).enable_ext_hfosc();
+        ctx.core.DCB.enable_trace();
+        ctx.core.DWT.enable_cycle_counter();
+        rtt_init_print!();
+
+        let p0 = Parts::new(ctx.device.P0);
+        let scl = p0.p0_30.into_floating_input().degrade();
+        let sda = p0.p0_31.into_floating_input().degrade();
+        let btn1 = p0.p0_11.into_pullup_input().degrade();
+        let btn2 = p0.p0_12.into_pullup_input().degrade();
+        let btn3 = p0.p0_24.into_pullup_input().degrade();
+        let btn4 = p0.p0_25.into_pullup_input().degrade();
+
+        let gpiote = Gpiote::new(ctx.device.GPIOTE);
+        gpiote.port().input_pin(&btn1).low();
+        gpiote.port().input_pin(&btn2).low();
+        gpiote.port().input_pin(&btn3).low();
+        gpiote.port().input_pin(&btn4).low();
+        gpiote.port().enable_interrupt();
+
+        let twim = Twim::new(ctx.device.TWIM0, Pins { scl, sda }, Frequency::K100);
+
+        init::LateResources {
+            twim,
+            gpiote,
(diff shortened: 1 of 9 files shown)


examples/twis-demo/src/main.rs
```diff
@@ -0,0 +1,120 @@
...
+        let sda = p0.p0_16.into_floating_input().degrade();
+
+        let twis = Twis::new(ctx.device.TWIS0, Pins { scl, sda }, ADDR0);
+        twis.address1(ADDR1) // Add a secondary i2c address
+            .enable_interrupt(TwiEvent::Write) // Trigger interrupt on WRITE command
+            .enable_interrupt(TwiEvent::Read) // Trigger interrupt on READ command
+            .enable();
+
+        let buffer0 = singleton!(: [u8; BUF0_SZ] = [0; BUF0_SZ]).unwrap();
+        let buffer1 = singleton!(: [u8; BUF1_SZ] = [0; BUF1_SZ]).unwrap();
```
jonas-schievink, line 51: These don't have to be created late, they can be put into `Resources` with an `#[init]` value

nrf-hal-common/src/twis.rs
```diff
@@ -0,0 +1,322 @@
...
+
+    /// Configures secondary I2C address.
+    #[inline(always)]
+    pub fn address1(&self, address1: u8) -> &Self {
+        self.0.address[1].write(|w| unsafe { w.address().bits(address1) });
+        self.0.config.modify(|_r, w| w.address1().enabled());
+        self
+    }
+
+    /// Sets ORC.
```
jonas-schievink, line 81:
```suggestion
    /// Sets the over-read character (character sent on over-read of the transmit buffer).
```

nrf-hal-common/src/twis.rs
```diff
@@ -0,0 +1,322 @@
...
+                    .drive()
+                    .s0d1()
+                    .sense()
+                    .disabled()
+            });
+        }
+
+        twis.psel.scl.write(|w| {
+            let w = unsafe { w.pin().bits(pins.scl.pin()) };
+            #[cfg(feature = "52840")]
```
jonas-schievink, line 56: Is the 52833 intentionally missing here?

Yatekii: Maybe we should make a combined feature flag (e.g. `extended-port`) which is required by all the chips with an extended port that is used for all those places, so we cannot forget any newly added cores (it would also make adding new cores easier).

jonas-schievink: Yeah, will do that at some point

kalkyl: Ahh no that was not intentional, good catch! 👍

examples/twim-demo/src/main.rs
```diff
@@ -0,0 +1,127 @@
...
+        gpiote::Gpiote,
+        pac::TWIM0,
+        twim::*,
+    },
+    nrf52840_hal as hal,
+    rtic::cyccnt::U32Ext,
+    rtt_target::{rprintln, rtt_init_print},
+};
+
+#[rtic::app(device = crate::hal::pac, peripherals = true,  monotonic = rtic::cyccnt::CYCCNT)]
```
jonas-schievink, line 21:
```suggestion
#[rtic::app(device = crate::hal::pac, peripherals = true, monotonic = rtic::cyccnt::CYCCNT)]
```

examples/twim-demo/src/main.rs
```diff
@@ -0,0 +1,127 @@
...
+        if ctx.resources.btn1.is_low().unwrap() {
+            rprintln!("\nREAD from address 0x1A");
+            let rx_buf = &mut [0; 8][..];
+            let res = twim.read(0x1A, rx_buf);
+            rprintln!("Result: {:?}\n{:?}", res, rx_buf);
+        }
+        if ctx.resources.btn2.is_low().unwrap() {
+            rprintln!("\nWRITE to address 0x1A");
+            let mut tx_buf = [0u8; 8];
+            tx_buf.copy_from_slice(&[1, 2, 3, 4, 5, 6, 7, 8]);
```
jonas-schievink, line 95:
```suggestion
            let mut tx_buf = [1, 2, 3, 4, 5, 6, 7, 8];
```

examples/twim-demo/src/main.rs
```diff
@@ -0,0 +1,127 @@
...
+        if ctx.resources.btn3.is_low().unwrap() {
+            rprintln!("\nREAD from address 0x1B");
+            let rx_buf = &mut [0; 4][..];
+            let res = twim.read(0x1B, rx_buf);
+            rprintln!("Result: {:?}\n{:?}", res, rx_buf);
+        }
+        if ctx.resources.btn4.is_low().unwrap() {
+            rprintln!("\nWRITE to address 0x1B");
+            let mut tx_buf = [0u8; 4];
+            tx_buf.copy_from_slice(&[9, 10, 11, 12]);
```
jonas-schievink, line 108:
```suggestion
            let mut tx_buf = [9, 10, 11, 12];
```

kalkyl: The reason for this was to ensure that the data resides in RAM, but will change to this and also pull in the embedded_hal traits that will take care of it if needed

PR #199, opened 2020-08-07
WIP: Usb device

dzervas:
This is a rebase of the work that @japaric did, with the test that was suggested on #144.

Problem is that I can't get it to work, I'm getting on `dmesg`:
```
[  +0.306298] usb 2-2: new full-speed USB device number 123 using xhci_hcd
[ +10.536670] usb 2-2: device descriptor read/all, error -110
```

I think that the problem is that the driver never responds to the OS and I don't know how to fix that. I'd admire some help or at least some pointers, as tockos & Adafruit's TinyUSB both use different RTOSes

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ members = [
   "examples/gpiote-demo",
   "examples/wdt-demo",
   "examples/comp-demo",
+  "examples/usb",
 ]
 
 [profile.dev]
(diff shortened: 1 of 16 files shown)


dzervas:
Even after using the following patch on `/Cargo.toml` as suggested in the original MR, it doesn't work...

```Rust
[patch.crates-io.usb-device]
git = "https://github.com/jonas-schievink/usb-device.git"
branch = "inhibit-setaddr-resp"
```

dzervas:
I gave up, I really can't figure how this can be done. I've tried replicating the USB stack of tinyusb (adafruit implementation), tock os, mbed, upstream nordic, I just can't figure it out. The USB loops in the suspend state.

PR #200, opened 2020-08-10
Add PWM module

kalkyl:
Added a HAL module for the PWM peripheral

LED demo of some different modes and features (nRF52840-DK):
https://github.com/kalkyl/nrf-hal/blob/pwm/examples/pwm-demo/src/main.rs

### DIFF
--- a/nrf-hal-common/src/pwm.rs
+++ b/nrf-hal-common/src/pwm.rs
@@ -2,6 +2,7 @@
 //!
 //! The pulse with modulation (PWM) module enables the generation of pulse width modulated signals on GPIO.
 
+use core::cell::RefCell;
 use core::sync::atomic::{compiler_fence, Ordering};
 
 #[cfg(any(feature = "52833", feature = "52840"))]
@@ -23,7 +24,7 @@ use crate::{
 #[derive(Debug)]
 pub struct Pwm<T: Instance> {
     pwm: T,
-    duty: [u16; 4],
+    duty: RefCell<[u16; 4]>,
 }
 
 impl<T> Pwm<T>
@@ -33,7 +34,7 @@ where
     /// Takes ownership of the peripheral and applies sane defaults.
     pub fn new(pwm: T) -> Pwm<T> {
         compiler_fence(Ordering::SeqCst);
-        let duty = [0u16; 4];
+        let duty = RefCell::new([0u16; 4]);
         pwm.enable.write(|w| w.enable().enabled());
         pwm.mode.write(|w| w.updown().up());
         pwm.prescaler.write(|w| w.prescaler().div_1());
@@ -272,7 +273,7 @@ where
     // Internal helper function that returns 15 bit duty cycle value.
     #[inline(always)]
     fn duty_on_value(&self, index: usize) -> u16 {
-        let val = self.duty[index];
+        let val = self.duty.borrow()[index];
         let is_inverted = (val >> 15) & 1 == 0;
         match is_inverted {
             false => val,
@@ -283,7 +284,7 @@ where
     // Internal helper function that returns 15 bit inverted duty cycle value.
     #[inline(always)]
     fn duty_off_value(&self, index: usize) -> u16 {
-        let val = self.duty[index];
+        let val = self.duty.borrow()[index];
         let is_inverted = (val >> 15) & 1 == 0;
         match is_inverted {
             false => self.max_duty() - val,
@@ -294,9 +295,9 @@ where
     /// Sets duty cycle (15 bit) for all PWM channels.
     pub fn set_duty_on_common(&self, duty: u16) {
         compiler_fence(Ordering::SeqCst);
-        unsafe {
-            *(self.duty.as_ptr() as *mut u16) = duty.min(self.max_duty()) & 0x7FFF;
-        }
+        self.duty
+            .borrow_mut()
+            .copy_from_slice(&[duty.min(self.max_duty()) & 0x7FFF; 4][..]);
         self.one_shot();
         self.set_load_mode(LoadMode::Common);
         self.pwm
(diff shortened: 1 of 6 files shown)


Mirabellensaft:
Hi! I have looked over your PR, the pwm.rs looks good to me so far. Looking at the demo, it would be really helpful to have comments that explain what the different parts of the demo are about.

kalkyl:
Thanks, good idea!

nrf-hal-common/src/pwm.rs
```diff
@@ -0,0 +1,1000 @@
...
+}
+
+#[derive(Debug)]
+pub enum Error {
+    DMABufferNotInDataMemory,
+}
+
+mod sealed {
+    use core::ops::Deref;
+    pub trait Instance: Deref<Target = crate::pac::pwm0::RegisterBlock> {}
```
Dirbaio, line 988: Making the trait private here makes it impossible to generically name the type  `Pwm<T>` in user code.

```rust
fn buzzer<T: pwm::sealed::Instance>(pwm: &mut pwm::Pwm<T>) {
```

complains that `sealed` is private.

Yatekii: That is the idea of this design. You use a sealed module when its types should stay inaccessible :) I am not sure what the exact intentions are in this case tho :)

Dirbaio: This makes it impossible to write a function that "does something with a PWM instance". I would like to do this:

```rust
fn buzzer<T: pwm::sealed::Instance>(pwm: &mut pwm::Pwm<T>) { .. }
```

but with the current API, I'm forced to copy-paste the function 4 times.

```rust
fn buzzer_pwm0(pwm: &mut pwm::Pwm<PWM0>) { .. }
fn buzzer_pwm1(pwm: &mut pwm::Pwm<PWM1>) { .. }
fn buzzer_pwm2(pwm: &mut pwm::Pwm<PWM2>) { .. }
fn buzzer_pwm3(pwm: &mut pwm::Pwm<PWM3>) { .. }
```

which is not ideal.

If you want to seal the trait, a better alternative would be to keep it public, but have it require a private trait, like [this](https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed). 

This way client code can still name the trait, but not add new impls.

kalkyl: Good idea! I was mostly just following the pattern used in the rest of this HAL, but i agree this is more useful

nrf-hal-common/src/pwm.rs
```diff
@@ -0,0 +1,1017 @@
...
+    /// Sets the PWM clock prescaler.
+    #[inline(always)]
+    pub fn set_prescaler(&self, div: Prescaler) -> &Self {
+        self.pwm.prescaler.write(|w| w.prescaler().bits(div.into()));
+        self
+    }
+
+    /// Sets the PWM clock prescaler.
+    #[inline(always)]
+    pub fn get_prescaler(&self) -> Prescaler {
```
jonas-schievink, line 73: The convention for getters is to omit the `get_` prefix, so:

```suggestion
    pub fn prescaler(&self) -> Prescaler {
```

(same for the others)

nrf-hal-common/src/pwm.rs
```diff
@@ -0,0 +1,1017 @@
...
+            false => self.get_max_duty() - val,
+            true => val & 0x7FFF,
+        }
+    }
+
+    /// Sets duty cycle (15 bit) for all PWM channels.
+    pub fn set_duty_on_common(&self, duty: u16) {
+        compiler_fence(Ordering::SeqCst);
+        unsafe {
+            *(self.duty.as_ptr() as *mut u16) = duty.min(self.get_max_duty()) & 0x7FFF;
```
jonas-schievink, line 298: This is UB since `as_ptr` only gives immutable access.

```suggestion
            *(self.duty.as_mut() as *mut u16) = duty.min(self.get_max_duty()) & 0x7FFF;
```

You also need to take a `&mut self` here.

(this pattern also comes up a lot in the functions below)

kalkyl: The problem is i can't use a &mut self because i want to be able to split out those channel and group "proxies"...
Would you suggest i use a RefCell instead?

jonas-schievink: Either that or somehow use a buffer per channel/group if that's possible.

kalkyl: Unfortunately that's not possible because of the way the DMA is implemented...
So I'll go with a RefCell then :)

nrf-hal-common/src/pwm.rs
```diff
@@ -0,0 +1,996 @@
...
+    pub fn set_seq_end_delay(&self, seq: Seq, periods: u32) -> &Self {
+        match seq {
+            Seq::Seq0 => self.pwm.seq0.enddelay.write(|w| unsafe { w.bits(periods) }),
+            Seq::Seq1 => self.pwm.seq1.enddelay.write(|w| unsafe { w.bits(periods) }),
+        }
+        self
+    }
+
+    /// Loads a sequence buffer.
+    /// NOTE: `buf` must live until the sequence is done playing.
```
jonas-schievink, line 469: Hmm, should this be `unsafe`? I see the same issue applies to `self.duty`. It *should* be fine from what I understand though, since all DMA does here is read.

The documentation should probably state what happens when this is not done correctly. And the same goes for functions that modify `self.duty` while PWM is running, if that makes sense.

kalkyl: The worst thing that could happen if the sequence buffer doesn't live long enough is that it will play a corrupted sequence, would that be considered unsafe?

Regarding the functions that modify `self.duty`, they all quit any playing sequence and the other channels will return to their previously set duty cycle, which i think would be considered the expected behaviour? I can't see anything unsafe there, although i agree the above perhaps should be mentioned in the docs.

jonas-schievink: Yeah, that should be fine then.

kalkyl: Cool! I'll add the docs :)

huntc:
Thanks for this great contribution @kalkyl.

Being a first-time user of PWM, I did find the example introducing many new concepts. I'm sure the example will be really useful to me over time, but I was wondering if you knew of "blinky-style" examples using `Pwm` and an RGB LED? I'm presently writing a BSP for the Thingy:91 and struggling a little: https://users.rust-lang.org/t/guidance-for-using-pwm-on-the-thingy-91/59516. I'd be happy to contribute an additional example here though if I can get mine to work.

PR #201, opened 2020-08-17
Add I2S module

kalkyl:
Added a HAL module for the I2S peripheral

Audio (Morse code) demo for I2S in "controller mode" (formerly referred to as "I2S master"),
transmitting audio signals to a DAC:
https://github.com/kalkyl/nrf-hal/blob/i2s/examples/i2s-controller-demo/src/main.rs
Tested on nRF52840-DK with a UDA1334a DAC

...or connect it to the...

"Peripheral mode" (formerly referred to as "I2S slave") demo,
using an RGB LED (APA102) to display the RMS level of the received I2S audio signal:
https://github.com/kalkyl/nrf-hal/blob/i2s/examples/i2s-peripheral-demo/src/main.rs
Tested on ItsyBitsy nRF52840 with its onboard RGB LED

### DIFF
--- a/nrf-hal-common/src/i2s.rs
+++ b/nrf-hal-common/src/i2s.rs
@@ -248,7 +248,7 @@ impl I2S {
             return Err(Error::DMABufferNotInDataMemory);
         }
 
-        if buf.maxcnt() > 4096 {
+        if buf.maxcnt() > 16_384 {
             return Err(Error::BufferTooLong);
         }
 
@@ -271,7 +271,7 @@ impl I2S {
             return Err(Error::DMABufferNotInDataMemory);
         }
 
-        if buf.maxcnt() > 4096 {
+        if buf.maxcnt() > 16_384 {
             return Err(Error::BufferTooLong);
         }
 
@@ -310,7 +310,7 @@ impl I2S {
     /// Sets the size (in 32bit words) of the receive and transmit buffers.
     #[inline(always)]
     pub fn set_buffersize(&self, n_32bit: u32) -> Result<(), Error> {
-        if n_32bit > 4096 {
+        if n_32bit > 16_384 {
             return Err(Error::BufferTooLong);
         }
         self.i2s.rxtxd.maxcnt.write(|w| unsafe { w.bits(n_32bit) });
--- a/nrf-hal-common/src/i2s.rs
+++ b/nrf-hal-common/src/i2s.rs
@@ -1,0 +1,576 @@
+//! HAL interface for the I2S peripheral.
+//!
+
+#[cfg(not(feature = "9160"))]
+use crate::pac::{i2s, I2S as I2S_PAC};
+#[cfg(feature = "9160")]
+use crate::pac::{i2s_ns as i2s, I2S_NS as I2S_PAC};
+use crate::{
+    gpio::{Floating, Input, Output, Pin, PushPull},
+    pac::generic::Reg,
+    target_constants::{SRAM_LOWER, SRAM_UPPER},
+};
+use core::sync::atomic::{compiler_fence, Ordering};
+use i2s::{_EVENTS_RXPTRUPD, _EVENTS_STOPPED, _EVENTS_TXPTRUPD, _TASKS_START, _TASKS_STOP};
+
+pub struct I2S {
+    i2s: I2S_PAC,
+}
+
+impl I2S {
+    /// Takes ownership of the raw I2S peripheral, returning a safe wrapper i controller mode.
+    pub fn new_controller(
+        i2s: I2S_PAC,
+        mck_pin: Option<&Pin<Output<PushPull>>>,
+        sck_pin: &Pin<Output<PushPull>>,
+        lrck_pin: &Pin<Output<PushPull>>,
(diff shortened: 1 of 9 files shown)


eldruin:
Cool stuff!
Have you seen the I2S traits proposal at https://github.com/rust-embedded/embedded-hal/pull/204?
It would be interesting to see how they fit for these MCUs.

kalkyl:
> Cool stuff!
> Have you seen the I2S traits proposal at [rust-embedded/embedded-hal#204](https://github.com/rust-embedded/embedded-hal/pull/204)?
> It would be interesting to see how they fit for these MCUs.

Nice, i'll have a look!

nrf-hal-common/src/i2s.rs
```diff
@@ -0,0 +1,576 @@
...
+            .rxtxd
+            .maxcnt
+            .write(|w| unsafe { w.bits(buf.maxcnt()) });
+
+        Ok(())
+    }
+
+    /// Sets the receive data buffer (RX).
+    #[inline(always)]
+    pub fn rx_buffer<B: I2SBuffer + ?Sized>(&self, buf: &mut B) -> Result<(), Error> {
```
jonas-schievink, line 269: This can cause UB if `buf`'s lifetime ends before the transfer. We either need to use `buf: &'static mut B`, or make use of the `embedded-dma` traits to allow shorter lifetimes.

Maybe the move to `embedded-dma` should be done for the whole crate (and also for transmit-only buffers, to avoid transmitting garbage). Seems like a bigger project.

nrf-hal-common/src/i2s.rs
```diff
@@ -0,0 +1,576 @@
...
+        if (buf.ptr() as usize) < SRAM_LOWER || (buf.ptr() as usize) > SRAM_UPPER {
+            return Err(Error::DMABufferNotInDataMemory);
+        }
+
+        if buf.maxcnt() > 16_384 {
```
jonas-schievink, line 274: Can you move the size limit to a constant? Is it a hard limit of the I2S peripheral or is it just the `EASY_DMA_SIZE` divided by 4 (the size of the largest I2S data type)? Might be good to explain that in a comment.

nrf-hal-common/src/i2s.rs
```diff
@@ -0,0 +1,576 @@
...
+        if (addr as usize) < SRAM_LOWER || (addr as usize) > SRAM_UPPER {
+            return Err(Error::DMABufferNotInDataMemory);
+        }
+        self.i2s.txd.ptr.write(|w| unsafe { w.ptr().bits(addr) });
+        Ok(())
+    }
+
+    /// Sets the receive buffer RAM start address.
+    #[inline(always)]
+    pub fn set_rx_ptr(&self, addr: u32) -> Result<(), Error> {
```
jonas-schievink, line 302:
```suggestion
    pub unsafe fn set_rx_ptr(&self, addr: u32) -> Result<(), Error> {
```

nrf-hal-common/src/i2s.rs
```diff
@@ -0,0 +1,576 @@
...
+            return Err(Error::DMABufferNotInDataMemory);
+        }
+        self.i2s.rxd.ptr.write(|w| unsafe { w.ptr().bits(addr) });
+        Ok(())
+    }
+
+    /// Sets the size (in 32bit words) of the receive and transmit buffers.
+    #[inline(always)]
+    pub fn set_buffersize(&self, n_32bit: u32) -> Result<(), Error> {
+        if n_32bit > 16_384 {
```
jonas-schievink, line 313: (the size constant could be reused here)

jonas-schievink (commented):
Did an initial pass, looks pretty good!

nrf-hal-common/src/i2s.rs
kalkyl: I'll go with `&'static mut B` for now, ok? And we'll look into the move to `embedded-dma` for the whole crate later?

jonas-schievink: Yeah, definitely

nrf-hal-common/src/i2s.rs
```diff
@@ -0,0 +1,580 @@
...
+        if (addr as usize) < SRAM_LOWER || (addr as usize) > SRAM_UPPER {
+            return Err(Error::DMABufferNotInDataMemory);
+        }
+        self.i2s.rxd.ptr.write(|w| w.ptr().bits(addr));
+        Ok(())
+    }
+
+    /// Sets the size (in 32bit words) of the receive and transmit buffers.
+    #[inline(always)]
+    pub fn set_buffersize(&self, n_32bit: u32) -> Result<(), Error> {
```
jonas-schievink, line 316: This one's also unsafe when used with `rx_buffer` since it writes an arbitrary value to `maxcnt`.

kalkyl: True! :)

jonas-schievink:
Ah, dang, looks like the i2s-peripheral-demo doesn't build anymore

kalkyl:
Ahh, so it doesn’t see it’s ’static...

PR #207, opened 2020-09-02
Use cargo-xtask in CI

jonas-schievink:
Tests can be run with `cargo test -p xtask`, no bash required. `cargo test` does not yet work, I'll see if it's possible to get that working after this PR.

My plan is to use `cargo-xtask` as a general automation tool for releases (allowing automatically bumping versions, `html_root_url`, and the changelog).


jonas-schievink:
Oh, and this removes the non-working i2s-peripheral-demo since we now check that every example has test metadata (cc @kalkyl)

kalkyl:
> Oh, and this removes the non-working i2s-peripheral-demo since we now check that every example has test metadata (cc @kalkyl)

Nice! Yeah i'm just finishing up a new PR tonight where I changed the implementation to work along the lines of the `embedded-dma` traits instead, so there's an updated working version of this demo in that PR :)

.cargo/config
```diff
@@ -1,8 +1,8 @@
...
+xtask = "run -p xtask --"
+
 [target.'cfg(all(target_arch = "arm", target_os = "none"))']
 runner = 'arm-none-eabi-gdb'
 rustflags = [
     "-C", "link-arg=-Tlink.x",
 ]
-
-[build]
-target = "thumbv7em-none-eabihf"
```
Yatekii, line 8: why is this gone? I would assume this is still relevant for development?

jonas-schievink: Sort of. We now explicitly pass the right target in CI, and I've found this default to be a bit confusing since it doesn't work on nRF51, the nRF52810, and the nRF9160.

Demos can add their own `.cargo/config` if they want though, that should make development easier.

jonas-schievink: Ah, I guess it does work on the nRF9160, but it's not usually the target you *want* to use there.

Yatekii: Hmm ok, so I guess you suggest just using the build script during development I guess :) I always hit cargo build out of habit =D but the change & its reasoning is fair enough for me!

jonas-schievink: Yeah, you can still do that if you `cd` to the demo project you're editing and add a `.cargo/config` there.

Yatekii: I don't care about the demo projects ;) I care about building the HAL if I change something (which hasn't happened in a log time).

jonas-schievink: The HALs should all build even for x86_64 (at least that's how docs.rs generates documentation for them)

Yatekii: Does cortex-m really build for x86 targets?

jonas-schievink:
Going to merge this tomorrow unless there are objections :)

.cargo/config
jonas-schievink: Yes, it goes out of its way to ensure that

PR #209, opened 2020-09-02
Use DMA Transfer API in I2S module

kalkyl:
I changed the I2S module to use a DMA `Transfer` API for TX and RX operations.
I also added support for full duplex transfers and the demos were updated according to the new API.

"Controller mode" demo (TX):
https://github.com/kalkyl/nrf-hal/blob/i2s-dma/examples/i2s-controller-demo/src/main.rs

"Peripheral mode" demo (RX):
https://github.com/kalkyl/nrf-hal/blob/i2s-dma/examples/i2s-peripheral-demo/src/main.rs

### DIFF
--- a/nrf-hal-common/src/i2s.rs
+++ b/nrf-hal-common/src/i2s.rs
@@ -10,11 +10,8 @@ use crate::{
     pac::generic::Reg,
     target_constants::{SRAM_LOWER, SRAM_UPPER},
 };
-use core::{
-    ops::{Deref, DerefMut},
-    sync::atomic::{compiler_fence, Ordering},
-};
-pub use stable_deref_trait::StableDeref;
+use core::sync::atomic::{compiler_fence, Ordering};
+use embedded_dma::*;
 
 use i2s::{_EVENTS_RXPTRUPD, _EVENTS_STOPPED, _EVENTS_TXPTRUPD, _TASKS_START, _TASKS_STOP};
 
@@ -23,7 +20,7 @@ pub struct I2S {
 }
 
 // I2S EasyDMA MAXCNT bit length = 14
-const MAX_DMA_MAXCNT: u32 = 16_384;
+const MAX_DMA_MAXCNT: u32 = 1 << 14;
 
 impl I2S {
     /// Takes ownership of the raw I2S peripheral, returning a safe wrapper in controller mode.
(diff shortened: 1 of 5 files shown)


jonas-schievink:
Thanks!

I think `Pin` might not be the right thing to use for DMA APIs (see https://github.com/rust-embedded/embedonomicon/issues/64). The [`embedded-dma`](https://docs.rs/embedded-dma/0.1.0/embedded_dma/) crate provides traits for DMA APIs though, I wonder if we should just use those.

kalkyl:
> Thanks!
> 
> I think `Pin` might not be the right thing to use for DMA APIs (see [rust-embedded/embedonomicon#64](https://github.com/rust-embedded/embedonomicon/issues/64)). The [`embedded-dma`](https://docs.rs/embedded-dma/0.1.0/embedded_dma/) crate provides traits for DMA APIs though, I wonder if we should just use those.

Ahh i see... I can't really figure out how to implement them for this at the moment, i guess i haven't fully understood how they're supposted to be used... Any hints?

jonas-schievink:
I've also never used them so far, unfortunately.

kalkyl:
I’ll do my homework ;)

kalkyl:
Ok i have a working implementation using the `embedded_dma` traits. The only problem is  `embedded_dma` doesn't implement `Word` for signed integer types, so we'd need upstream support for that to make it usable.

eldruin:
FYI, I just published `embedded-dma` 0.1.1

nrf-hal-common/src/i2s.rs
```diff
@@ -341,28 +330,24 @@ impl I2S {
...
-            return Err(Error::DMABufferNotInDataMemory);
-        }
-
-        if buffer.maxcnt() > MAX_DMA_MAXCNT {
+        let (ptr, len) = unsafe { buffer.read_buffer() };
+        let maxcnt = (len / (core::mem::size_of::<u32>() / core::mem::size_of::<W>())) as u32;
+        if maxcnt > MAX_DMA_MAXCNT {
             return Err(Error::BufferTooLong);
         }
+        if (ptr as usize) < SRAM_LOWER || (ptr as usize) > SRAM_UPPER {
```
jonas-schievink, line 342: Wouldn't this have to be

```suggestion
        if (ptr as usize) < SRAM_LOWER || (ptr as usize + maxcnt) > SRAM_UPPER {
```

to check for the *whole* buffer, not just its start?

kalkyl: Ahh, i was thinking it'd always be located either "inside RAM" or not?

jonas-schievink: Hmm, yeah, that's true

jonas-schievink (approved):
Looks good, thanks!

PR #217, opened 2020-09-10
Adding Port on GPIOTE Configuration

### LINKED ISSUE #216: Bug: GPIOTE does not set the port of a pin
https://github.com/nrf-rs/nrf-hal/blob/4f41e4827ddbe3df6bbb347ae08bf34aebe60146/nrf-hal-common/src/gpiote.rs#L260

# Summary

Trying to use GPIOTE on the nrf dongle for catching a button interrupt (p1.06) and changing the state of the led (p0.06) the nrf dongle got stuck. Other pin combinations worked.

# Steps to Reproduce

```
#![no_main]
#![no_std]

use nrf52840_hal as hal;

use hal::gpio::{p0, p1};
use hal::gpio::Level;
use hal::gpiote::Gpiote;
use hal::prelude::*;

use core::panic::PanicInfo;
use cortex_m::asm;

#[cortex_m_rt::entry]
fn main() -> ! {
    let p = hal::pac::Peripherals::take().unwrap();

    let port0 = p0::Parts::new(p.P0);
    let port1 = p1::Parts::new(p.P1);
    
    let mut led = port0.p0_06.into_push_pull_output(Level::Low).degrade();
    let button = port1.p1_06.into_pullup_input().degrade();
    

    let gpiote = Gpiote::new(p.GPIOTE);
    gpiote
        .channel0()
        .input_pin(&button)
        .hi_to_lo()
        .enable_interrupt();

    led.set_low().unwrap();
    loop {
        continue;
    }
}

#[panic_handler]
fn panic(_: &PanicInfo) -> ! {
    loop {
        continue;
    }
}
```

# Additional Debug Information

According to @dirbaio there is a bug in `gpiote.rs` L260 (https://github.com/nrf-rs/nrf-hal/blob/master/nrf-hal-common/src/gpiote.rs#L260) where only the pin but not the port is being set.

trembel:
The Configuration of the GPIOTE needs the Port to be set (on devices
with multiple ports)
This resolves #216 .

Signed-off-by: trembel <silvano.cortesi@hotmail.com>

### DIFF
--- a/nrf-hal-common/src/gpiote.rs
+++ b/nrf-hal-common/src/gpiote.rs
@@ -257,6 +257,15 @@ fn config_channel_event_pin<P: GpioteInputPin>(
             EventPolarity::None => w.mode().event().polarity().none(),
             EventPolarity::Toggle => w.mode().event().polarity().toggle(),
         };
+
+        #[cfg(any(feature = "52833", feature = "52840"))]
+        {
+            match pin.port() {
+                Port::Port0 => w.port().clear_bit(),
+                Port::Port1 => w.port().set_bit(),
+            };
+        }
+
         unsafe { w.psel().bits(pin.pin()) }
     });
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Refuse to build nRF52+ HALs for thumbv6m targets ([#203]).
 - Refuse to build `nrf52810-hal` for hard-float targets, and `nrf51-hal` for thumbv7+ targets
   ([#206]).
+- GPIOTE Port 1 and 0 differentiation for nRF52833 and nRF52840 ([#217])
 
 ### Breaking Changes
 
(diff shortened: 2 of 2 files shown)


trembel:
Ok, I'll adjust that (probably by calling a cfg function), since it does not even compile on the other targets because the `port` register/bit is missing

nrf-hal-common/src/gpiote.rs
```diff
@@ -257,6 +259,15 @@ fn config_channel_event_pin<P: GpioteInputPin>(
             EventPolarity::None => w.mode().event().polarity().none(),
             EventPolarity::Toggle => w.mode().event().polarity().toggle(),
         };
+
+        cfg_if! {
```
jonas-schievink, line 263: You don't need this macro here, the cfg attribute can just be applies to a block

trembel:
Does `cfg_if!` adds some runtime costs?

nrf-hal-common/src/gpiote.rs
trembel: thank you, I'll adjust this. Does this add some overhead?

jonas-schievink: No. cfg_if already expands to it anyways.

trembel: I was not aware that attributes can also be applied to blocks.

jonas-schievink:
Thanks! Could you add a changelog entry too?

jonas-schievink:
Thanks!

PR #218, opened 2020-09-11
saadc.rs: add documentation

Lotterleben:
Extends the SAADC documentation with an example and some info on trait impls and default values.

### DIFF
--- a/nrf-hal-common/src/saadc.rs
+++ b/nrf-hal-common/src/saadc.rs
@@ -1,6 +1,7 @@
 //! HAL interface to the SAADC peripheral.
 //!
-//! example usage:
+//! Example usage:
+//!
 //! ```no_run
 //! # use nrf_hal_common as hal;
 //! # use hal::pac::{saadc, SAADC};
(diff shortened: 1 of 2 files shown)


nrf-hal-common/src/saadc.rs
```diff
@@ -23,6 +47,9 @@ pub use saadc::{
 // multiple channels should work (See "scan mode" in the datasheet).
 // Issue: https://github.com/nrf-rs/nrf-hal/issues/82
 
+/// Interface for the SAADC peripheral.
```
jonas-schievink, line 50:
```suggestion
/// Interface for the SAADC peripheral.
///
```

nrf-hal-common/src/saadc.rs
```diff
@@ -65,18 +92,57 @@ impl Saadc {
     }
 }
 
+/// Used to configure the SAADC peripheral.
```
jonas-schievink, line 95:
```suggestion
/// Used to configure the SAADC peripheral.
///
```

nrf-hal-common/src/saadc.rs
```diff
@@ -1,4 +1,28 @@
 //! HAL interface to the SAADC peripheral.
+//!
+//! example usage:
```
jonas-schievink, line 3:
```suggestion
//! Example usage:
//!
```

jonas-schievink (commented):
Very nice! We should find a way to build the doc tests in CI though.

PR #221, opened 2020-09-12
Fix init behavior for P1 pins

jamesmunns:


jonas-schievink (approved):
LGTM. Can you add a changelog entry?

Yatekii (approved):

PR #223, opened 2020-09-16
Enable CCM demo to work on nrf51

therealprof:
Signed-off-by: Daniel Egger <daniel@eggers-club.de>

### DIFF
--- a/examples/ccm-demo/Cargo.toml
+++ b/examples/ccm-demo/Cargo.toml
@@ -11,6 +11,7 @@ cortex-m-rt = "0.6.12"
 rtt-target = {version = "0.2.0", features = ["cortex-m"] }
 rand_core = "0.5.1"
 
+nrf51-hal = { path = "../../nrf51-hal", features = ["rt"], optional = true }
 nrf52810-hal = { path = "../../nrf52810-hal", features = ["rt"], optional = true }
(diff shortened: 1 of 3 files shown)


thalesfragoso:
> nRF51 devices only support payloads of up to 27 bytes.

Maybe we should change the example to use smaller payloads, I remember that I used those sizes because I wanted to try some bench-marking. Since I already included the example as a "benchmark" (see stopwatch.rs) maybe we should use some `cfg`s also on the buffer size.

therealprof:
Huh? Where does that quoted comment come from? 😅

thalesfragoso:
From the docs:
https://github.com/nrf-rs/nrf-hal/blob/af83e6ed69c5ec03b1e0f6b4adae53258f57e83a/nrf-hal-common/src/ccm.rs#L40

And from the manual:

Encryption (Page 121):
> The AES CCM is limited to read maximum 27 bytes of the unencrypted payload (PL) regardless of what is specified in
the length field of the unencrypted packet.

Decryption (Page 122):
> The AES CCM is limited to read maximum 27 bytes of the encrypted payload and four bytes of the MIC regardless of
what is specified in the length field of the encrypted packet.

therealprof:
I'll modify the example to use less payload on nrf51 when I find time (or someone beats me to it 😉 ).

PR #224, opened 2020-09-17
Exclude .DS_Store from workspace

jonas-schievink:
Apparently the glob includes dotfiles, which seems like a Cargo bug. Explicitly exclude `.DS_Store` to fix building on macOS.


jonas-schievink:
This is already fixed on beta and nightly Cargo, so closing.

PR #225, opened 2020-09-17
Add check for I2S buffer pointer alignment

kalkyl:
Buffer pointer addresses need to be 4 byte aligned to be used with the I2S DMA.
This PR adds a check in the `rx`, `transfer` and `tx` `Transfer` constructors and returns an `Error::BufferMisaligned` if this is not fulfilled.

Added trait bounds for supported word sizes (<=4 byte words)

Controller and peripheral demos were updated to utilize `Aligned` to ensure correct alignment.
Tested and verified to be working correctly on hardware.

### DIFF
--- a/examples/i2s-controller-demo/Cargo.toml
+++ b/examples/i2s-controller-demo/Cargo.toml
@@ -12,7 +12,6 @@ cortex-m = "0.6.2"
 cortex-m-rtic = "0.5.3"
 rtt-target = {version = "0.2.0", features = ["cortex-m"] }
 nrf52840-hal = { features = ["rt"], path = "../../nrf52840-hal" }
-aligned = "0.3.4"
 heapless = "0.5.5"
 small_morse = "0.1.0"
 
--- a/examples/i2s-controller-demo/Cargo.toml
+++ b/examples/i2s-controller-demo/Cargo.toml
@@ -12,6 +12,7 @@ cortex-m = "0.6.2"
 cortex-m-rtic = "0.5.3"
 rtt-target = {version = "0.2.0", features = ["cortex-m"] }
 nrf52840-hal = { features = ["rt"], path = "../../nrf52840-hal" }
+aligned = "0.3.4"
 heapless = "0.5.5"
 small_morse = "0.1.0"
 
--- a/examples/i2s-peripheral-demo/src/main.rs
+++ b/examples/i2s-peripheral-demo/src/main.rs
@@ -4,6 +4,7 @@
 // I2S `peripheral mode` demo
 // Signal average level indicator using an RGB LED (APA102 on ItsyBitsy nRF52840)
 
+use aligned::{Aligned, A4};
 use embedded_hal::blocking::spi::Write;
 use {
     core::{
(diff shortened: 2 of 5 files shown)


nrf-hal-common/src/i2s.rs
```diff
@@ -259,14 +259,20 @@ impl I2S {
...
+    /// Buffer address must be 4 byte aligned and located in RAM.
     /// Returns a value that represents the in-progress DMA transfer.
     #[allow(unused_mut)]
     pub fn rx<W, B>(mut self, mut buffer: B) -> Result<Transfer<B>, Error>
     where
+        W: SupportedWordSize,
         B: WriteBuffer<Word = W>,
     {
         let (ptr, len) = unsafe { buffer.write_buffer() };
+        if ptr as u32 % 4 != 0 {
```
thalesfragoso, line 273:
```suggestion
        if ptr as u32 & 3 != 0 {
```

The compiler will almost certainly do the right thing here on release, probably even on debug, but just to be sure we could change to a operation with less overhead. Same thing for the other parts of the code where you do this checking.

thalesfragoso (commented):
Thanks again for this PR, I left some suggestions.

nrf-hal-common/src/i2s.rs
jonas-schievink: Not sure if I'm a fan of that. That's the job of a compiler, after all.

examples/i2s-controller-demo/Cargo.toml
```diff
@@ -12,6 +12,7 @@ cortex-m = "0.6.2"
 cortex-m-rtic = "0.5.3"
 rtt-target = {version = "0.2.0", features = ["cortex-m"] }
 nrf52840-hal = { features = ["rt"], path = "../../nrf52840-hal" }
+aligned = "0.3.4"
```
jonas-schievink, line 15: Let's not add a dependency on aligned; it pulls in an unreasonable amount of transitive deps for something that is really just `#[repr(align(4))]`.

examples/i2s-peripheral-demo/src/main.rs
```diff
@@ -34,7 +35,8 @@ const APP: () = {
 
     #[init]
     fn init(ctx: init::Context) -> init::LateResources {
-        static mut RX_BUF: [i16; 128] = [0; 128];
+        // The I2S buffer address must be 4 byte aligned.
+        static mut RX_BUF: Aligned<A4, [i16; 128]> = Aligned([0; 128]);
```
jonas-schievink, line 39: You can define this wrapper to avoid the `aligned` crate:

```rust
#[repr(align(4))]
struct Aligned<T: ?Sized>(T);
```

nrf-hal-common/src/i2s.rs
kalkyl: It definately makes the code harder to understand... So can we assume the comiler will do it?

examples/i2s-controller-demo/Cargo.toml
kalkyl: Great idea!

nrf-hal-common/src/i2s.rs
jonas-schievink: It definitely will as soon as any optimizations are enabled, and I don't think we should care too much about performance or binary size with optimizations turned off (Rust should generally be expected to have ~Python-like performance in debug mode, so people will turn on optimizations if they want something better)

kalkyl: Agree!

PR #226, opened 2020-09-21
Add SPIS module

kalkyl:
Added a module for the SPIS peripheral, using the `embedded-dma` traits.
Tested on nRF52840.

Demo of usage (pinout is for the ItsyBitsy nRF52840):
https://github.com/kalkyl/nrf-hal/blob/spis/examples/spis-demo/src/main.rs

### DIFF
--- a/nrf-hal-common/src/spis.rs
+++ b/nrf-hal-common/src/spis.rs
@@ -1,3 +1,7 @@
+//! HAL interface to the SPIS peripheral.
+//!
+//! A module for SPI communication in peripheral mode.
+
 use core::{
     ops::Deref,
     sync::atomic::{compiler_fence, Ordering},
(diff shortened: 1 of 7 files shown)


nrf-hal-common/src/spis.rs
```diff
@@ -0,0 +1,590 @@
...
+pub struct Spis<T: Instance>(T);
+
+impl<T> Spis<T>
+where
+    T: Instance,
+{
+    /// Takes ownership of the raw SPIS peripheral, returning a safe wrapper.
+    pub fn new(
+        spis: T,
+        sck_pin: &Pin<Input<Floating>>,
```
jamesmunns, line 39: No strong feeling, but we might want to make a `Pins` struct to match SPIM, etc.

nrf-hal-common/src/spis.rs
```diff
@@ -0,0 +1,590 @@
...
+    /// Disables the SPIS module.
+    #[inline(always)]
+    pub fn disable(&self) -> &Self {
+        self.0.enable.write(|w| w.enable().disabled());
+        self
+    }
+
+    /// Requests acquiring the SPIS semaphore and waits until acquired.
+    #[inline(always)]
+    pub fn acquire(&self) -> &Self {
```
jamesmunns, line 158: We might want a `try_acquire` method that attempts to acquire, but does not block.

nrf-hal-common/src/spis.rs
```diff
@@ -0,0 +1,590 @@
...
+    ) -> Result<TransferSplit<T, TxB, RxB>, Error>
+    where
+        TxB: ReadBuffer<Word = TxW>,
+        RxB: WriteBuffer<Word = RxW>,
+    {
+        let (rx_ptr, rx_len) = unsafe { rx_buffer.write_buffer() };
+        let (tx_ptr, tx_len) = unsafe { tx_buffer.read_buffer() };
+        let rx_maxcnt = rx_len * core::mem::size_of::<RxW>();
+        let tx_maxcnt = tx_len * core::mem::size_of::<TxW>();
+        if rx_maxcnt > EASY_DMA_SIZE || tx_maxcnt > EASY_DMA_SIZE {
```
jamesmunns, line 357:
```suggestion
        if rx_maxcnt.max(tx_maxcnt) > EASY_DMA_SIZE {
```

nrf-hal-common/src/spis.rs
```diff
@@ -0,0 +1,590 @@
+use core::{
```
jamesmunns, line 1: Consider adding docs, and perhaps some example code?

nrf-hal-common/src/spis.rs
```diff
@@ -0,0 +1,590 @@
...
+use crate::pac::{SPIS1, SPIS2};
+
+use crate::{
+    gpio::{Floating, Input, Pin},
+    pac::{generic::Reg, Interrupt},
+    target_constants::{EASY_DMA_SIZE, SRAM_LOWER, SRAM_UPPER},
+};
+use embedded_dma::*;
+
+pub struct Spis<T: Instance>(T);
```
jamesmunns, line 30: Also docs here

jamesmunns (commented):
Overall looks good, I'll plan on using this pretty much immediately!

I did have a couple nits, but nothing huge :)

nrf-hal-common/src/spis.rs
jamesmunns: Oh wait. Stronger objection. I think `Spis` should own these pins. Even though it never uses them, it would be incorrect to allow the user to retain ownership of them and potentially modify the pin states.

jamesmunns (changes requested):

jamesmunns:
Also, could you please add this to the changelog?

kalkyl:
> Overall looks good, I'll plan on using this pretty much immediately!
> 
> I did have a couple nits, but nothing huge :)

Thanks for your comments, agreed and fixed :)

jamesmunns (approved):

PR #229, opened 2020-09-22
Update Embed.toml of all examples to new defaults

andresovela:


jonas-schievink (approved):
Thanks!

In the long term, it seems to me that we might want to get rid of these files entirely, since this is a lot of churn, and they contain a lot of info that isn't very relevant.

Yatekii:
Yeah, people mostly copy the entire file. It is sufficient to just set the bits one needs different from the default (for example the chip). All the other config flags will default to a sane setting automatically.
so we can get rid of 95% of the contents :)
Nobody is bothered by the GDB scripts etc either, so I don't see why this would be an issue :)

PR #230, opened 2020-09-22
Non-blocking TWIS, use embedded-dma

kalkyl:
- Added non-blocking DMA transactions to the TWIS module and renamed the old blocking ones to `rx_blocking` and `tx_blocking`. Using the `embedded-dma` traits.
- Added support for 9160.
- Added `Sealed` bound and `Interrupt` const to `Instance`.
- Added some missing getters and events.
- Renamed some setter functions for consistency.
- Added demo of non-blocking usage: https://github.com/kalkyl/nrf-hal/blob/twis-dma/examples/twis-dma-demo/src/main.rs


nrf-hal-common/src/twis.rs
```diff
@@ -309,14 +486,80 @@ pub enum Error {
...
+    Error,
+    RxStarted,
+    TxStarted,
     Write,
+    Read,
+}
+
+/// A DMA transfer
+pub struct Transfer<T: Instance, B> {
+    inner: Option<Inner<T, B>>,
```
jonas-schievink, line 499: Why is this in an `Option` if it's only `None` when the `Transfer` is being destroyed? Because the `Drop` impl prevents moving out of fields? If you move the `Drop` impl to `Inner` you can move out of `Transfer`'s fields again.

jonas-schievink (commented):
Nice!

nrf-hal-common/src/twis.rs
```diff
@@ -278,6 +395,66 @@ where
         Ok(())
     }
 
+    /// Receives data into the given `buffer`. Buffer must be located in RAM.
+    /// Returns a value that represents the in-progress DMA transfer.
+    #[allow(unused_mut)]
+    pub fn rx<W, B>(mut self, mut buffer: B) -> Result<Transfer<T, B>, Error>
```
jonas-schievink, line 401: This seems to rely on `Drop` being called, right? If a transfer is started and then leaked, the caller could access the original `buffer` again if I'm not mistaken (if it passes a borrowed buffer type like `&mut [T]`).

kalkyl: Goos point, I'll add a + 'static lifetime bound on the buffer to deal with this!

nrf-hal-common/src/twis.rs
kalkyl: The reason was to be able to move out fields in `wait()` and i can't figure out a better way to do that... Or did i misunderstand your suggestion? :)

jonas-schievink: `wait` takes `self` by value, so it can move out of its fields already, without the `Option` (if you apply the `Drop` suggestion)

kalkyl: But it's the fields from `inner` i want to move out, which would be prevented by impl `Drop` on `Inner` no? Sorry if i'm thinking slow here, fever brain...

jonas-schievink: Oh, right, sorry. Then maybe put a FIXME comment there that explains this?

thalesfragoso: Well, you could have `inner` as a struct with no fields, and conjure the registers on `Drop` to be able to stop the DMA. Or do something like I did here: 
https://github.com/stm32-rs/stm32f4xx-hal/blob/36ad03a633145592c8f6bb8c651e25acfb323b08/src/dma/mod.rs#L1095-L1108

I don't have strong feelings about this though.

kalkyl: Good ideas! Although to me the `Option` alternative feels like less mental overhead and it's still not too semantically off no? No strong feelings here either...

nrf-hal-common/src/twis.rs
```diff
@@ -278,6 +395,66 @@ where
         Ok(())
     }
 
+    /// Receives data into the given `buffer`. Buffer must be located in RAM.
+    /// Returns a value that represents the in-progress DMA transfer.
+    #[allow(unused_mut)]
```
jonas-schievink, line 400: Why is this needed?

nrf-hal-common/src/twis.rs
```diff
@@ -140,11 +181,45 @@ where
...
+        self.0.events_stopped.read().bits() != 0
+    }
+
+    /// Returns number of bytes received in last granted transaction.
+    #[inline(always)]
+    pub fn amount(&self) -> u32 {
+        self.0.rxd.amount.read().bits()
+    }
+
+    /// Checks if RX buffer overflow was detected, and prevented.
```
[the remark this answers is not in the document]
kalkyl: Oups, this was just copy/paste from the reference manual without further thought. It makes no sense to me either, I'll remove it from the comment :)

nrf-hal-common/src/twis.rs
kalkyl: It's not, thanks :)

jonas-schievink:
@kalkyl looks like this has some merge conflicts now. If you address those I can merge.

PR #231, opened 2020-09-22
Use embedded-dma trait in PWM module

kalkyl:
- Updated the PWM module to use `embedded-dma` trait for sequences.
- Added support for 52810 & 52811
- Added `Interrupt` const to `Instance`
- Changed internal duty buffers to be static `Cell`
- New proposed API for dealing with sequences, a `PwmSeq` that wraps the `Pwm<T>` together with the loaded sequence buffer(s). The wrapped contents can be accessed by calling `split()` on it.
See updated demo: https://github.com/kalkyl/nrf-hal/blob/pwm-dma/examples/pwm-demo/src/main.rs


nrf-hal-common/src/pwm.rs
```diff
@@ -473,37 +474,29 @@ where
 
     /// Loads a sequence buffer.
     /// NOTE: `buf` must live until the sequence is done playing, or it might play a corrupted sequence.
-    pub fn load_seq(&self, seq: Seq, buf: &[u16]) -> Result<(), Error> {
-        if (buf.as_ptr() as usize) < SRAM_LOWER || (buf.as_ptr() as usize) > SRAM_UPPER {
+    pub fn load_seq<B>(&self, seq: Seq, buf: B) -> Result<(), Error>
+    where
+        B: ReadBuffer<Word = u16>,
+    {
+        let (ptr, len) = unsafe { buf.read_buffer() };
```
jonas-schievink, line 481: The safety contract here seems to be the caller's responsibility, so this `unsafe` doesn't look correct here (the caller could call `&mut self` methods on the buffer if it passes `&[u16]`).

Well, I suppose it's not any worse than before though :)

kalkyl: Hmm, sorry i don't quite follow what you suggest changing?

kalkyl: Something like the `Transfer` api that we use in the other DMA implementations? The problem i saw in this instance was how to make it sensible to use, with all the possible sequence playback variants / loops etc...

jonas-schievink: Yeah, my issue is that the `'static` bound on the buffer is missing, basically. I know that technically this is fine (at worst PWM will read garbage data), but that means that using the DMA traits is not really necessary here either.

nrf-hal-common/src/pwm.rs
```diff
@@ -973,22 +1090,56 @@ pub enum Error {
     BufferTooLong,
 }
 
-pub trait Instance: private::Sealed {}
+pub trait Instance: private::Sealed + Deref<Target = crate::pac::pwm0::RegisterBlock> {
+    const INTERRUPT: Interrupt;
 
-impl Instance for PWM0 {}
+    /// Provides access to the associated internal duty buffer for the instance.
+    fn buffer() -> &'static mut Cell<[u16; 4]>;
```
jonas-schievink, line 1097: Using this signature will create aliasing mutable references when called twice, can we return an immutable ref instead? Since it's a `Cell` you can still use interior mutability.

kalkyl: Good idea!

PR #233, opened 2020-09-27
Differential ADC

Yatekii:
This is just a preview on what the API could be like.
I use this successfully on my hardware I and I tested all of it (including the conversion methods). Surely there will be corner cases as usual :/

IF you folks like it, I will try and clean it up a little.

Also, it would be cool if anyone has an idea how to integrate the `OneShot` trait.
Currently the trait does only allow for a single input pin and you have to pass it everytime you start a read transaction, even tho that is not required.
This is why I have made a custom channel config API so you can configure the channel once and then just read.
I will extend this further such that the ADC multiplexer & sequencer can be used properly, even with multiple channels at once :)

Excited to hear your feedback :)

### DIFF
--- a/nrf-hal-common/src/saadc.rs
+++ b/nrf-hal-common/src/saadc.rs
@@ -40,7 +40,10 @@ use embedded_hal::adc::OneShot;
 
 pub use saadc::{
     ch::{
-        config::{GAIN_A as Gain, REFSEL_A as Reference, RESP_A as Resistor, TACQ_A as Time},
+        config::{
+            GAIN_A as Gain, MODE_A as Mode, REFSEL_A as Reference, RESP_A as Resistor,
+            TACQ_A as Time,
+        },
         pseln::PSELN_A as Pseln,
         pselp::PSELP_A as Pselp,
     },
@@ -107,6 +110,7 @@ impl<'a> Channel<'a> {
         let SaadcChannelConfig {
             reference,
             gain,
+            mode,
             resistor,
             time,
         } = config;
@@ -115,7 +119,7 @@ impl<'a> Channel<'a> {
             w.refsel().variant(reference);
             w.gain().variant(gain);
             w.tacq().variant(time);
-            w.mode().se();
+            w.mode().variant(mode);
             w.resp().variant(resistor);
             w.resn().bypass();
             w.burst().enabled();
@@ -213,6 +217,8 @@ pub struct SaadcChannelConfig {
     pub reference: Reference,
     /// Gain used to control the effective input range of the SAADC.
     pub gain: Gain,
+    /// The mode of the differential ADC
+    pub mode: Mode,
     /// Positive channel resistor control.
     pub resistor: Resistor,
     /// Acquisition time in microseconds.
@@ -266,6 +272,7 @@ impl Default for SaadcChannelConfig {
         SaadcChannelConfig {
             reference: Reference::VDD1_4,
             gain: Gain::GAIN1_4,
+            mode: Mode::SE,
             resistor: Resistor::BYPASS,
             time: Time::_20US,
         }
--- a/nrf-hal-common/src/saadc.rs
+++ b/nrf-hal-common/src/saadc.rs
@@ -30,7 +30,7 @@
 use crate::pac::{saadc_ns as saadc, SAADC_NS as SAADC};
 
 #[cfg(not(feature = "9160"))]
-use crate::pac::{saadc, SAADC};
+use crate::pac::{generic::Variant, saadc, SAADC};
 
 use core::{
     hint::unreachable_unchecked,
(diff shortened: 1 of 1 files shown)


jonas-schievink:
Looks like this needs a rebase

Yatekii:
Rebasing gave me some trouble. I hope merging is okay (I don't know what policy we have). If not, we can squash maybe.

I was using this code since Sept/Oct for inductive battery charging and nothing exploded so far :P

Sure it could use some extensions but I think the additional functionality warrants a merge :)

nrf-hal-common/src/saadc.rs
```diff
@@ -74,23 +80,129 @@ impl Saadc {
...
 
-        saadc.ch[0].config.write(|w| {
+        // Calibrate
+        saadc.tasks_calibrateoffset.write(|w| unsafe { w.bits(1) });
+        while saadc.events_calibratedone.read().bits() == 0 {}
+
+        Saadc { saadc }
+    }
+
+    pub fn channel(&mut self, n: usize) -> Channel {
```
jonas-schievink, line 90:
```suggestion
    pub fn channel(&mut self, n: usize) -> Channel<'_> {
        assert!(n < 8);
```

nrf-hal-common/src/saadc.rs
```diff
@@ -74,23 +80,129 @@ impl Saadc {
...
+        self.saadc.ch[self.channel]
+            .pselp
+            .write(|w| w.pselp().variant(pselp.into()));
+        self.saadc.ch[self.channel]
+            .pseln
+            .write(|w| w.pseln().variant(pseln.into()));
+    }
+
+    /// Sample channel `PIN` for the configured ADC acquisition time in differential input mode.
+    /// Note that this is a blocking operation.
```
jonas-schievink, line 138: It should probably not return `nb::Result` if it ends up blocking regardless

nrf-hal-common/src/saadc.rs
```diff
@@ -74,23 +80,129 @@ impl Saadc {
...
+        let mode = match self.saadc.ch[self.channel].config.read().mode().variant() {
+            Mode::SE => 1,
+            Mode::DIFF => 2,
+        };
+
+        let lsbs = match self.saadc.resolution.read().val().variant() {
+            Variant::Val(Resolution::_8BIT) => 256,
+            Variant::Val(Resolution::_10BIT) => 1024,
+            Variant::Val(Resolution::_12BIT) => 4096,
+            Variant::Val(Resolution::_14BIT) => 16384,
```
jonas-schievink, line 181:
```suggestion
            Variant::Val(Resolution::_8BIT) => 1 << 8,
            Variant::Val(Resolution::_10BIT) => 1 << 10,
            Variant::Val(Resolution::_12BIT) => 1 << 12,
            Variant::Val(Resolution::_14BIT) => 1 << 14,
```

nrf-hal-common/src/saadc.rs
```diff
@@ -74,23 +80,129 @@ impl Saadc {
...
+            return Err(nb::Error::Other(()));
+        }
+
+        // Second fence to prevent optimizations creating issues with the EasyDMA-modified `val`.
+        compiler_fence(SeqCst);
+
+        Ok(val)
+    }
+
+    pub fn as_micros(&self, value: i16) -> i32 {
```
jonas-schievink, line 171: Since we use raw integers everywhere (and not something like `uom`), let's clarify that this returns microvolts and has nothing to do with `Duration::as_micros`

```suggestion
    /// Converts a raw reading from this channel to µV.
    pub fn to_microvolts(&self, value: i16) -> i32 {
```

nrf-hal-common/src/saadc.rs
```diff
@@ -74,23 +80,129 @@ impl Saadc {
...
+
+        let reference = match self.saadc.ch[self.channel].config.read().refsel().variant() {
+            Reference::INTERNAL => (3, 5),
+            Reference::VDD1_4 => (33, 40),
+        };
+
+        (value as i64 * 1000 * 1000 / lsbs * gain.1 * reference.0 / gain.0 / reference.1) as i32
+    }
+
+    pub fn as_millis(&self, value: i16) -> i32 {
```
jonas-schievink, line 204:
```suggestion
    pub fn to_millivolts(&self, value: i16) -> i32 {
```

nrf-hal-common/src/saadc.rs
```diff
@@ -74,23 +80,129 @@ impl Saadc {
...
+    }
+}
+
+pub struct Channel<'a> {
+    saadc: &'a mut SAADC,
+    channel: usize,
+}
+
+impl<'a> Channel<'a> {
+    pub fn configure(
```
jonas-schievink, line 104: Does this *have* to be called for every channel that is used? What's the default state that would be used by the existing impls if this isn't called?

nrf-hal-common/src/saadc.rs
```diff
@@ -265,3 +418,32 @@ impl Channel<Saadc> for InternalVdd {
 #[cfg(not(feature = "9160"))]
 /// Channel that doesn't sample a pin, but the internal VDD voltage.
 pub struct InternalVdd;
+
+macro_rules! psel_mappings {
+    ( $($psel:ident => $pin:ident,)*) => {
+        $(
+            impl<STATE> Into<Pseln> for crate::gpio::p0::$pin<STATE> {
```
jonas-schievink, line 425: Is it okay to use pins in *any* state here, even `Disconnected`?

jonas-schievink (commented):
Thanks! Finally got around to reviewing this.

eflukx:
Would be a very welcome addition.. Is this change still alive?

Yatekii:
I kind of forgot about this. Mostly because I do not work on this paid anymore and my sparetime is already full with other stuff. If you want to you are more than welcome to take this over :)

PR #234, opened 2020-09-28
Working to impl embedded-dma for SPIM

jamesmunns:
CC @kalkyl, also fixed a SPIS item

### DIFF
--- a/nrf-hal-common/src/spim.rs
+++ b/nrf-hal-common/src/spim.rs
@@ -15,6 +15,7 @@ pub use embedded_hal::spi::{Mode, Phase, Polarity, MODE_0, MODE_1, MODE_2, MODE_
 pub use spim0::frequency::FREQUENCY_A as Frequency;
 
 use core::iter::repeat_with;
+use core::sync::atomic::Ordering;
 
 #[cfg(any(feature = "52832", feature = "52833", feature = "52840"))]
 use crate::pac::{SPIM1, SPIM2};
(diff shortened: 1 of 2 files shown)


kalkyl:
Looks great to me!

nrf-hal-common/src/spim.rs
```diff
@@ -361,8 +390,41 @@ where
...
+        (self.periph, self.pins)
+    }
+
+    pub fn dma_transfer_split<TxW, RxW, TxB, RxB>(
+        mut self,
+        tx_buffer: TxB,
+        mut rx_buffer: RxB,
+    ) -> Result<TransferSplit<T, TxB, RxB>, (Self, Error)>
+    where
+        TxB: ReadBuffer<Word = TxW>,
```
kalkyl, line 403:
```suggestion
        TxB: ReadBuffer<Word = TxW> + 'static,
```
Based on the matrix channel discussion today it looks like we need + 'static bounds on these buffers aswell

nrf-hal-common/src/spim.rs
```diff
@@ -361,8 +390,41 @@ where
...
+        RxB: WriteBuffer<Word = RxW>,
```
kalkyl, line 404:
```suggestion
        RxB: WriteBuffer<Word = RxW> + 'static,
```

nrf-hal-common/src/spim.rs
```diff
@@ -157,67 +166,89 @@ where
...
 
     /// Internal helper function to setup and execute SPIM DMA transfer.
     fn do_spi_dma_transfer(&mut self, tx: DmaSlice, rx: DmaSlice) -> Result<(), Error> {
+        self.start_spi_dma_transfer(&tx, &rx);
+
+        // Wait for END event.
+        //
+        // This event is triggered once both transmitting and receiving are
+        // done.
+        while !self.is_spi_dma_transfer_complete() {}
```
derekdreery, line 183: Is there an alternative to spinning here: could we setup an interrupt and go to sleep? I'm not an expert at all so apologies if this is a silly suggestion.

derekdreery:
I'm trying to use this SPI master abstraction, but am struggling because power consumption is very important and I can't see a way to disable the SPI in the current API. This patch would solve the problem for me I think.

PR #235, opened 2020-09-28
Update Product Specification links and add DK Reference Guide links

lachlan-smith:
The links to the product specifications were out of date on multiple devices. These have been updated and I have added links to the devices DK references guides as well. Formatted in a table with links shown as version numbers so it is clear what version the README is linking too.


Yatekii (approved):
LGTM, thank you!

PR #239, opened 2020-09-29
Add Timer task/event endpoint getters for PPI

kalkyl:
This PR adds task- and event endpoint getters to the Timer module so the timers can be used with PPI.

### DIFF
--- a/nrf-hal-common/src/timer.rs
+++ b/nrf-hal-common/src/timer.rs
@@ -148,30 +148,35 @@ where
 
     /// Returns reference to the `START` task endpoint for PPI.
     /// Starts timer.
+    #[inline(always)]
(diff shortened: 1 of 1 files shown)


nrf-hal-common/src/timer.rs
```diff
@@ -400,6 +502,69 @@ impl Instance for TIMER4 {
     }
 }
 
+#[cfg(any(feature = "52832", feature = "52833", feature = "52840"))]
+/// Adds task- and event PPI endpoint getters for CC[4] and CC[5] on supported instances.
```
jonas-schievink, line 506:
```suggestion
/// Adds task- and event PPI endpoint getters for CC[4] and CC[5] on supported instances.
#[cfg(any(feature = "52832", feature = "52833", feature = "52840"))]
```

PR #241, opened 2020-09-29
Improve RTC API to reflect that RTC requires LFCLK

### LINKED ISSUE #240: Update API to reflect that RTC requires LFCLK
From the PS: "The software has to explicitely start LFCLK before using the RTC."
With the current API it is possible to configure the RTC without having LFCLK enabled (and start debugging to investigate why it is not working).

timokroeger:
Fixes #240 
Includes an additional fix for a race condition in the event handling.

### DIFF
--- a/examples/rtc-demo/Cargo.toml
+++ b/examples/rtc-demo/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "rtc-demo"
+version = "0.1.0"
+authors = ["Timo Kröger"]
+edition = "2018"
+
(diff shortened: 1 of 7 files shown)


timokroeger:
I just realized this token based approach does only work for one RTC instance.
A second RTC instance cannot be created because the token was consumed already.

jonas-schievink:
The token could also be make `Copy` instead. This would only cause issues when you want to turn the clock back off, which is something that requires extra scrutiny in any case, I think.

PR #242, opened 2020-09-30
TWIM improvements.

### LINKED ISSUE #208: Disallow zero-length buffers in TWIM
The TWIM hal functions don't check for zero-length buffers, and write 0 to TXD.MAXCNT / RXD.MAXCNT. 

However Nordic's docs say the value should be in range `[1..0xFFFF]`, so 0 is not allowed: https://infocenter.nordicsemi.com/topic/ps_nrf52840/twim.html?cp=4_0_0_5_30_6_22#register.RXD.MAXCNT

In practice. doing a zero-length write or read makes the current implementation hang forever. Maybe this should be checked, and return an error like `Error::ZeroLengthBuffer`

Dirbaio:
- Derive Copy, Clone, Eq, PartialEq for Error
- Disallow zero-length buffers, fixes #208 
- Handle all errors. Previously only AddressNack was handled, and the transfer would hang if another error ocurred.
- remove chunking in `copy_write_then_read`.  The previous implementation would send a TWI start condition for each chunk, so the slave chip would see each as a separate transaction. This is not equivalent to `write_then_read`, which was the original goal of adding chunking. It seems TWIM has no way to send multiple buffers in a single transaction, therefore the ability to chunk is removed.

There are some issues left with TWIM:

- There's no way for the user use SUSPEND manually if they want to do a more complex repeated start sequence. Nordic's lib has a "hold" flag that suspends instead of stopping so you can do any combination of reads and writes. 

  If we do this it could be used for implementing `write_then_read` and  `copy_write_then_read` to reduce code duplication
  
  This is probably left for a future PR :) 

<details><summary>old, no longer apply</summary>
<p>
- `copy_write_then_read` splits written data into 1024 byte chunks and txs each individually. This actually causes a repeated start condition for each chunk, instead of a single start condition at the beginning like it would happen with `write_then_read`. 

  This should be at least documented, but I think in practice the chunking is useless because devices will take data after a repeated start as a new transaction (expecting you to send command/address bytes again).
  
  Maybe we should remove the chunking?

- `copy_write_then_read` should use SUSPEND. It seems that once LASTRX/LASTTX happens it's not OK to leave TWIM hanging, you must immediately STOP, SUSPEND, or STARTTX/STARTRX. This can be a problem if using TWIM from low priority code. Nordic says [here](https://infocenter.nordicsemi.com/topic/ps_nrf52840/twim.html?cp=4_0_0_5_30_3#concept_ebf_cwp_xr):

  > If a more complex repeated start sequence is needed and the TWI firmware drive is serviced in a low priority interrupt it may be necessary to use the SUSPEND task and SUSPENDED event to guarantee that the correct tasks are generated at the correct time. 

  This was a problem before with `write_then_read` too, but this PR changes it to use SHORTS.
</p></details>

### DIFF
--- a/nrf-hal-common/src/twim.rs
+++ b/nrf-hal-common/src/twim.rs
@@ -568,7 +568,7 @@ pub struct Pins {
     pub sda: Pin<Input<Floating>>,
 }
 
-#[derive(Debug)]
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
 pub enum Error {
     TxBufferTooLong,
     RxBufferTooLong,
--- a/nrf-hal-common/src/twim.rs
+++ b/nrf-hal-common/src/twim.rs
@@ -94,27 +94,17 @@ where
         Twim(twim)
     }
 
-    /// Write to an I2C slave.
-    ///
-    /// The buffer must have a length of at most 255 bytes on the nRF52832
-    /// and at most 65535 bytes on the nRF52840.
-    pub fn write(&mut self, address: u8, buffer: &[u8]) -> Result<(), Error> {
+    /// Set TX buffer, checking that it is in RAM and has suitable length.
+    unsafe fn set_tx_buffer(&mut self, buffer: &[u8]) -> Result<(), Error> {
         slice_in_ram_or(buffer, Error::DMABufferNotInDataMemory)?;
 
+        if buffer.len() == 0 {
+            return Err(Error::TxBufferZeroLength);
+        }
         if buffer.len() > EASY_DMA_SIZE {
             return Err(Error::TxBufferTooLong);
         }
 
-        // Conservative compiler fence to prevent optimizations that do not
-        // take in to account actions by DMA. The fence has been placed here,
-        // before any DMA action has started.
-        compiler_fence(SeqCst);
-
-        self.0
-            .address
-            .write(|w| unsafe { w.address().bits(address) });
-
-        // Set up the DMA write.
         self.0.txd.ptr.write(|w|
             // We're giving the register a pointer to the stack. Since we're
             // waiting for the I2C transaction to end before this stack pointer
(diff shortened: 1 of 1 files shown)


nrf-hal-common/src/twim.rs
```diff
@@ -132,37 +122,116 @@ where
...
-        self.0.errorsrc.write(|w| w.anack().bit(true));
+        Ok(())
+    }
 
-        // Start write operation.
-        self.0.tasks_starttx.write(|w|
-            // `1` is a valid value to write to task registers.
-            unsafe { w.bits(1) });
+    /// Set RX buffer, checking that it has suitable length.
+    fn set_rx_buffer(&mut self, buffer: &mut [u8]) -> Result<(), Error> {
```
jonas-schievink, line 129: Since the lifetime constraints aren't enforced, this should probably be `unsafe` (even if it's private)
```suggestion
    unsafe fn set_rx_buffer(&mut self, buffer: &mut [u8]) -> Result<(), Error> {
```

nrf-hal-common/src/twim.rs
```diff
@@ -94,27 +94,17 @@ where
         Twim(twim)
     }
 
-    /// Write to an I2C slave.
-    ///
-    /// The buffer must have a length of at most 255 bytes on the nRF52832
-    /// and at most 65535 bytes on the nRF52840.
-    pub fn write(&mut self, address: u8, buffer: &[u8]) -> Result<(), Error> {
+    /// Set TX buffer, checking that it is in RAM and has suitable length.
+    fn set_tx_buffer(&mut self, buffer: &[u8]) -> Result<(), Error> {
```
jonas-schievink, line 98:
```suggestion
    unsafe fn set_tx_buffer(&mut self, buffer: &[u8]) -> Result<(), Error> {
```

nrf-hal-common/src/twim.rs
```diff
@@ -284,92 +304,33 @@ where
...
-            // Note that that nrf52840 maxcnt is a wider
-            // type than a u8, so we use a `_` cast rather than a `u8` cast.
-            // The MAXCNT field is thus at least 8 bits wide and accepts the
-            // full range of values that fit in a `u8`.
-            unsafe { w.maxcnt().bits(rd_buffer.len() as _) });
-
-        // Clear address NACK.
-        self.0.errorsrc.write(|w| w.anack().bit(true));
+        // Clear events
+        self.0.events_stopped.write(|w| w);
```
jonas-schievink, line 312:
```suggestion
        self.0.events_stopped.reset();
```

nrf-hal-common/src/twim.rs
```diff
@@ -284,92 +304,33 @@ where
...
+        self.0.events_error.write(|w| w);
```
jonas-schievink, line 313:
```suggestion
        self.0.events_error.reset();
```

nrf-hal-common/src/twim.rs
```diff
@@ -193,57 +255,27 @@ where
...
-            // type than a u8, so we use a `_` cast rather than a `u8` cast.
-            // The MAXCNT field is thus at least 8 bits wide and accepts the
-            // full range of values that fit in a `u8`.
-            unsafe { w.maxcnt().bits(buffer.len() as _) });
+        self.set_rx_buffer(buffer)?;
 
-        // Clear address NACK.
-        self.0.errorsrc.write(|w| w.anack().bit(true));
+        // Clear events
+        self.0.events_stopped.write(|w| w);
```
jonas-schievink, line 261:
```suggestion
        self.0.events_stopped.reset();
```

nrf-hal-common/src/twim.rs
```diff
@@ -193,57 +255,27 @@ where
...
+        self.0.events_error.write(|w| w);
```
jonas-schievink, line 262:
```suggestion
        self.0.events_error.reset();
```

nrf-hal-common/src/twim.rs
```diff
@@ -132,37 +122,116 @@ where
...
+            return Err(Error::DataNack);
+        }
+        Ok(())
+    }
+
+    /// Wait for stop or error
+    fn wait(&mut self) {
+        loop {
+            if self.0.events_stopped.read().bits() != 0 {
+                self.0.events_stopped.write(|w| w);
```
jonas-schievink, line 188:
```suggestion
                self.0.events_stopped.reset();
```

nrf-hal-common/src/twim.rs
```diff
@@ -132,37 +122,116 @@ where
...
+                break;
+            }
+            if self.0.events_error.read().bits() != 0 {
+                self.0.events_error.write(|w| w);
```
jonas-schievink, line 192:
```suggestion
                self.0.events_error.reset();
```

jonas-schievink (commented):
Nice! Just left a few nits.

Dirbaio:
All nits addressed (commits amended).

PR #243, opened 2020-10-03
Fix race condition in RTC event handling

timokroeger:
Also add a minimal RTC example to show how to use compare events.


examples/rtc-demo/Embed.toml
```diff
@@ -0,0 +1,63 @@
...
+log_enabled = false
+# Where to save rtt history buffer relative to manifest path.
+log_path = "./logs"
+
+[default.gdb]
+# Whether or not a GDB server should be opened after flashing.
+# This is exclusive and cannot be used with RTT at the moment.
+enabled = false
+# The connection string in host:port format wher the GDB server will open a socket.
+# gdb_connection_string
```
Yatekii, line 63:
```suggestion
[default.rtt]
enabled = true
```

Yatekii (commented):
Hey, I added a suggestion to remove most of the Embed.toml

It is enough to just add the bits that deviate from the default configuration, which is basically just enabling RTT :) The rest will be merged in from the default config inside cargo-embed :)

jonas-schievink:
Thanks!

PR #245, opened 2020-10-06
gpio: change `Input<Disconnected>` to `Disconnected`.

Dirbaio:
This changes the default Pin state from `Pin<Input<Disconnected>>` to `Pin<Disconnected>`.

This is desirable because it makes `is_high`, `is_low` not available in `Disconnected` state, since they don't actually work. Nordic docs state this:

> The input buffer of a GPIO pin can be disconnected from the pin to enable power savings when the pin is not used as an input, see GPIO port and the GPIO pin details. Inputs must be connected to get a valid input value in the IN register, and for the sense mechanism to get access to the pin. 

I will add an `into_disconnected` function in a future PR, since it's desirable to disconnect input pins when not in use to save power.


jonas-schievink (approved):
Thanks! I thought you'd have to change more than that, neat.

PR #250, opened 2020-10-18
gpio: add conf() utility function.

Dirbaio:
This adds a `conf()` utility function to get the pin's config register.

Slight code duplication reduction :)


jonas-schievink:
Nice, thanks!

PR #252, opened 2020-10-20
Remove `Started`/`Stopped` type state from `Rtc`

timokroeger:
The type states make it difficult to enable and disable the RTC on demand with RTIC.
`Started`/`Stopped` types were only used to restrict the usage of `set_prescaler()` to a disabled RTC.
Pass the prescaler as argument to the costructor instead.


jonas-schievink (commented):
Thanks, this looks reasonable!

This does mean that you can't change the prescaler after creating the RTC though. I don't think that's a common thing to do, so it should be fine, but we might want to add a `fn free(self) -> RTC` to enable that use case again.

jonas-schievink:
Oh, it already has a `release` function that does exactly that.

PR #253, opened 2020-10-20
Fix TWIM warnings

timokroeger:
run CI


jonas-schievink:
Oh, we should probably use `RUSTFLAGS=-Dwarnings` in CI. Thanks!

PR #260, opened 2020-10-26
Make error enums #[non_exhaustive]

### LINKED ISSUE #257: Mark error enums as `#[non_exhaustive]`
This makes adding more variants non-breaking

jonas-schievink:
Closes #257

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@
 - 🦭 PPI traits ([#259]).
 - Various TWIM fixes and improvements - removes automatic transfer splitting ([#242]).
 - Remove typestate from RTC to make it easier to use ([#252]).
+- Mark error enums as `#[non_exhaustive]` ([#260]).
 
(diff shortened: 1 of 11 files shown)


Yatekii:
I don't like non-exhaustive enums personnally. Now I never have the guarantee to have handled all cases.

jonas-schievink:
Yeah, this might not be the best way to do it. We might want to audit all error types and complete them. Maybe move DMA usage errors to a common enum.

Yatekii:
I love the guarantees that Rust gives me with its enums, contrary to a classic Exception system, because it forces me to check all the cases. So if new features get added which require more error cases, or, even worse, a bug, I now worked around gets fixed and introduces a new error I wanna be notified about these changes, not silently burry them :)
Maybe that's just me :)

jonas-schievink:
Yeah, closing this in favor of just auditing out error types in a future release

PR #261, opened 2020-10-29
Also release owned `Pins` with `Usart::free()`

timokroeger:
Return ownership of the pins back to the user when calling `free()`
Useful to multiplex a single USARTE instance to multiple serial connections on different pins (for smaller NRF52 devices which only have one USARTE instance).

Alternative: Only take a reference to the pins in in `new()`.
Please let me know which one is the preferred approach.

### DIFF
--- a/nrf-hal-common/src/gpio.rs
+++ b/nrf-hal-common/src/gpio.rs
@@ -43,8 +43,8 @@ pub enum Port {
     Port1,
 }
 
+#[cfg(any(feature = "52833", feature = "52840"))]
 impl Port {
-    #[cfg(any(feature = "52833", feature = "52840"))]
     pub(crate) fn bit(&self) -> bool {
         match self {
             Port::Port0 => false,
@@ -52,7 +52,6 @@ impl Port {
         }
     }
 
-    #[cfg(any(feature = "52833", feature = "52840"))]
     pub(crate) fn from_bit(bit: bool) -> Port {
         if bit {
             Port::Port1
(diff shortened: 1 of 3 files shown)


jonas-schievink:
> Alternative: Only take a reference to the pins in in `new()`.

This would mean that `Usart<_>: 'static` never holds and prevent usage of `Usart` in RTIC resources, so we probably shouldn't do that.

I think we can do better than storing the `Pins` instance in every `Usart` though – what if we just read from the USART's pin configuration registers in `free` and create the pins from those? That way we don't have to pay for the extra `Pins` storage.

timokroeger:
Reading back the pins from the peripheral is a nifty idea. Going to try that tomorrow.

Yatekii:
Cool stuff, thanks for doing this! I required this a few times already but just for debugging things and I just used unsfe for that :P

> I think we can do better than storing the `Pins` instance in every `Usart` though – what if we just read from the USART's pin configuration registers in `free` and create the pins from those? That way we don't have to pay for the extra `Pins` storage.

I fully agree with this :)

nrf-hal-common/src/gpio.rs
```diff
@@ -51,6 +51,15 @@ impl Port {
             Port::Port1 => true,
         }
     }
+
+    #[cfg(any(feature = "52833", feature = "52840"))]
```
jonas-schievink, line 55: You can move this onto the impl and remove the cfgs on `fn bit`

jonas-schievink (commented):
Cool! Can you also add a changelog entry?

PR #266, opened 2020-11-18
Enabling and disabling the TWIM instance

hannes-hochreiner:
When a new TWIM instance is created, it is automatically enabled. This is very convenient in normal use. For a low power sensor, I found myself in the situation that I needed to deactivate (and later reactivate) the instance in order to safe energy. The simplest way I found to do this, was to add the "enable" and "disable" functions to the instance.

Please let me know, whether you would like to integrate this change into the main branch.

### DIFF
--- a/nrf-hal-common/src/twim.rs
+++ b/nrf-hal-common/src/twim.rs
@@ -86,14 +86,14 @@ where
         Twim(twim)
     }
 
-    /// Re-enable the instance after it was previously disabled.
-    pub fn enable(&mut self) {
-        self.0.enable.write(|w| w.enable().enabled());
-    }
-
     /// Disable the instance.
     pub fn disable(&mut self) {
         self.0.enable.write(|w| w.enable().disabled());
+    }
+
+    /// Re-enable the instance after it was previously disabled.
+    pub fn enable(&mut self) {
+        self.0.enable.write(|w| w.enable().enabled());
     }
 
     /// Set TX buffer, checking that it is in RAM and has suitable length.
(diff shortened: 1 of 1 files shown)


Dirbaio:
+1 to this! I have done hacks with `unsafe` in my project to do this from outside the hal, having it in the hal would be very nice!

- Maybe there should be some safety against trying to do transfers while disabled?
- For more power saving, disconnect the sda/scl pins on disable (dir=input, input=disconnected), and connect them again on enable.

nrf-hal-common/src/twim.rs
```diff
@@ -86,6 +86,16 @@ where
         Twim(twim)
     }
 
+    /// Enable the instance.
```
jonas-schievink, line 89:
```suggestion
    /// Re-enable the instance after it was previously disabled.
```

jonas-schievink: Maybe `enable` should be moved below `disable`, since you always call them in that order.

nrf-hal-common/src/twim.rs
```diff
@@ -86,6 +86,16 @@ where
...
+    pub fn enable(&mut self) {
+        self.0.enable.write(|w| w.enable().enabled());
+    }
+
+    /// Disable the instance.
```
jonas-schievink, line 94: This should probably mention what the impact of disabling it is.

jonas-schievink (commented):
Thanks, this looks reasonable!

nrf-hal-common/src/twim.rs
hannes-hochreiner: Addressed in commit 65cbd22.

PR #271, opened 2020-11-29
CHANGELOG: Explain what sealing means

dbrgn:
Wording is by @jonas-schievink (in the Matrix channel), I thought it was a good explanation and copied it 1:1 :slightly_smiling_face: 

Also, I removed the emoji. Even though they're cute, they're hard to recognize and understand by non-native English speakers and by visually impaired people.


jonas-schievink (approved):
Thanks!

PR #276, opened 2020-12-29
Make GPIO `Port` methods public

timokroeger:
This change allows external code to assign pins to peripherals.
Prevents hotfixes like: https://github.com/akiles/embassy/blob/267ec334ac766097efb4c847c14b6e2fde44f9af/embassy-nrf/src/buffered_uarte.rs#L158-L164


jonas-schievink:
Is there any chance of Nordic releasing an MCU with 3 or more GPIO ports? In that case this API wouldn't work anymore

timokroeger:
You are right!
What we actually need is something like a `fn psel_bits() -> u32` method for pins.
Closing for now.

PR #278, opened 2020-12-29
Fix #277 (re)setting GPIOTE interrupts for channels

### LINKED ISSUE #277: GPIOTE interrupt not triggering after it was enabled
I was trying to configure an input pin to trigger an interrupt. Following the example, I set up the pin and the gpiote.

```rust
let drdy_pin = port0.p0_30.into_floating_input().degrade();
let gpiote = Gpiote::new(device.GPIOTE);
gpiote.channel0().input_pin(&drdy_pin).hi_to_lo().enable_interrupt();
```
However, the interrupt never triggers.

Looking at the source code, the problems seems to be that the pin number is written to the `intenset` and `intenclr` registers. 
```rust
 pub fn enable_interrupt(&self) -> &Self {
        unsafe {
            self.gpiote
                .intenset
                .modify(|r, w| w.bits(r.bits() | self.pin.pin() as u32))
        }
        self
}
```
According to the Nordic documentation the register has a bit field for the channels. Therefore, it should be more like:
```rust
 pub fn enable_interrupt(&self) -> &Self {
        unsafe {
            self.gpiote
                .intenset
                .modify(|r, w| w.bits(r.bits() | 1 << self.channel as u32))
        }
        self
}
```
Same goes for disabling the interrupt.
I'll test later and prepare a PR. Feel free to assign the issue to me.

hannes-hochreiner:
I updated the write commands and the comments. I tested the code by setting two interrupts on a nRF52810.

### DIFF
--- a/nrf-hal-common/src/gpiote.rs
+++ b/nrf-hal-common/src/gpiote.rs
@@ -221,23 +221,21 @@ impl<'a, P: GpioteInputPin> GpioteChannelEvent<'_, P> {
         config_channel_event_pin(self.gpiote, self.channel, self.pin, EventPolarity::None);
         self
     }
-
-    /// Enables GPIOTE interrupt for pin.
+    /// Enables GPIOTE interrupt for channel.
     pub fn enable_interrupt(&self) -> &Self {
         unsafe {
             self.gpiote
                 .intenset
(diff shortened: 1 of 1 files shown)


jonas-schievink (approved):
Looks correct, thanks!

PR #280, opened 2021-01-04
Return correct error code in uarte start_read

lulf:
From reading the code, it seems to me that the wrong error code is returned when the provided rx_buffer u8::max size check fails.


jonas-schievink:
Thanks!

PR #281, opened 2021-01-06
Add embedded_hal::serial implementation for uarte

lulf:
I figured 2 existing PRs on the UARTE interface wasn't enough, so I created another one :laughing: 

This adds a sub-module named "serial" that splits TX and RX parts of
UART into separate types that implements the embedded_hal::serial
interfaces.

This allows using the nRF UART for drivers that rely on the
embedded_hal::serial traits.

The buffer management is primitive in order to support the semantics
of the more restrictive embedded_hal traits.

This is somewhat inspired by #102 and the TX/RX split is similar, but a lot more "dumb" without any advanced DMA and simplified only with the goal of having something that works with the embedded_hal traits with separate tx and rx.

### DIFF
--- a/nrf-hal-common/src/uarte.rs
+++ b/nrf-hal-common/src/uarte.rs
@@ -371,12 +371,12 @@ where
         )
     }
 
-    // Split into implementations of embedded_hal::serial traits. The buffers passed here must outlive any DMA transfers
+    // Split into implementations of embedded_hal::serial traits. The size of the slices passed to this method will determine the size of the DMA transfers performed.
     // that are initiated by the UartTx and UartRx.
     pub fn split<'a>(
         self,
-        tx_buf: &'a mut [u8],
-        rx_buf: &'a mut [u8],
+        tx_buf: &'static mut [u8],
+        rx_buf: &'static mut [u8],
     ) -> Result<(UarteTx<'a, T>, UarteRx<'a, T>), Error> {
         let tx = UarteTx::new(tx_buf)?;
         let rx = UarteRx::new(rx_buf)?;
@@ -453,6 +453,7 @@ where
 {
     _marker: core::marker::PhantomData<T>,
     tx_buf: &'a mut [u8],
+    written: u16,
 }
 
 /// Interface for the RX part of a UART instance that can be used independently of the TX part.
@@ -470,14 +471,19 @@ where
 {
     fn new(tx_buf: &'a mut [u8]) -> Result<UarteTx<'a, T>, Error> {
         slice_in_ram_or(tx_buf, Error::BufferNotInRAM)?;
-        if tx_buf.len() > 0 {
-            Ok(UarteTx {
-                _marker: core::marker::PhantomData,
-                tx_buf,
-            })
-        } else {
-            Err(Error::TxBufferTooSmall)
+        if tx_buf.len() == 0 {
+            return Err(Error::TxBufferTooSmall);
         }
+
+        if tx_buf.len() > EASY_DMA_SIZE {
+            return Err(Error::TxBufferTooLong);
+        }
+
+        Ok(UarteTx {
+            _marker: core::marker::PhantomData,
+            tx_buf,
+            written: 0,
+        })
     }
 }
 
@@ -487,14 +493,18 @@ where
 {
     fn new(rx_buf: &'a mut [u8]) -> Result<UarteRx<'a, T>, Error> {
         slice_in_ram_or(rx_buf, Error::BufferNotInRAM)?;
-        if rx_buf.len() > 0 {
-            Ok(UarteRx {
-                _marker: core::marker::PhantomData,
-                rx_buf,
-            })
-        } else {
-            Err(Error::RxBufferTooSmall)
+        if rx_buf.len() == 0 {
+            return Err(Error::RxBufferTooSmall);
         }
+
+        if rx_buf.len() > EASY_DMA_SIZE {
+            return Err(Error::RxBufferTooLong);
+        }
+
+        Ok(UarteRx {
+            _marker: core::marker::PhantomData,
+            rx_buf,
+        })
     }
 }
 
@@ -557,8 +567,9 @@ where
 
 pub mod serial {
 
-    ///! Implementation of the embedded_hal::serial::* traits for UartTx and UartRx.
+    ///! Implementation of the embedded_hal::serial::* and embedded_hal::blocking::serial::*  traits for UartTx and UartRx.
     use super::*;
+    use embedded_hal::blocking::serial as bserial;
     use embedded_hal::serial;
     use nb;
 
@@ -568,48 +579,21 @@ pub mod serial {
     {
         type Error = Error;
 
-        /// Write a single byte non-blocking. Returns nb::Error::WouldBlock if not yet done.
+        /// Write a single byte to the internal buffer. Returns nb::Error::WouldBlock if buffer is full.
         fn write(&mut self, b: u8) -> nb::Result<(), Self::Error> {
             let uarte = unsafe { &*T::ptr() };
 
-            // If txstarted is set, we are in the process of transmitting.
-            let in_progress = uarte.events_txstarted.read().bits() == 1;
+            // Prevent writing to buffer while DMA transfer is in progress.
+            if uarte.events_txstarted.read().bits() == 1 {
+                return Err(nb::Error::WouldBlock);
+            }
 
-            if in_progress {
-                self.flush()
+            let written = self.written as usize;
+            if written < self.tx_buf.len() {
+                self.tx_buf[written] = b;
+                self.written += 1;
+                Ok(())
             } else {
-                // Start a new transmission, copy value into transmit buffer.
-
-                self.tx_buf[0] = b;
-
-                // Conservative compiler fence to prevent optimizations that do not
-                // take in to account actions by DMA. The fence has been placed here,
-                // before any DMA action has started.
-                compiler_fence(SeqCst);
-
-                // Reset the events.
-                uarte.events_endtx.reset();
-                uarte.events_txstopped.reset();
-
-                // Set up the DMA write.
-                // We're giving the register a pointer to the tx buffer.
-                //
-                // The PTR field is a full 32 bits wide and accepts the full range
-                // of values.
-                uarte
-                    .txd
-                    .ptr
-                    .write(|w| unsafe { w.ptr().bits(self.tx_buf.as_ptr() as u32) });
-
-                // We're giving it a length of 1 to transmit 1 byte at a time.
-                //
-                // The MAXCNT field is 8 bits wide and accepts the full range of
-                // values.
-                uarte.txd.maxcnt.write(|w| unsafe { w.maxcnt().bits(1) });
-
-                // Start UARTE Transmit transaction.
-                // `1` is a valid value to write to task registers.
-                uarte.tasks_starttx.write(|w| unsafe { w.bits(1) });
                 Err(nb::Error::WouldBlock)
             }
         }
@@ -618,10 +602,12 @@ pub mod serial {
         fn flush(&mut self) -> nb::Result<(), Self::Error> {
             let uarte = unsafe { &*T::ptr() };
 
+            // If txstarted is set, we are in the process of transmitting.
             let in_progress = uarte.events_txstarted.read().bits() == 1;
-            let endtx = uarte.events_endtx.read().bits() != 0;
-            let txstopped = uarte.events_txstopped.read().bits() != 0;
+
             if in_progress {
+                let endtx = uarte.events_endtx.read().bits() != 0;
+                let txstopped = uarte.events_txstopped.read().bits() != 0;
                 if endtx || txstopped {
                     // We are done, cleanup the state.
                     uarte.events_txstarted.reset();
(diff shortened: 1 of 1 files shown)


Dirbaio:
This is unsound. A user can call `UarteRx::read()` to start DMA, then move the UarteRx instance. When a byte does arrive, the DMA will write it to the previous location, overwriting whatever's there.

To make this sound you need to require UarteRx and UarteTx to be pinned before use.

Dirbaio:
You also need to stop DMA on drop in UarteRx/UarteTx. 
Otherwise a user can start DMA, then drop the UarteRx. The DMA would overwrite one byte of whatever's in the same memory location when receiving a byte.

timokroeger:
You could also wait for the DMA to complete and have a blocking implementation that never returns `WouldBlock`.
This defeats the purpose of the `nb` return type but you do not have to worry about possible unsoundness.

jonas-schievink:
In that case we should just implement `blocking::serial::Write`, not `serial::Write`

lulf:
@timokroeger @jonas-schievink @Dirbaio Thank you for the review. I had not considered the case where it could be moved. 
For my particular use case (which is to make it possible to use nRF UARTE for a driver relying on serial::Write) I was originally looking for the non-blocking behavior.

Could an alternative approach to pinning, be to pass the tx and rx buffers to split() as I have pushed now? I realize this move more work to the user on initialization, but it seems more flexible and could potentially allow implementations to do larger DMA transfers if buffer sizes allow it. I'm not sure if it would still be sound/unsound (intuitively I think the compiler would prevent moving the mutable tx/rx buf slices?). 

I have also made an attempt at implementing Drop.

jonas-schievink:
They can still be leaked, and then the mutable slices can be freely used again, so this is still unsound

lulf:
@jonas-schievink I've added a required 'static lifetime for the buffers, I would like to avoid introducing the heapless crate as a dependency. I wasn't sure if it would be more appropriate to rely on the embedded_dma traits instead, wdyt? 

I've also made it so that one can provide a larger tx buffer and do multibyte DMA if desirable within the contract of the serial::Write trait.

nrf-hal-common/src/uarte.rs
```diff
@@ -372,6 +372,18 @@ where
             },
         )
     }
+
+    // Split into implementations of embedded_hal::serial traits. The size of the slices passed to this method will determine the size of the DMA transfers performed.
+    // that are initiated by the UartTx and UartRx.
```
jonas-schievink, line 377: Broken sentence (also please convert to real doc comments with `///`)

nrf-hal-common/src/uarte.rs
```diff
@@ -372,6 +372,18 @@ where
...
+    pub fn split<'a>(
```
jonas-schievink, line 378: `'a` is unbounded, should it be removed from `UarteTx`/`UarteRx`?

nrf-hal-common/src/uarte.rs
```diff
@@ -408,18 +422,324 @@ pub enum Error {
...
+    _marker: core::marker::PhantomData<T>,
+    rx_buf: &'a mut [u8],
+}
+
+impl<'a, T> UarteTx<'a, T>
+where
+    T: Instance,
+{
+    fn new(tx_buf: &'a mut [u8]) -> Result<UarteTx<'a, T>, Error> {
+        slice_in_ram_or(tx_buf, Error::BufferNotInRAM)?;
```
jonas-schievink, line 475: This is unnecessary, `&mut` buffers are always in RAM

nrf-hal-common/src/uarte.rs
```diff
@@ -408,18 +422,324 @@ pub enum Error {
...
+        })
+    }
+}
+
+impl<'a, T> UarteRx<'a, T>
+where
+    T: Instance,
+{
+    fn new(rx_buf: &'a mut [u8]) -> Result<UarteRx<'a, T>, Error> {
+        slice_in_ram_or(rx_buf, Error::BufferNotInRAM)?;
```
jonas-schievink, line 497: (same here)

nrf-hal-common/src/uarte.rs
```diff
@@ -408,18 +422,324 @@ pub enum Error {
...
+
+            // Ensure the above is done
+            compiler_fence(SeqCst);
+        }
+    }
+}
+
+pub mod serial {
+
+    ///! Implementation of the embedded_hal::serial::* and embedded_hal::blocking::serial::*  traits for UartTx and UartRx.
```
jonas-schievink, line 572:
```suggestion
/// Implementation of the embedded_hal::serial::* and embedded_hal::blocking::serial::*  traits for UartTx and UartRx.
pub mod serial {
```

jonas-schievink (commented):
Please move the duplicated implementation code into common functions, we don't really want to maintain two implementations of the whole TX and RX logic

nrf-hal-common/src/uarte.rs
jonas-schievink: This diff doesn't seem to have been applied

jonas-schievink: Oh actually the whole module should be removed and the impls put into the parent, there's nothing else in there

lulf: It was only intended to contain the embedded_hal::serial import. I can move it to the parent module, thats fine.

lulf: fixed.

nrf-hal-common/src/uarte.rs
```diff
@@ -8,7 +8,11 @@ use core::fmt;
 use core::ops::Deref;
 use core::sync::atomic::{compiler_fence, Ordering::SeqCst};
 
+use embedded_hal::blocking::serial as bserial;
 use embedded_hal::digital::v2::OutputPin;
+use embedded_hal::serial;
+
+use nb;
```
jonas-schievink, line 15: This shouldn't be needed, it's in the extern prelude

nrf-hal-common/src/uarte.rs
```diff
@@ -406,18 +434,249 @@ pub enum Error {
...
+        compiler_fence(SeqCst);
+
+        let in_progress = uarte.events_rxstarted.read().bits() == 1;
+        if in_progress && uarte.events_endrx.read().bits() == 0 {
+            return Err(nb::Error::WouldBlock);
+        }
+
+        if in_progress {
+            let b = self.rx_buf[0];
+            uarte.events_rxstarted.write(|w| w);
```
jonas-schievink, line 667: This should be equivalent to

```suggestion
            uarte.events_rxstarted.reset();
```

nrf-hal-common/src/uarte.rs
```diff
@@ -406,18 +432,249 @@ pub enum Error {
...
+{
+    type Error = Error;
+    fn read(&mut self) -> nb::Result<u8, Self::Error> {
+        let uarte = unsafe { &*T::ptr() };
+
+        compiler_fence(SeqCst);
+
+        let in_progress = uarte.events_rxstarted.read().bits() == 1;
+        if in_progress && uarte.events_endrx.read().bits() == 0 {
+            return Err(nb::Error::WouldBlock);
```
jonas-schievink, line 662: If we wait for ENDRX, we'll end up returning `WouldBlock` even if data has been received, right? That seems like an issue, especially since there is no timeout logic.

nrf-hal-common/src/uarte.rs
```diff
@@ -406,18 +434,249 @@ pub enum Error {
...
+        if uarte.events_txstarted.read().bits() == 1 {
+            return Err(nb::Error::WouldBlock);
+        }
+
+        if self.written < self.tx_buf.len() {
+            self.tx_buf[self.written] = b;
+            self.written += 1;
+            Ok(())
+        } else {
+            Err(nb::Error::WouldBlock)
```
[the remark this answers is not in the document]
lulf: I have usually written the application with having flush called in the event of nb::WouldBlock, but to your point I think we should probably call flush() here. I've pushed a commit for that .

nrf-hal-common/src/uarte.rs
lulf: Yes, that is a consequence of the embedded hal interface which doesn't allow you to read more than one byte at a time, so using rx_buf with size > 1 doesn't make sense right now. I was thinking to make this a bit more advanced where it would fill the rx_buf if it was empty, otherwise read from it until there are no more bytes to read. I'll push a fix for that.

lulf: Hmm, without some sort of timer, the RX would not be able read more than 1 byte at a time, as it seems it won't ever set  ENDRX until the expected number of bytes is received. I didn't originally want to require a timer for this, so I wonder if we should instead restrict to reading 1 byte no matter how big the rx_buf is, at least for now?

jonas-schievink: Yeah, I think we'll have to make it read one byte at a time.

nrf-hal-common/src/uarte.rs
```diff
@@ -326,7 +326,11 @@ fn stop_write(uarte: &uarte0::RegisterBlock) {
 
 /// Start a UARTE read transaction by setting the control
 /// values and triggering a read task.
-fn start_read(uarte: &uarte0::RegisterBlock, rx_buffer: &mut [u8]) -> Result<(), Error> {
+fn start_read(
+    uarte: &uarte0::RegisterBlock,
+    rx_buffer: &mut [u8],
+    nbytes: usize,
```
jonas-schievink, line 332: Why not just pass a shortened `rx_buffer` here?

nrf-hal-common/src/uarte.rs
lulf: Ok, done. Added a comment about the reason for that.

nrf-hal-common/src/uarte.rs
lulf: Oh, hello rust. Fixed.

nrf-hal-common/src/uarte.rs
```diff
@@ -676,7 +672,7 @@ where
             }
             Ok(self.rx_buf[0])
         } else {
-            start_read(&uarte, self.rx_buf, 1)?;
+            start_read(&uarte, &mut self.rx_buf[..1])?;
```
jonas-schievink, line 677: If we only ever use the first byte of the buffer, I don't see why we shouldn't take a `&'static mut [u8; 1]` in `split` instead of a slice

lulf: Yeah, we could. Though if we don't, we can potentially extend the UarteRx API later with a read() taking a rx_buffer and a timer without having to modify split(). I'm fine either way.

lulf: @jonas-schievink I've updated it to require a slice of size 1, I guess changing the API of split() to take any slice at a later point will not break anything.

hargoniX:
Hi, I stumbled upon this PR during the rewrite of the discovery book to micro:bit v1/v2. Basically I need to write code that is compatible with both the microbit v1 (nrf51) and v2 (nrf52833). Since the nrf52833-hal exposes only the UARTE peripheral the read() methods of both HALs are however not compatible with each other (UARTE requires an rx buffer, nrf51 just returns a single byte). Is there any chance this (or the other two PRs mentioned in the original PR message...which I couldn't find) will be merged soon / can I do something to make that happen? It's sadly really blocking me on the discovery book at the moment.

nrf-hal-common/src/uarte.rs
```diff
@@ -406,18 +440,250 @@ pub enum Error {
     BufferNotInRAM,
 }
 
-pub trait Instance: Deref<Target = uarte0::RegisterBlock> + sealed::Sealed {}
+pub trait Instance: Deref<Target = uarte0::RegisterBlock> + sealed::Sealed {
+    fn ptr() -> *const uarte0::RegisterBlock;
```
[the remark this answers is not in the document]
hargoniX: Is that it already? Shall I just fork @lulf then and "re PR" this I guess?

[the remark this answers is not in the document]
jonas-schievink: I should take a closer look at this, but that's one thing I noticed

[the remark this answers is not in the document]
hargoniX: So after reading a bit through the current implementation I saw that it still uses this Deref super trait https://github.com/nrf-rs/nrf-hal/blob/master/nrf-hal-common/src/uarte.rs#L452 (as well as other peripherals) so I'm quite unsure how to get rid off it? (I do get why the ptr() can be get rid off of course). I'm unfortunately not really familiar with the project so could you go into a bit of detail as to why we don't need it here?

nrf-hal-common/src/uarte.rs
```diff
@@ -406,18 +440,250 @@ pub enum Error {
...
+impl<T> Drop for UarteTx<T>
+where
+    T: Instance,
+{
+    fn drop(&mut self) {
+        let uarte = unsafe { &*T::ptr() };
+
+        let in_progress = uarte.events_txstarted.read().bits() == 1;
+        // Stop any ongoing transmission
+        if in_progress {
```
jonas-schievink, line 538: Shouldn't we reset `events_txstarted` in this `if` block? Otherwise it seems to me that it'd stay on indefinitely

nrf-hal-common/src/uarte.rs
```diff
@@ -406,18 +440,250 @@ pub enum Error {
...
+impl<T> Drop for UarteRx<T>
+where
+    T: Instance,
+{
+    fn drop(&mut self) {
+        let uarte = unsafe { &*T::ptr() };
+
+        let in_progress = uarte.events_rxstarted.read().bits() == 1;
+        // Stop any ongoing reception
+        if in_progress {
```
jonas-schievink, line 560: Same here for `events_rxstarted`

nrf-hal-common/src/uarte.rs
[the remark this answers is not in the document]
jonas-schievink: I'd like to remove the `Deref` supertrait, since it's essentially redundant with the `ptr` method. I guess this doesn't have to be done in *this* PR though.

hargoniX:
Continued in #343.

lulf:
@hargoniX Sorry, but the PR was not really abandoned from my POV. I didn't get any new reviews on it and was told it would need to wait for a breaking release in order to be merged. Then I didn't really see that it had conflicts until I saw your comment just now.

I've been on holiday the past few weeks, so I haven't been able to respond until now on your comment. 

However, I'm happy if you want to do the remaining fixes :) Next time, please ping the author directly and add a longer "timeout", in case the person you ping are temporarily away and actually intended to work on it.

PR #282, opened 2021-01-06
ppi: add clear_fork_task_endpoint

Dirbaio:
Add a function to clear the fork task endpoint of a PPI channel.

Nordic's docs are not very explicit on whether writing 0 to the `FORK[n].TEP` register is allowed. However nrfx's docs do say it's possible: https://github.com/NordicSemiconductor/nrfx/blob/d779b49fc59c7a165e7da1d7cd7d57b28a059f16/drivers/include/nrfx_ppi.h#L118


jonas-schievink (approved):
Looks good, thanks!

PR #285, opened 2021-01-12
Refactor Pin Selection and Reconstructions

timokroeger:
Moves the scattered `#[cfg()]` to a single place in the gpio module.


jonas-schievink (approved):
Great idea!

Issue #289, opened 2021-02-09
[docs] UART data received outside a blocking `Uarte::{read,read_timeout}` call is discarded


japaric:
I don't consider this a bug per se but I find the behavior surprising so I think *it* (the issue title) should be documented in the API docs.

### Steps to reproduce

Running this program

``` rust
let mut rx_buffer = [0];
defmt::info!("GO");
// do `$ echo hi > /dev/ttyACM0` in the middle of the delay
asm::delay(100_000_000);
defmt::info!("WAIT");

// do `$ echo yo > /dev/ttyACM0` in the middle of the infinite loop
loop {
    const TIMEOUT: u32 = 1_000_000;

    let res = uarte.read_timeout(&mut rx_buffer, &mut timer, TIMEOUT);

    if res.is_ok() {
        defmt::info!("{:?}", rx_buffer);
    } else {
        defmt::error!("timeout");
    }
}
```

And performing these serial operations:

``` console
$ # in parallel as specified in the source code
$ echo hi > /dev/ttyACM0
$ echo yo > /dev/ttyACM0
```

Produces this output:

``` console
$ cargo r
0.000000 INFO  WAIT
0.000000 INFO  GO
0.000000 ERROR timeout
0.000000 INFO  [121]
0.000000 INFO  [111]
0.000000 INFO  [13]
0.000000 INFO  [10]
```

`[121, 111, 13, 10]` = `b"yo\r\n"`
The data sent by the first `echo` command (`b"hi\r\n"`) is discarded -- I guess at the hardware level -- so it's never returned by the `read_timeout` operation
The same behavior is observed with the `read` API.

This is different from other HALs / microcontrollers where incoming serial data is stored in a hardware buffer and `embedded_hal::serial::Read` implementations can still return that old data (module overrun errors)

P.S. this behavior may be present in other `read` APIs (I2C)? I haven't tested that

### Initialization code

<details>

``` rust
let p = unwrap!(pac::Peripherals::take());
let port0 = p0::Parts::new(p.P0);

let mut timer = Timer::one_shot(p.TIMER0);

let rxd = port0.p0_08.into_floating_input().degrade();
let txd = port0.p0_06.into_push_pull_output(Level::High).degrade();

let pins = Pins {
    rxd,
    txd,
    cts: None,
    rts: None,
};

let mut uarte = Uarte::new(p.UARTE0, pins, Parity::EXCLUDED, Baudrate::BAUD9600);
```

</details>

[...]

PR #291, opened 2021-02-10
test the nrf52840 HAL using defmt-test

japaric:
this PR adds tests that exercise the GPIO API
3 connections are required:
- P0.03 <-> GND
- P0.04 <-> VDD
- P0.28 <-> P0.29

the first two are used to test the Input API against reference voltages
the short-circuit connection is used to test the Input and Output API using the input API that was
previously tested

I tried testing the Uarte API but ran into #289. It seems to me that we can't test TX/RX loopback or overrun errors with the existing `read` API


.cargo/config
```diff
@@ -1,8 +1,2 @@
 [alias]
 xtask = "run -p xtask --"
-
```
japaric, line 3: moved these into `examples/.cargo/config` because rustflags in `.cargo/config` compose and w/o the move I ran into knurling-rs/probe-run#143

jonas-schievink:
> the first two are used to test the Input API against reference voltages

Would it be possible to use the internal pull-up/down resistors to create reference voltages instead? That way it keeps the external wiring to a minimum.

japaric:
> Would it be possible to use the internal pull-up/down resistors to create reference voltages instead? That way it keeps the external wiring to a minimum.

that's a different test in my opinion (and it's already there in the `pulldown_is_low` and `pullup_is_high`).

if we do that and remove `vdd_is_high` and `gnd_is_low` then we would have less test coverage because we would not be testing the *floating* input mode *independently*. (we have the `set_low_is_low` and `set_high_is_high` tests but those are testing GPIO output against a GPIO input API *that's known to work*)

PR #292, opened 2021-02-14
Add USB tests

Disasm:


Disasm:
How to run these?
For the test_class example you need to clone [usb-device](https://github.com/mvirkkunen/usb-device), cd to it and run `cargo test` with the connected hardware running the `test_class` example. Or first observe dmesg, it shows errors right now.

For the serial example you need to clone [serial-threadmark](https://github.com/mvirkkunen/serial-threadmark) and run `cargo run --release /dev/serial/by-id/usb-Fake_company_Serial_port_TEST-if00 1000000` a couple of times. None of the runs should fail.

jonas-schievink:
these were added in https://github.com/nrf-rs/nrf-hal/pull/295, so closing this

PR #295, opened 2021-02-17
USB support

Disasm:
This PR is based on https://github.com/nrf-rs/nrf-hal/pull/144 and previous attempts.


Disasm:
Current problems:
* [ ] IN packets sometimes get lost. This happens because there is no status tracking implemented for IN endpoints and `write` never returns `WouldBlock`.
* [ ] full IN SETUP transactions with size == 0 (mod ep0_size) are not handled correctly. The size of the transaction should be tracked and status stage should be signaled when the transaction finishes (instead of sending a ZLP).

nrf-hal-common/src/usbd/mod.rs
```diff
@@ -0,0 +1,772 @@
...
+        buf[7] = regs.wlengthh.read().wlengthh().bits();
+
+        ep0_state.direction = match regs.bmrequesttype.read().direction().is_host_to_device() {
+            false => UsbDirection::In,
+            true => UsbDirection::Out,
+        };
+
+        ep0_state.is_set_address = (buf[0] == 0x00) && (buf[1] == 0x05);
+
+        if regs.bmrequesttype.read().direction().is_host_to_device() {
```
wez, line 199: nit: this is a bit shorter and feels easier to understand after the assignment just above

```suggestion
        if ep0_state.direction == UsbDirection::Out  {
```

nrf-hal-common/src/usbd/mod.rs
```diff
@@ -0,0 +1,772 @@
...
+    periph: Mutex<USBD>,
+    // argument passed to `UsbDeviceBuilder.max_packet_size_0`
+    max_packet_size_0: u16,
+    unalloc_buffers: &'static mut [u8],
+    bufs: Buffers,
+    used_in: u8,
+    used_out: u8,
+    iso_in_used: bool,
+    iso_out_used: bool,
+    ep0_state: Mutex<Cell<EP0State>>,
```
wez, line 80: Heh, I have literally the same field name and struct name in my non-working copy of this :-)
I only got as far as effectively the same thing as your `TransferState`, but structured a little differently.

nrf-hal-common/src/usbd/mod.rs
```diff
@@ -0,0 +1,772 @@
...
+                let regs = self.periph.borrow(cs);
+
+                let ep0_state = self.ep0_state.borrow(cs).get();
+
+                if ep0_state.is_set_address {
+                    // Inhibit
+                    return true;
+                }
+
+                if ep0_state.direction == UsbDirection::Out {
```
wez, line 433: makes sense!

nrf-hal-common/src/usbd/mod.rs
```diff
@@ -0,0 +1,772 @@
...
+                regs.shorts.modify(|_, w| {
+                    if is_short_packet {
+                        w.ep0datadone_ep0status().set_bit()
+                    } else {
+                        w.ep0datadone_ep0status().clear_bit()
+                    }
+                });
+
+                // Hack: send status stage if the IN transfer is not acknowledged after a few frames
+                let frame_counter = regs.framecntr.read().framecntr().bits();
```
wez, line 503: ah, that's a neat way to detect this!

nrf-hal-common/src/usbd/mod.rs
```diff
@@ -0,0 +1,772 @@
...
+                let is_short_packet = buf.len() < self.max_packet_size_0 as usize;
+                regs.shorts.modify(|_, w| {
+                    if is_short_packet {
+                        w.ep0datadone_ep0status().set_bit()
+                    } else {
+                        w.ep0datadone_ep0status().clear_bit()
+                    }
+                });
+
+                // Hack: send status stage if the IN transfer is not acknowledged after a few frames
```
wez, line 502:
```suggestion
                // Hack: trigger status stage if the IN transfer is not acknowledged after a few frames,
                // so record the current frame here; the actual test and status stage activation happens
                // in the poll method.
```

nrf-hal-common/src/usbd/mod.rs
```diff
@@ -0,0 +1,772 @@
...
+        interrupt::free(|cs| {
+            let regs = self.periph.borrow(cs);
+
+            // Control EP 0 is special
+            if i == 0 {
+                // Control setup packet is special, since it is put in registers, not a buffer.
+                if regs.events_ep0setup.read().events_ep0setup().bit_is_set() {
+                    regs.events_ep0setup.reset();
+
+                    let ep0_state = unsafe { &mut *self.ep0_state.borrow(cs).as_ptr() };
```
wez, line 546: can you avoid the unsafe bits?  They look a bit magical/noisy:

```suggestion
                    let mut ep0_state = self.ep0_state.borrow(cs);
```

then just pass `&mut ep0_state` below?

Alternatively, pass the `cs: &CriticalSection` to `read_control_setup` and borrow `ep0_state` in there?

nrf-hal-common/src/usbd/mod.rs
```diff
@@ -0,0 +1,772 @@
...
+                &regs.epout7,
+            ];
+            epout[i].ptr.write(|w| unsafe { w.bits(buf.as_ptr() as u32) });
+            // MAXCNT must match SIZE
+            epout[i].maxcnt.write(|w| unsafe { w.bits(size) });
+
+            dma_start();
+            regs.events_endepout[i].reset();
+            regs.tasks_startepout[i].write(|w| w.tasks_startepout().set_bit());
+            while regs.events_endepout[i].read().events_endepout().bit_is_clear() {}
```
wez, line 597: I suppose that this busy wait works more reliably and should be fine for default buffer sizes.
I also appreciate avoiding an extra buffer copy just to prepare for DMA, so this feels good to me.

Are we concerned about latency/blocking here for larger buffer sizes?

nrf-hal-common/src/usbd/mod.rs
```diff
@@ -0,0 +1,772 @@
...
+                    return PollResult::Resume;
+                } else {
+                    regs.events_usbevent.reset();
+                }
+            }
+
+            if regs.events_sof.read().events_sof().bit_is_set() {
+                regs.events_sof.reset();
+
+                // Check if we have a timeout for EP0 IN transfer
```
wez, line 680: :+1:

wez (commented):
Looks like you're making great progress!  I got about 60% of the way that you did; I'm pleased to see some similarities with the stuff in my local hack branch, and happy that spending the time on that means that I can more easily reason about your PR!

nrf-hal-common/src/usbd/mod.rs
Disasm: Free platform-agnostic delay :)

nrf-hal-common/src/usbd/mod.rs
Disasm: I don't know about nrf52 DMA, but on STM32s DMA usually works faster than memcpy. The only way we can optimize this even more is to make this code async (which is not something supported by `usb-device`). Also I don't think nrf52 family has chips with HighSpeed support, so for max 64-byte packet sizes latency/blocking shouldn't be a problem.

Disasm:
Current status: serial stress test works better but still fails. One of the IN packets is not followed by EPDATASTATUS change, so "busy" flag is not erased and driver ends up in a locked state.

nrf-hal-common/src/usbd/mod.rs
```diff
@@ -0,0 +1,714 @@
...
+impl<'c> Usbd<'c> {
+    /// Creates a new USB bus, taking ownership of the raw peripheral.
+    ///
+    /// # Parameters
+    ///
+    /// * `periph`: The raw USBD peripheral.
+    #[inline]
+    pub fn new<L, LSTAT>(
+        periph: USBD,
+        _clocks: &'c Clocks<ExternalOscillator, L, LSTAT>,
```
chris-ricketts, line 90: I'm testing out this branch in conjunction with [nrf-softdevice](https://github.com/embassy-rs/nrf-softdevice) with the goal of writing some BLE dongle firmware. The SoftDevice takes ownership of the clock peripheral and makes it impossible to use `nrf52840_hal::Clocks`. I use the [sd_clock_hfclk_request](https://github.com/embassy-rs/nrf-softdevice/blob/77db337aef97c58d0824b160778b100a31a46ebf/softdevice/s140/headers/nrf_soc.h#L698) function to request that the external oscillator be used and poll until it is done. Am I right in thinking the `_clock` is used here as a compile time check to ensure that this is done? Could there be another function like `new_with_clock_set` that has a runtime check to increase flexibility?

Disasm:
There was a problem in the serial example: it doesn't call `pool()` while trying to send the whole buffer. Since `pool()` was the only place where IN endpoints are unblocked, this resulted in an endpoint that was blocked forever.

Disasm:
I tested this version of the driver:
* [x] usb-device tests passed
* [x] serial test passed
* [x] almost all USB compliance tests passed: two problems relate to usb-device (alt.settings and LPM are not supported)

examples/usb/examples/serial.rs
```diff
@@ -0,0 +1,79 @@
...
+        .is_vbus_present()
+    {}
+
+    // wait until USB 3.3V supply is stable
+    while !periph
+        .POWER
+        .events_usbpwrrdy
+        .read()
+        .events_usbpwrrdy()
+        .bit_is_clear()
```
Dirbaio, line 31: This waits for the bit to be 0. Shouldn't it wait for 1 instead?

Dirbaio: Changing it to wait for 1 makes it wait forever :rofl:. It seems it's not necessary anyway.

Disasm: I also don't think this code is not needed as devices should work without VBUS detection mechanism.

Dirbaio: I have a board where 5v is permanently wired to VBUS (so the nrf thinks usb is always plugged in) and yes, it still works fine, including hotplugging.

Disasm: Removed useless code, thank you :)

nrf-hal-common/src/usbd/mod.rs
```diff
@@ -0,0 +1,718 @@
...
+                PollResult::None
+            }
+        })
+    }
+
+    fn force_reset(&self) -> usb_device::Result<()> {
+        interrupt::free(|cs| {
+            let regs = self.periph.borrow(cs);
+            regs.usbpullup.write(|w| w.connect().disabled());
+            // TODO delay needed?
```
jonas-schievink, line 712: I couldn't really find a conclusive answer about the appropriate delay here, but I think there should be a delay of one millisecond here, to make sure that any ongoing transmission (where the host controls the bus and would not notice the device removing the pullup) ends before the pull-up is reattached.

jonas-schievink (changes requested):
Did a first pass, this looks pretty good so far!

nrf-hal-common/src/usbd.rs
```diff
@@ -0,0 +1,21 @@
+use crate::clocks::ExternalOscillator;
+use crate::pac::USBD;
+use crate::Clocks;
+
+pub use nrf_usbd::Usbd;
+
+#[allow(dead_code)] // fields are unused and only hold ownership
```
Dirbaio, line 7: minor nit: I think if you name the field `_clocks` the warning goes away

nrf-hal-common/src/usbd.rs
```diff
@@ -0,0 +1,21 @@
...
+pub struct UsbPeripheral<'a> {
+    usbd: USBD,
+    clocks: &'a (),
```
Dirbaio, line 10: `&()` is not ZST even if `()` is. Maybe use `PhantomData<&'a ()>`?

Dirbaio (changes requested):
Looks great! Excited to finally have USB support in :) 

Commented a few minor nits

Dirbaio (approved):
:rocket: 

bors r+

PR #298, opened 2021-03-01
saadc: Clear events_calibratedone before calibration

Dirbaio:
Without this, if you initialize SAADC for a second time, the events_calibratedone is still set from before, so the while loop doesn't wait for calibration to be done. This causes garbage readings if you read immediately after, because calibration is still running.


Yatekii:
LGTM :)

bors r+

Yatekii (approved):

PR #301, opened 2021-03-09
[RFC] flip the GPIO types inside-out

Dirbaio:
Currently, the PAC+HAL ecosystem has two kinds of structs:

- Owned peripheral structs: These structs denote *ownership* over a chip's resource. Only one instance can exist of each at a time, and owning it represents exclusive access to the peripheral and its registers. They have no *functionality* by themselves. Example: `nrf52840_pac::UARTE0`
- Driver structs: These are constructed with owned peripheral structs, and *add functionality* to them. A given peripheral may have multiple drivers for different purposes (ex: ble and ieee802154 for RADIO), even in different crates. Example: `nrf52840_hal::Uarte`

It is very advantageous for the ecosystem to have this division. Having a single "owned peripheral" struct for a peripheral across the ecosystem acts as a single "point of coordination" to ensure no two drivers for the same peripheral can be created at the same time, even across different crates from different projects. The "owned peripheral" struct having no functionality allows each driver to build from scratch without having to "fight" any builtin functionality.

However, this is not the case for GPIO pin structs.

nrf-hal currently has PX_XX structs that do BOTH: they represent ownership over a pin, AND grant functionality to use the pin as input or output. The core idea behind this RFC is that we *can* also have this separation for GPIO pin structs, which brings us the above detailed benefits, and improving the implementation in the process.

How do we do this? 

First, we define the *owned structs*:
```rust
// Represents a concrete pin, zero-sized
struct P0_00 {}
struct P0_01 {}
//.. etc

// represents any pin, type-erased
struct AnyPin { pin_port: u8 }
```

To make use of these, we implement a `Pin` trait for all the structs:
```rust
pub trait Pin {
    fn pin_port(&self) -> u8;
}

impl Pin for AnyPin { fn pin_port(&self) -> u8 { self.pin_port }}
impl Pin for P0_00  { fn pin_port(&self) -> u8 { 0 }}
impl Pin for P0_01  { fn pin_port(&self) -> u8 { 1 }}
// etc
```

The structs above don't implement any functionality, they simply represent ownership of the pins. We can now implement *drivers* for them, such as Input or Output:

```rust
pub struct Input<T: Pin> {
    pin: T,
}

impl<T: Pin> Input<T> {
    pub fn new(pin: T, pull: Pull) -> Self {
        pin.conf().write(|w| { /* configure pin as input */ });
        Self { pin }
    }
}

impl<T: Pin> Drop for Input<T> {
    fn drop(&mut self) {
        // disconnect pin
        self.pin.conf().reset();
    }
}

impl<T: Pin> InputPin for Input<T> {
  // is_high, is_low
}
```

`Input` is a *driver* for structs like `P0_13`. Just like `nrf52840_hal::Uarte` is a driver for `nrf52840_pac::UARTE3`. It takes an owned peripheral struct and exposes some functionality.

Therefore, this flips the types of GPIOs "inside out" :). Before it was `P1_04<Input<..>>`, now it is `Input<P1_04>`. 

Drivers like Uarte would take owned pins, instead of Input or Output, and do the pin configuration themselves. This allows drivers to set up any configuration they want, and frees the user from doing it.

---

This PR contains working PoC code for this idea. I have updated the UARTE driver to showcase how would a driver work with the new API. The `blinky-button-demo` and `uarte` examples are functional.

Example GPIO usage (full code in `blinky-button-demo` example)

```rust
    let port0 = hal::gpio::p0::Parts::new(p.P0);
    let button = Input::new(port0.p0_11, Pull::Up);
    let mut led = Output::new(port0.p0_13, Level::Low);
    loop {
        if button.is_high().unwrap() {
            led.set_high().unwrap();
        } else {
            led.set_low().unwrap();
        }
    }
```

Example UARTE usage (full code in `uarte` example):

```rust
    let port0 = hal::gpio::p0::Parts::new(p.P0);
    let pins = uarte::Pins {
        rxd: port0.p0_08.degrade(),
        txd: port0.p0_06.degrade(),
        rts: None,
        cts: None,
    };
    let mut u = Uarte::new(p.UARTE0, pins, Parity::EXCLUDED, Baudrate::BAUD115200);
    u.write(&whatever).unwrap();
```

Advantages:

- Simpler public API:
  - The user doesn't have to configure pins as input/output/opendrain/etc to use a peripheral.
  - No need for a "Disconnected" typestate, having a "bare" `AnyPin` or `PX_XX` struct implies the Pin is disconnected. `Input` and `Output` disconnect the pin on drop.
- Simpler implementation in nrf-hal:
  - No code duplication between `AnyPin` and `PX_XX`. All the logic for Input and Output is defined once, no need for an uber-macro.
  - Drivers setting "weird" pin configurations that are not supported by the public API is no longer a [hack](https://github.com/nrf-rs/nrf-hal/blob/aae17943efc24baffe30b9a808fb099836e6bf94/nrf-hal-common/src/twim.rs#L54). Drivers configuring GPIOs it is now the standard way of doing things.
  - Drivers can be generic over `AnyPin` and `PX_XX`  by taking `P: Pin`.

Disadvantages:

- If user wants to use a pin as GPIO input, then wants to use it as a peripheral input, the peripheral driver will do one extra register write to set it as input even if not needed.

### DIFF
--- a/examples/blinky-button-demo/src/main.rs
+++ b/examples/blinky-button-demo/src/main.rs
@@ -3,9 +3,10 @@
 
 use embedded_hal::digital::v2::InputPin;
 use embedded_hal::digital::v2::OutputPin;
-use nrf52832_hal as hal;
-use nrf52832_hal::gpio::Level;
+use nrf52840_hal as hal;
 use rtt_target::{rprintln, rtt_init_print};
+
+use hal::gpio::{Input, Level, Output, Pull};
 
 #[panic_handler] // panicking behavior
 fn panic(_: &core::panic::PanicInfo) -> ! {
@@ -19,8 +20,8 @@ fn main() -> ! {
     rtt_init_print!();
     let p = hal::pac::Peripherals::take().unwrap();
     let port0 = hal::gpio::p0::Parts::new(p.P0);
-    let button = port0.p0_13.into_pullup_input();
-    let mut led = port0.p0_17.into_push_pull_output(Level::Low);
+    let button = Input::new(port0.p0_11, Pull::Up);
+    let mut led = Output::new(port0.p0_13, Level::Low);
 
     rprintln!("Blinky button demo starting");
     loop {
--- a/examples/uarte/Cargo.toml
+++ b/examples/uarte/Cargo.toml
@@ -1,14 +1,14 @@
 [package]
-name = "blinky-button-demo"
+name = "uarte"
 version = "0.1.0"
-authors = ["Andres O. Vela"]
 edition = "2018"
+publish = false
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
 cortex-m = "0.6.2"
-cortex-m-rt = "0.6.12"
+cortex-m-rt = "0.6.7"
 rtt-target = {version = "0.2.0", features = ["cortex-m"] }
 nrf52840-hal = { features = ["rt"], path = "../../nrf52840-hal" }
 
--- a/examples/uarte/Embed.toml
+++ b/examples/uarte/Embed.toml
@@ -31,7 +31,7 @@ halt_afterwards = false
 
 [default.general]
 # The chip name of the chip to be debugged.
-chip = "nRF52840_xxAA"
+chip = "nRF52840"
 # A list of chip descriptions to be loaded during runtime.
 chip_descriptions = []
 # The default log level to be used. Possible values are one of:
--- a/examples/uarte/src/main.rs
+++ b/examples/uarte/src/main.rs
@@ -1,0 +1,46 @@
+#![no_main]
+#![no_std]
+
+use nrf52840_hal as hal;
+use rtt_target::{rprintln, rtt_init_print};
+
+use hal::gpio::PinExt;
+use hal::uarte::{self, Baudrate, Parity};
+use hal::Uarte;
+
+#[panic_handler] // panicking behavior
+fn panic(_: &core::panic::PanicInfo) -> ! {
+    loop {
+        cortex_m::asm::bkpt();
+    }
+}
+
+#[cortex_m_rt::entry]
+fn main() -> ! {
+    rtt_init_print!();
+    let p = hal::pac::Peripherals::take().unwrap();
+    let port0 = hal::gpio::p0::Parts::new(p.P0);
+
+    let pins = uarte::Pins {
+        rxd: port0.p0_08.degrade(),
+        txd: port0.p0_06.degrade(),
+        rts: None,
+        cts: None,
+    };
+    rprintln!("dgjaksdjf!");
+
+    let mut u = Uarte::new(p.UARTE0, pins, Parity::EXCLUDED, Baudrate::BAUD115200);
+
+    rprintln!("hai!");
+
+    let mut salute = *b"Hello there!";
+    u.write(&salute).unwrap();
+
+    rprintln!("written!");
+
+    loop {
+        let mut rx_buf = [0u8; 1];
+        u.read(&mut rx_buf).unwrap();
+        rprintln!("readed!");
+    }
+}
--- a/nrf-hal-common/src/gpio.rs
+++ b/nrf-hal-common/src/gpio.rs
@@ -1,35 +1,16 @@
-use core::marker::PhantomData;
-
-/// Disconnected pin in input mode (type state, reset value).
-pub struct Disconnected;
-
-/// Input mode (type state).
-pub struct Input<MODE> {
-    _mode: PhantomData<MODE>,
-}
-
-/// Floating input (type state).
-pub struct Floating;
-/// Pulled down input (type state).
-pub struct PullDown;
-/// Pulled up input (type state).
-pub struct PullUp;
-
-/// Output mode (type state).
-pub struct Output<MODE> {
-    _mode: PhantomData<MODE>,
-}
-
-/// Push pull output (type state).
-pub struct PushPull;
-/// Open drain output (type state).
-pub struct OpenDrain;
-
 /// Represents a digital input or output level.
 #[derive(Debug, Eq, PartialEq)]
 pub enum Level {
     Low,
     High,
+}
+
+/// Represents a pull setting for an input.
+#[derive(Debug, Eq, PartialEq)]
+pub enum Pull {
+    None,
+    Up,
+    Down,
 }
 
 /// A GPIO port with up to 32 pins.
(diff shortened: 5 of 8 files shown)


bradleyharden:
@Dirbaio, a few months ago I refactored the GPIO module for `atsamd-hal`. I really liked how it turned out. I've been thinking about gathering some of my ideas for it as an RFC, but I wasn't sure where to propose them. I think this is a good opportunity.

Overall, I think my approach has some similarities to yours, but I don't abandon type-level tracking of the pin modes. I think I found a way to make it ergonomic, which is the main feature I wanted to highlight.

For the sake of brevity, I'm going to skim over a **lot** of details. What I describe here is a simplified version of the actual implementation. Most of the added complexity comes from supporting both a type-level, compile-time API and a type-erased, run-time API without duplicating code.

Start by defining marker types for each pin.

```rust
pub enum PA00 {}
pub enum PA01 {}
// ...
pub enum PB31 {}
```

Each type will implement a `PinId` trait.

```rust
pub trait PinId {}

impl PinId for PA00 {}
impl PinId for PA01 {}
// ...
impl PinId for PB31 {}
```

Define marker types for the pin modes in two layers. First you have the configuration layer.

```rust
pub trait InputConfig {}

pub enum Floating {}
pub enum PullDown {}
pub enum PullUp {}

impl InputConfig for Floating {}
impl InputConfig for PullDown {}
impl InputConfig for PullUp {}

// ...
```

Then you have the mode layer.

```rust
pub trait PinMode {}

pub struct Input<C: InputConfig> {}
impl<C: InputConfig> PinMode for Input<C> {}

// ...
```

We can make our lives slightly easier with some type aliases.

```rust
pub type FloatingInput = Input<Floating>;
pub type PullDownInput = Input<PullDown>;
// ...
```

Finally, you define a `Pin` as any combination of the two

```rust
pub struct Pin<I, M>
where
    I: PinId,
    M: PinMode,
{
    id: PhantomData<I>,
    mode: PhantomData<M>,
}
```

There are a few advantages to this approach. First, all pins now share a common type. Every pin is a `Pin`, and they differ only by their marker types. I believe most other HALs define a totally unique type for each pin, which is limiting. This approach is strictly more powerful, because it can easily express the other approach with aliases.

```rust
pub type Pa0<M> = Pin<PA00, M>;
pub type Pa1<M> = Pin<PA01, M>;
pub type Pb31<M> = Pin<PB31, M>;
```

However, this approach really shines with what I call the `Any*` trait pattern. I'll explain it in a few steps. First, let's define the trait as:

```rust
pub trait AnyPin {
    type Id: PinId;
    type Mode: PinMode;
}

impl<I, M> AnyPin for Pin<I, M>
where
    I: PinId,
    M: PinMode,
{
    type Id = I;
    type Mode = M;
}
```

What we've done here is store the type parameters for each `Pin` type as associated types in the `AnyPin` trait. We can define a type alias to recover the specific pin type from a type that implements `AnyPin`.

```rust
pub type SpecificPin<P> = Pin<<P as AnyPin>::Id, <P as AnyPin>::Mode>;
```

With this approach, you've abstracted all the different `Pin` types into a common trait, but you haven't lost any information. You can still use the `AnyPin` trait to constrain function inputs to specific types.

```rust
fn example_1(pin: impl AnyPin<Id = PA27>) {}

fn example_2(pin: impl AnyPin<Mode = PullDownInput>) {}
```

Or you can impose constraints on the associated types.

```rust
fn example_3<P>(pin: P)
where
    P: AnyPin,
    P::Id: SomeDriverTrait,
{
}
```

There one problem though. The compiler doesn't know that `AnyPin` is only implemented for `Pin` types. From its perspective, any type could potentially implement `AnyPin`. However, with some clever trait bounds on `AnyPin`, you can make it trivial to convert between some `P: AnyPin` and the corresponding, concrete `SpecificPin<P>`.

Let's modify the definition:

```rust
pub trait AnyPin
where
    Self: From<SpecificPin<Self>>,
    Self: Into<SpecificPin<Self>>,
    Self: AsRef<SpecificPin<Self>>,
    Self: AsMut<SpecificPin<Self>>,
    SpecificPin<Self>: AsRef<Self>,
    SpecificPin<Self>: AsMut<Self>,
{
    type Id: PinId;
    type Mode: PinMode;
}
```

We are requiring that any `P`, `&P` or `&mut P`, where `P: AnyPin` can be freely converted back an forth to `SpecificPin<P>`. Essentially, `AnyPin` now acts as a type constructor for `Pin`.

You can accept an `AnyPin` and then recover an instance of `Pin` like this.

```rust
fn example_4<P: AnyPin>(mut any_pin: P) {
    let pin_mut: &mut SpecificPin<P> = any_pin.as_mut();
    let pin_ref: &SpecificPin<P> = any_pin.as_ref();
    let pin: SpecificPin<P> = any_pin.into();
}
```

Note that I've provided the type annotations here, but they can almost always be inferred. The compiler doesn't know anything about `P` other than the fact that it can be converted into `SpecificPin<P>`, so it's almost always able to infer the type.

The `AnyPin` trait ends up being **really** powerful. It gives you all the ergonomics of a unified pin type, but it still lets you track the types at compile time and reject invalid configurations. I've actually started using this pattern **everywhere** in my drivers. It's *incredibly* useful.

Compared to your approach, `Pin` acts as the owner of a `PinId`, where each `PinId` gives ownership over the corresponding registers. The implementation in `atsamd-hal` is a bit too complicated for a short post here, but I can explain in more detail, if you want.

The difference is that `Pin` also owns the `PinMode`, which can also be tracked at compile time.

## Other cool features

- I've made it possible to convert between `PinMode`s using `From`/`Into`. All of the following do the same thing.

```rust
let pa27 = pins.pa27.into_floating_input();
let pa27 = pins.pa27.into_mode::<FloatingInput>();
let pa27: Pin<_, FloatingInput> = pins.pa27.into();
```

This makes it really convenient for users to switch modes.

- I provide a `DynPin`, type-erased pin type. I think this goes further than your `AnyPin` type, because it erases both the `PinId` type and the `PinMode` type. Everything is specified at run time.

You convert from `Pin` to `DynPin` using `From`/`Into`, and you convert back using `TryFrom`/`TryInto`, because the run-time values might not match the compile-time types.

## Comparison

I'll try to argue for the advantages of my approach relative to your proposal.

> - Simpler public API:
>    - No typestates. Input and Output act as "typestates-lite", but that's it.

I don't see this as an advantage. You lose compile-time guarantees with this approach. I think it's valuable to know from the type signature that a pin is, for example, in `PinMode` `AlternateM`.

> The user doesn't have to configure pins as input/output/opendrain/etc to use a peripheral.

With my approach, the peripheral driver can accept `P: AnyPin` and convert as necessary. It's ergonomic for both the user and the driver author.

> No need for a "Disconnected" typestate, having a "bare" AnyPin or PX_XX struct implies the Pin is disconnected. Input and Output disconnect the pin on drop.

I don't particularly like this. What if I want to disable the pin temporarily? Once I drop it, I can't get it back (safely). And what if I want to configure a pin but not carry it around forever. I would prefer dropping a pin have no effect.

> No code duplication between AnyPin and PX_XX. All the logic for Input and Output is defined once, no need for an uber-macro.

The great thing about my approach is that everything uses types and traits. In fact, my entire refactor began as an attempt to get rid of our uber-macro. See the section below for information about viewing the implementation.

> Drivers setting "weird" pin configurations that are not supported by the public API is no longer a hack. Drivers configuring GPIOs it is now the standard way of doing things.

I'm not sure I understand this. But in any case, users shouldn't have to configure anything. The driver should always be responsible.

> Drivers can be generic over AnyPin and PX_XX by taking P: Pin.

With my `AnyPin` trait, drivers can be generic over both `PinId` and `PinMode`.

> If user wants to use a pin as GPIO input, then wants to use it as a peripheral input, the peripheral driver will do one extra register write to set it as input even if not needed.

This is not a limitation with my approach.

## Further details

My implementation has been in place for a while now, and we are slowly converting the rest of our HAL to use it. I just recently made a PR that cleans up a lot of the logic around setting the registers. If you want to see the implementation, I would look there.

https://github.com/atsamd-rs/atsamd/pull/413

Also, here's a [link](https://atsamd-rs.github.io/docs/samd51n/thumbv7em-none-eabihf/doc/atsamd_hal/common/gpio/v2/pin/index.html) to the documentation, although it looks like it's slightly out of date. I'm not sure what happened.

bradleyharden:
One other thought. The OP approach can be expressed with my approach using type aliases. If you get rid of the `InputConfig` typestates and just use `Input` as a `PinMode`, you can do this.

```rust
pub type InputPin<I> = Pin<I, Input>;
pub type OutputPin<I> = Pin<I, Output>;
// ...
```

Just thought I would add that as an option.

PR #311, opened 2021-05-06
Enables PWM for the nRF9160

huntc:
Without it, we can't have blinky on the Thingy:91 as its RGB LEDs require PWM!

This PR assumes just unsecured peripheral access for now.

I've also included a new `pwm-blinky-demo` and ensure that this and the original `pwm-demo` are included with the `xtask` tests. Additionally, the `pwm-blinky-demo` is tested for the nRF9160 target.

Tested on an nRF9160-dk via the probe-rs/embed tool. The result from `cargo embed`:

image

cc @kalkyl


jonathanpallant:
Works on my nRF9160-DK. It might be worth a second PR just confirming which boards this examples are designed to run on.

jonathanpallant:
I also hacked it to blink at varying intensity levels to check the PWM worked correctly. We should note that on an nRF9160-DK maximum duty turns the LED off, and 0 turns the LED on (they must have an inverter in there, or they switch the ground not the Vcc).

jonas-schievink:
bors r=jonathanpallant

PR #317, opened 2021-05-14
Uart example

huntc:
Provides the simplest example of using UART for both the 52840 and 9160.

Tested successfully on:

* nrf52840-dk
* nrf9160-dk
* Thingy:91 (having modified the GPIO pins locally)

### DIFF
--- a/examples/hello-world/Cargo.toml
+++ b/examples/hello-world/Cargo.toml
@@ -1,0 +1,30 @@
+[package]
+name = "hello-world"
+version = "0.1.0"
+authors = ["Christopher Hunt"]
+edition = "2018"
+publish = false
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+cortex-m = "0.6.2"
+cortex-m-rt = "0.6.12"
+
+[dependencies.embedded-hal]
+version = "0.2.3"
+features = ["unproven"]
+
+[dependencies.nrf9160-hal]
+features = ["rt"]
+path = "../../nrf9160-hal"
+optional = true
+
+[dependencies.nrf52840-hal]
+features = ["rt"]
+path = "../../nrf52840-hal"
+optional = true
+
+[features]
+9160 = ["nrf9160-hal"]
+52840 = ["nrf52840-hal"]
--- a/examples/hello-world/Embed.toml
+++ b/examples/hello-world/Embed.toml
@@ -1,0 +1,63 @@
+[default.probe]
+# USB vendor ID
+# usb_vid = "1337"
+# USB product ID
+# usb_pid = "1337"
+# Serial number
+# serial = "12345678"
+# The protocol to be used for communicating with the target.
+protocol = "Swd"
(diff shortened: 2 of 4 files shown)


huntc (commented):
@jonathanpallant Can you elaborate further on the changes re https://github.com/nrf-rs/nrf-hal/pull/317/commits/1ba36303a1625802ac98999f781aab5aceb1c4f8 and why they are necessary? I was expecting to implement the nrfx workaround when initialising the UART for the 53 and 91 boards (see description), not a change for all boards. Thanks.

huntc:
> @jonathanpallant Can you elaborate further on the changes re [1ba3630](https://github.com/nrf-rs/nrf-hal/commit/1ba36303a1625802ac98999f781aab5aceb1c4f8) and why they are necessary? I was expecting to implement the nrfx workaround when initialising the UART for the 53 and 91 boards (see description), not a change for all boards. Thanks.

I've just seen https://github.com/nrf-rs/nrf-hal/pull/319. I shall revert your commit here @jonathanpallant and turn this PR into simply providing an example of using UART for nrf-hal. Thanks.

huntc:
@jonathanpallant Reverted your commit given the UART improvements/fixes elsewhere and made this just an example of how to set the UART up.

examples/hello-world/memory.x
```diff
@@ -0,0 +1,32 @@
...
+   sources added the attribute `#[link_section = ".ram2bss"]` to the data
+   you want to place there. */
+/* Note that the section will not be zero-initialized by the runtime! */
+/* SECTIONS {
+     .ram2bss (NOLOAD) : ALIGN(4) {
+       *(.ram2bss);
+       . = ALIGN(4);
+     } > RAM2
+   } INSERT AFTER .bss;
+*/
```
huntc, line 32: @jonathanpallant Unsure whether I still need the memory.x and build.rs - I think the pac now defaults to these settings?

jonathanpallant: The *HAL* includes a functional memory.x file, yes, so it's no longer required in the examples. The PAC only defines the interrupt handlers.

huntc:
No more changes expected here. Please review/merge accordingly.

thejpster (changes requested):
I think we should remove embed.toml, and add 9160 to the xtask list, but otherwise looks OK.

huntc:
> I think we should remove embed.toml

I provided the `embed.toml` to be consistent with the other examples. Could we look at the example deployment strategy as a separate issue to this PR?

> and add 9160 to the xtask list, but otherwise looks OK.

I will do this shortly. Thanks.

huntc:
The 9160 for the hello world example is now included in the xtask.

thejpster (approved):

thejpster:
> 
> 
> > I think we should remove embed.toml
> 
> I provided the `embed.toml` to be consistent with the other examples. Could we look at the example deployment strategy as a separate issue to this PR? All but one of the examples has an `embed.toml`.
> 

Oh! I totally didn't spot they'd been added to all the other examples. Yes, of course, we should be consistent.

PR #318, opened 2021-05-14
Enable PWM for nrf52832

Dirbaio:
It compiles cleanly, except for the PAC missing fields for the tasks registers. Replaced those with `.bits(1)`, which works on all chips.

Untested on hardware as I don't have a nrf52832 at hand.


jonathanpallant:
Tested OK on a DWM1001C.

jonathanpallant (approved):

jonas-schievink:
`@bors` is for the rust-lang instance, the one we use doesn't use `@`:

bors r=jonathanpallant

PR #319, opened 2021-05-14
Apply the UARTE workaround for nrf91 and nrf53

jonathanpallant:
See ttps://github.com/NordicSemiconductor/nrfx/blob/master/drivers/src/nrfx_uarte.c#L197

### DIFF
--- a/nrf-hal-common/src/uarte.rs
+++ b/nrf-hal-common/src/uarte.rs
@@ -120,7 +120,7 @@ where
         // NB Safety: This is taken from Nordic's driver -
         // https://github.com/NordicSemiconductor/nrfx/blob/master/drivers/src/nrfx_uarte.c#L197
         if unsafe { core::ptr::read_volatile(txenable_reg) } == 1 {
-            self.0.tasks_stoprx.write(|w| unsafe { w.bits(1) });
+            self.0.tasks_stoptx.write(|w| unsafe { w.bits(1) });
         }
 
         // NB Safety: This is taken from Nordic's driver -
--- a/nrf-hal-common/src/uarte.rs
+++ b/nrf-hal-common/src/uarte.rs
@@ -73,9 +73,6 @@ where
             }
         });
 
-        // Enable UARTE instance.
-        uarte.enable.write(|w| w.enable().enabled());
-
         // Configure.
         let hardware_flow_control = pins.rts.is_some() && pins.cts.is_some();
         uarte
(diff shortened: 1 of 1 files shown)


jonathanpallant:
This also removes the check for txstopped, which doesn't work on the nrf9160 (the event is always set).

nrf-hal-common/src/uarte.rs
```diff
@@ -43,6 +43,16 @@ where
     T: Instance,
 {
     pub fn new(uarte: T, mut pins: Pins, parity: Parity, baudrate: Baudrate) -> Self {
+        // Is the UART already on? It might be if you had a bootloader
+        if uarte.enable.read().bits() != 0 {
+            while uarte.events_txstopped.read().bits() == 0 {
```
Dirbaio, line 48: `events_txstopped` only fires if whoever was using it before did a `stoptx`. This might spin forever if that's not the case.

Dirbaio: This is exactly why the errata workaround checks for rxenable/txenable, but we can't rely on those being present in nrf52 since they're undocuemnted...

nrf-hal-common/src/uarte.rs
```diff
@@ -135,31 +201,26 @@ where
...
         // Wait for transmission to end.
-        let mut endtx;
-        let mut txstopped;
-        loop {
-            endtx = self.0.events_endtx.read().bits() != 0;
-            txstopped = self.0.events_txstopped.read().bits() != 0;
-            if endtx || txstopped {
-                break;
-            }
+        while self.0.events_endtx.read().bits() == 0 {
```
Dirbaio, line 209: the `events_txstopped.reset();` above is no longer needed.

nrf-hal-common/src/uarte.rs
```diff
@@ -135,31 +201,26 @@ where
...
+            // TODO: Do something here which uses less power. Like `wfi`.
         }
 
         // Conservative compiler fence to prevent optimizations that do not
         // take in to account actions by DMA. The fence has been placed here,
         // after all possible DMA actions have completed.
         compiler_fence(SeqCst);
 
-        if txstopped {
```
Dirbaio, line 153: Maybe remove the `Transmit` error? it'd be a breaking change though.

nrf-hal-common/src/uarte.rs
```diff
@@ -135,31 +201,26 @@ where
...
-            return Err(Error::Transmit);
-        }
-
         // Lower power consumption by disabling the transmitter once we're
         // finished.
         self.0.tasks_stoptx.write(|w|
             // `1` is a valid value to write to task registers.
             unsafe { w.bits(1) });
 
+        // Wait for transmitter to stop.
```
Dirbaio, line 224: The `tasks_stoptx` write and this spin is unneeded. If you don't abort a tx, just receiving endtx already means the tx machinery is fully stopped.

Dirbaio (changes requested):

nrf-hal-common/src/uarte.rs
jonathanpallant: Ah, I see. With no check, I splatter the last line emitted by SPM. What's a good thing to check for here? Or do we not care that we configure the UART whilst it may still be enabled?

nrf-hal-common/src/uarte.rs
jonathanpallant: From the enum? Might as well leave it in for now in case we do need it again.

nrf-hal-common/src/uarte.rs
jonathanpallant: According to Nordic:

```
            // Transmitter has to be stopped by triggering the STOPTX task to achieve
            // the lowest possible level of the UARTE power consumption.
```

jonathanpallant: Not that we seem to care about power consumption (as we busy-wait during a DMA transmit to the UART)

nrf-hal-common/src/uarte.rs
jonathanpallant: Noted. Will remove.

nrf-hal-common/src/uarte.rs
Dirbaio: there might be side effects if it's not disabled yeah :S 

Maybe always do a stoptx? I think if tx is stopped stoptx still causes a new txstopped, so the loop is guaranteed to hang? I'm not 100% sure though

nrf-hal-common/src/uarte.rs
Dirbaio: Oh huh, strange. When I experimented with this, I saw UARTE using power as long as it's enabled, no matter what the tx/rx state is, so to have low power you have to disable it anyway... Embassy only waits for txstopped when disabling it, and only if the tx was aborted, which seems to work fine.

Seeing nordic has this code I'm now OK with leaving this in, though I don't fully get what the purpose is.

nrf-hal-common/src/uarte.rs
jonathanpallant: OK, will do.

nrf-hal-common/src/uarte.rs
```diff
@@ -109,7 +181,6 @@ where
 
         // Reset the events.
         self.0.events_endtx.reset();
-        self.0.events_txstopped.reset();
```
Dirbaio, line 112: Oops. Actually this was needed if we keep the blocking wait for txstopped below :sweat_smile: ... It should probably be moved next to that so it's more obvious why it's needed.

jonathanpallant: 🤣

Dirbaio (approved):
Looks great

bors r+

PR #321, opened 2021-05-14
Update 9160 HAL with latest memory map.

jonathanpallant:
Also adds a README to explain why we need SPM.

### DIFF
--- a/nrf9160-hal/README.md
+++ b/nrf9160-hal/README.md
@@ -1,7 +1,7 @@
 # Hardware Abstration Layer for the Nordic nRF9160
 
 This crate is a Hardware Abstraction Layer (HAL) for the Nordic nRF9160. It
-wraps the PAC (`nrf9160-pac`) provides high level wrappers for the chip's
+wraps the PAC (`nrf9160-pac`) and provides high level wrappers for the chip's
 peripherals.
 
 This crate knows nothing about your PCB layout, or which pins you have assigned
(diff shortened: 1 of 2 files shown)


nrf9160-hal/memory.x
```diff
@@ -1,28 +1,43 @@
...
-  FLASH : ORIGIN = 0x00040000, LENGTH = 768K
-  RAM : ORIGIN = 0x20020000, LENGTH = 128K
+	/*
+	 * This is where the Bootloader, Secure Partition Manager or
+	 * Trusted-Firmware-M lives.
+	 */
+	SECURE_FLASH : ORIGIN = 0x00000000, LENGTH = 256K
+	/*
+	 * This is where your non-secure Rust application lives. Note that SPM must agree this
+    * is where your application lives, or it will jump to garbage and crash the CPU.
```
jonas-schievink, line 14: This line does not use a hard tab
```suggestion
	 * is where your application lives, or it will jump to garbage and crash the CPU.
```

That said, is this copied from Nordic, or should be rather use 4-space indentation like everywhere else here?

jonathanpallant: Fixed.

nrf9160-hal/README.md
```diff
@@ -0,0 +1,78 @@
+# Hardware Abstration Layer for the Nordic nRF9160
+
+This crate is a Hardware Abstraction Layer (HAL) for the Nordic nRF9160. It
+wraps the PAC (`nrf9160-pac`) provides high level wrappers for the chip's
```
jonas-schievink, line 4:
```suggestion
wraps the PAC (`nrf9160-pac`) and provides high level wrappers for the chip's
```

jonas-schievink (approved):
One minor grammar nit, otherwise this seems good

huntc (commented):
Nice!

PR #324, opened 2021-05-28
Enhances the RTC example with an interrupt

huntc:
I find it common to use an interrupt with RTC so I've enhanced this example to illustrate how an interrupt can be established and managed.


thejpster (approved):
Looks good to me

huntc:
Thanks @thejpster !

PR #330, opened 2021-06-13
Fix SPIS Transfer is_done() mistake

kalkyl:
Noticed there was a mistake in the `Transfer` impl of `is_done()` where it would `take()` `inner` instead of `as_mut()`.


jonas-schievink:
bors d+

PR #337, opened 2021-07-13
Provides common Nvmc functionality

### LINKED ISSUE #336: A common HAL NVMC abstraction - needed?
Do we need a common-hal abstraction for the NVMC as per https://github.com/NordicSemiconductor/nrfx/blob/7eca6c2dc02b24cbdaa3ba0e63a7195e34ebe07c/hal/nrf_nvmc.h#L259

Writing code to interact with the NVMC on the nRF9160 is quite error prone e.g. to write out a 32 bit slice given a static mut of CONFIG positioned in flash:

```rust
// Erase anything that's there
nvmc.configns.write(|w| w.wen().een());
unsafe {
    CONFIG[0] = 0xffffffff;
}
nvmc.configns.reset();

// Write the slice
nvmc.configns.write(|w| w.wen().wen());
unsafe {
    let mut k = 0;
    for i in (0..sliced_len).step_by(4) {
        let mut word = ...
        while !nvmc.ready.read().ready().bit_is_set() {}
        CONFIG[k] = word;
        cortex_m::asm::dmb();
        k += 1;
    }
}
nvmc.configns.reset();
```

huntc:
The [embedded-storage](https://github.com/rust-embedded-community/embedded-storage) traits are implemented to provide a higher-level abstraction for reading and writing from/to flash storage. In the first instance, support is provided for the nRF52 boards and the nRF9160, but other boards should be relatively easy to support if required in the future.

Sample usage of writing to flash given some 32 bit aligned slice of bytes:

```rust
// Erase a page
let _ = nvmc.try_erase(0, 4096);

// Write the 32 bit aligned slice
let _ = nvmc.try_write(0, some_u8_slice_on_32_bit_bounds);
```

An example of establishing the `Nvmc` on the nRF9160:

```rust
let mut nvmc = Nvmc::new(board.NVMC_NS, unsafe { &mut CONFIG });
```

...where CONFIG points to flash:

```rust
extern "C" {
    #[link_name = "_config"]
    static mut CONFIG: [u32; 1024];
}
```

...which can be expressed in a memory layout:

```
MEMORY
{
  /* NOTE 1 K = 1 KiBi = 1024 bytes */
  FLASH : ORIGIN = 0x00040000, LENGTH = 764K
  CONFIG : ORIGIN = ORIGIN(FLASH) + LENGTH(FLASH), LENGTH = 4K /* 4K is the flash page size */
  RAM : ORIGIN = 0x20020000, LENGTH = 128K
}

_config = ORIGIN(CONFIG);
```

Testing on the nRF52840 device can be achieved via `cargo test --test nvmc` from within the `nrf52840-hal-tests` folder.

Fixes #336 

TODO:

* [x] Implement the `ReadNorFlash` methods
* [x] Implement the `NorFlash` methods
* [x] Provide an `Instance` for the nRF52840
* [x] Provide an `Instance` for the nRF9160
* [x] Provide a nvmc module in nrf52840-tests for the new abstraction
* [x] Write some more tests to test some boundary conditions
* [x] Go through the code finely and compare it again with the nrfx library
* [x] Provide an example of reading, writing and erasing from/to storage

### DIFF
--- a/nrf-hal-common/src/nvmc.rs
+++ b/nrf-hal-common/src/nvmc.rs
@@ -1,0 +1,202 @@
+//! HAL interface to the Non-Volatile Memory Controller (NVMC) peripheral.
+
+use core::ops::Deref;
+
+#[cfg(not(feature = "9160"))]
+use crate::pac::nvmc;
+#[cfg(feature = "9160")]
+use crate::pac::nvmc_ns as nvmc;
+#[cfg(not(feature = "9160"))]
+use crate::pac::NVMC;
+#[cfg(feature = "9160")]
+use crate::pac::NVMC_NS as NVMC;
+
+use embedded_storage::nor_flash::{NorFlash, ReadNorFlash};
+
+/// Interface to an NVMC instance.
+pub struct Nvmc<T: Instance> {
+    nvmc: T,
+    storage: &'static mut [u32],
+}
+
+impl<T> Nvmc<T>
+where
+    T: Instance,
+{
+    /// Takes ownership of the peripheral and storage area.
+    pub fn new(nvmc: T, storage: &'static mut [u32]) -> Nvmc<T> {
+        Self { nvmc, storage }
+    }
+
+    /// Consumes `self` and returns back the raw peripheral and associated storage.
+    pub fn free(self) -> (T, &'static mut [u32]) {
+        (self.nvmc, self.storage)
+    }
+
+    fn enable_erase(&self) {
+        #[cfg(not(feature = "9160"))]
+        self.nvmc.config.write(|w| w.wen().een());
+        #[cfg(feature = "9160")]
+        self.nvmc.configns.write(|w| w.wen().een());
+    }
+
+    fn enable_read(&self) {
+        #[cfg(not(feature = "9160"))]
+        self.nvmc.config.write(|w| w.wen().ren());
+        #[cfg(feature = "9160")]
+        self.nvmc.configns.write(|w| w.wen().ren());
+    }
+
+    fn enable_write(&self) {
+        #[cfg(not(feature = "9160"))]
+        self.nvmc.config.write(|w| w.wen().wen());
+        #[cfg(feature = "9160")]
+        self.nvmc.configns.write(|w| w.wen().wen());
+    }
+
+    #[inline]
+    fn wait_ready(&self) {
+        while !self.nvmc.ready.read().ready().bit_is_set() {}
+    }
+
+    #[cfg(feature = "9160")]
+    #[inline]
+    fn wait_write_ready(&self) {
+        while !self.nvmc.readynext.read().readynext().bit_is_set() {}
+    }
+
+    #[cfg(not(feature = "9160"))]
+    #[inline]
+    fn erase_page(&mut self, offset: usize) {
+        let bits = &mut (self.storage[offset as usize >> 2]) as *mut _ as u32;
+        self.nvmc.erasepage().write(|w| unsafe { w.bits(bits) });
+        self.wait_ready();
+    }
+
+    #[cfg(feature = "9160")]
+    #[inline]
+    fn erase_page(&mut self, offset: usize) {
+        self.storage[offset as usize >> 2] = 0xffffffff;
+        self.wait_ready();
+    }
+
+    #[inline]
+    fn write_word(&mut self, offset: usize, word: u32) {
+        #[cfg(not(feature = "9160"))]
+        self.wait_ready();
+        #[cfg(feature = "9160")]
+        self.wait_write_ready();
+        self.storage[offset] = word;
+        cortex_m::asm::dmb();
+    }
+}
+
+impl<T> ReadNorFlash for Nvmc<T>
+where
+    T: Instance,
+{
+    type Error = NvmcError;
+
+    const READ_SIZE: usize = 4;
+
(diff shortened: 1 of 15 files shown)


huntc:
https://github.com/titanclass/thingy91-lorawan-nbiot/pull/2/commits/4a9f6730e329758f4bc7a78be3f221466cbcfb81 shows the positive impact of this abstraction.

nrf-hal-common/src/nvmc.rs
```diff
@@ -0,0 +1,101 @@
...
+    fn try_erase(&mut self, _from: u32, _to: u32) -> Result<(), Self::Error> {
+        unimplemented!()
+    }
+
+    fn try_write(&mut self, _offset: u32, _bytes: &[u8]) -> Result<(), Self::Error> {
+        unimplemented!()
+    }
+}
+
+pub trait Instance {
```
jonas-schievink, line 64: This trait should be sealed like our other `Instance` traits

nrf-hal-common/src/nvmc.rs
```diff
@@ -0,0 +1,101 @@
...
+#[cfg(any(feature = "52840"))]
+use crate::pac::NVMC;
+#[cfg(any(feature = "9160"))]
+use crate::pac::NVMC_NS;
+use embedded_storage::nor_flash::{NorFlash, ReadNorFlash};
+
+/// Interface to an NVMC instance.
+pub struct Nvmc<'a, T: Instance, const N: usize> {
+    nvmc: T,
+    storage: &'a mut [u32; N],
```
jonas-schievink, line 12: I'm not sure how you're planning on using this, but if you plan on offering non-blocking APIs we need to make sure this is `mem::forget`-safe (presumably by making it `'static`)

Also some rationale on why this is const-generic instead of a dynamic slice would be nice, and why this is using `u32` instead of `u8` as the element type – does the NVMC have alignment requirements?

nrf-hal-common/src/nvmc.rs
```diff
@@ -0,0 +1,101 @@
...
+        unimplemented!()
+    }
+
+    fn try_write(&mut self, _offset: u32, _bytes: &[u8]) -> Result<(), Self::Error> {
+        unimplemented!()
+    }
+}
+
+pub trait Instance {
+    fn enable_erase();
```
Dirbaio, line 65: I'm not sure if these methods are needed. I'd just do the `#[cfg]`s in the logic above, which imo is much simpler.

nrf-hal-common/src/nvmc.rs
Dirbaio: Yes, writes must be aligned to u32. I agree on the const generics though, a dynamic slice seems simpler.

Also, I'm not sure if there's UB issues with having a `&mut` to the data while you modify it / erase it through magic register writes...?

nrf-hal-common/src/nvmc.rs
huntc: Commit c71362c35521009a442f727845b6272d8c2a5e63 attends to this.

nrf-hal-common/src/nvmc.rs
huntc: Commit ecca3729154d8bd89791e8149d685b9a0d45865b attends to this.

nrf-hal-common/src/nvmc.rs
huntc: > I'm not sure how you're planning on using this, but if you plan on offering non-blocking APIs we need to make sure this is `mem::forget`-safe (presumably by making it `'static`)

Commit 70aad9449bb83c0ad2fa3eb19f17044c270a6f9b makes its 'static.

huntc: > Also some rationale on why this is const-generic instead of a dynamic slice would be nice.

Commit 70aad9449bb83c0ad2fa3eb19f17044c270a6f9b changes this to a dynamic slice. Dunno what I was thinking. :-)

huntc: > Also, I'm not sure if there's UB issues with having a `&mut` to the data while you modify it / erase it through magic register writes...?

I don't fully understand the question. However, the data isn't actually modified via registers... the NVMC is put into various modes and then regular memory operations ensue.

nrf-hal-common/src/nvmc.rs
huntc: Commit 7958ca2c5be51555b09662955344c06740e33bd9 reverts back to the trait methods because there is no common type representing the NVMC configuration. For the 52840, we have the CONFIG type which is a register of a _CONFIG struct. For the 9160 it is CONFIGNS and _CONFIGNS respectively.

huntc: Actually, commit e0e44824a62d4bca846db6f2d1b25ff9d98f3764 figures it out. `Deref` FTW.

huntc: Commit 90b8818617177c76470ed46a4226e4f6461eac7f also adds 52840 considerations.

huntc:
@jonas-schievink @Dirbaio We now have tests for the NVMC via nRF52840-hal-tests, and things appear to be working! I want to go through things finely before converting from a draft PR, and write an example. I also want to do some boundary checks via my tests. Your feedback is appreciated!

huntc:
This PR is complete and ready for a thorough review.

Issue #339, opened 2021-07-18
Can't use with onewire due to InputPin + OutputPin requirement


nevi-me:
When trying to use `nrf-hal-common` with OneWire (https://crates.io/crates/onewire), InputPin needs to be implemented for OutputPin.

I tried to implement this myself, but I seem to be a bit out of my depth :(

Example code is:

```rust
let peripherals = Peripherals::take().unwrap();

let pin1 = nrf52840_hal::gpio::p1::Parts::new(peripherals.P1);
let mut temp_pin = pin1.p1_08.degrade().into_open_drain_output(
    OpenDrainConfig::Disconnect0HighDrive1, Level::High
);
let mut wire = OneWire::new(&mut temp_pin, false);
```

This results in the error:

```rust
error[E0271]: type mismatch resolving `<nrf52840_hal::gpio::Pin<nrf52840_hal::gpio::Output<OpenDrain>> as nrf52840_hal::prelude::OutputPin>::Error == ()`
  --> src/main.rs:30:29
   |
30 | let mut wire = OneWire::new(&mut temp_pin, false);
   |                             ^^^^^^^^^^^^^ expected `()`, found enum `void::Void`
   |
   = note: required for the cast to the object type `dyn OpenDrainOutput<(), Error = (), Error = ()>`

error[E0277]: the trait bound `nrf52840_hal::gpio::Pin<nrf52840_hal::gpio::Output<OpenDrain>>: _embedded_hal_digital_InputPin` is not satisfied
  --> src/main.rs:30:29
   |
30 | let mut wire = OneWire::new(&mut temp_pin, false);
   |                             ^^^^^^^^^^^^^ the trait `_embedded_hal_digital_InputPin` is not implemented for `nrf52840_hal::gpio::Pin<nrf52840_hal::gpio::Output<OpenDrain>>`
   |
   = note: required because of the requirements on the impl of `nrf52840_hal::prelude::InputPin` for `nrf52840_hal::gpio::Pin<nrf52840_hal::gpio::Output<OpenDrain>>`
```

jonas-schievink:
I've opened https://github.com/nrf-rs/nrf-hal/pull/340 which should resolve this

unrelentingtech:
Previously: https://github.com/nrf-rs/nrf51-hal/pull/25

Finomnis:
I've opened #401 which should finally resolve this.

PR #340, opened 2021-07-20
Implement `InputPin` regardless of mode

jonas-schievink:
This should resolve https://github.com/nrf-rs/nrf-hal/issues/339

I *think* this is fine, since you can always read any pin regardless of configuration (even for disconnected pins, their value will just be floating).

### DIFF
--- a/nrf-hal-common/src/gpio.rs
+++ b/nrf-hal-common/src/gpio.rs
@@ -259,7 +259,7 @@ impl<MODE> Pin<MODE> {
     }
 }
 
-impl<MODE> InputPin for Pin<Input<MODE>> {
+impl<MODE> InputPin for Pin<MODE> {
     type Error = Void;
 
     fn is_high(&self) -> Result<bool, Self::Error> {
@@ -529,7 +529,7 @@ macro_rules! gpio {
                     }
                 }
 
-                impl<MODE> InputPin for $PXi<Input<MODE>> {
+                impl<MODE> InputPin for $PXi<MODE> {
                     type Error = Void;
 
                     fn is_high(&self) -> Result<bool, Self::Error> {


jonas-schievink:
Ah, it seems like this doesn't work without also setting `cnf.input().connected()` for all output modes. Apparently we already do this for push-pull outputs for some reason:

https://github.com/nrf-rs/nrf-hal/blob/19df1633e60bfd5d882a350a3fffd47592b517c6/nrf-hal-common/src/gpio.rs#L205

Not sure why though? And what are the consequences of doing this anyways? Higher power usage?

Dirbaio:
> Not sure why though? And what are the consequences of doing this anyways? Higher power usage?

gpios have an "input buffer" you can connect/disconnect. If you don't connect it, you read garbage. You can enable output and input at the same time. For PushPull outputs you just read what you write. For OpenDrain you can drive the line low or you can listen whether the *other* device drives it down.

Enabling input when not needed wastes power. The Product Specification doesn't say how much though :( . Probably very little.

I think always implementing InputPin is a bit of a footgun though. Too easy to configure a PushPull output and accidentally pass it to a driver that requires InputPin. IMO it should be something the user explicitly opts in. Embassy has a `FlexPin` for this, which allows using input+output at the same time.

oddstr13:
Is IoPin https://github.com/rust-embedded/embedded-hal/pull/269 relevant?

https://github.com/rust-embedded/embedded-hal/releases/tag/v1.0.0-alpha.5

eflukx:
> Is IoPin [rust-embedded/embedded-hal#269](https://github.com/rust-embedded/embedded-hal/pull/269) relevant?

Seems reasonable if that's the path the embedded-wg goes, nrf-hal follows along?

Currently I'm in need of this kind of pin handling (in my case tri-stating GPIO pins, not using them as inputs per se), for handling charlieplexed LEDs and a device with a bit too much leakage on its IO-pins (i.e. floating the pins lowers idle/quiescent current draw) .

Finomnis:
Here is my attempt to implement this: https://github.com/nrf-rs/nrf-hal/pull/340

Finomnis:
@jonas-schievink  Can probably be closed; was superseded by #401

PR #343, opened 2021-07-29
 Add embedded_hal::serial implementation for uarte (continuation)

hargoniX:
Since #281 has been abandoned by the author this is a continuation of his work in a seperate PR.

### DIFF
--- a/nrf-hal-common/src/uarte.rs
+++ b/nrf-hal-common/src/uarte.rs
@@ -709,6 +709,8 @@ where
             }
             Ok(self.rx_buf[0])
         } else {
+            // We can only read 1 byte at a time, otherwise ENDTX might not be raised,
+            // causing the read to stall forever.
             start_read(&uarte, &mut self.rx_buf[..1])?;
             Err(nb::Error::WouldBlock)
         }
--- a/nrf-hal-common/src/uarte.rs
+++ b/nrf-hal-common/src/uarte.rs
@@ -202,7 +202,7 @@ where
     ///
     /// The buffer must have a length of at most 255 bytes.
     pub fn read(&mut self, rx_buffer: &mut [u8]) -> Result<(), Error> {
-        start_read(&*self.0, rx_buffer, rx_buffer.len())?;
+        start_read(&*self.0, rx_buffer)?;
 
         // Wait for transmission to end.
         while self.0.events_endrx.read().bits() == 0 {}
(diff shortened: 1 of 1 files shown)


nrf-hal-common/src/uarte.rs
```diff
@@ -406,18 +443,252 @@ pub enum Error {
...
+{
+    type Error = Error;
+
+    /// Write a single byte to the internal buffer. Returns nb::Error::WouldBlock if buffer is full.
+    fn write(&mut self, b: u8) -> nb::Result<(), Self::Error> {
+        let uarte = unsafe { &*T::ptr() };
+
+        // Prevent writing to buffer while DMA transfer is in progress.
+        if uarte.events_txstarted.read().bits() == 1 {
+            return Err(nb::Error::WouldBlock);
```
jonas-schievink, line 589: Where does `events_txstarted` get reset if we hit this path?

jonas-schievink (commented):
How was this code tested? This seems potentially tricky to get right, so I'm a bit hesitant to merge this without proper testing and review (I'd appreciate others to review this for correctness as well)

hargoniX:
I tested the UARTE with this little program on the micro:bit:
```rs
#![no_main]
#![no_std]

use panic_halt as _;

use core::fmt::Write;

use microbit::{
    hal::uarte,
    hal::uarte::{Baudrate, Parity},
    hal::prelude::*,
};

static mut TX_BUF: [u8; 1] = [0; 1];
static mut RX_BUF: [u8; 1] = [0; 1];

use cortex_m_rt::entry;


#[entry]
fn main() -> ! {
    let board = microbit::Board::take().unwrap();

    let (mut tx, mut rx) = {
        let serial = uarte::Uarte::new(
            board.UARTE0,
            board.uart.into(),
            Parity::EXCLUDED,
            Baudrate::BAUD115200,
        );
        serial.split(
            unsafe { &mut TX_BUF },
            unsafe { &mut RX_BUF }
        ).unwrap()
    };

    loop {
        write!(tx, "Hello World:\r\n").unwrap();
        let input = nb::block!(rx.read()).unwrap();
        write!(tx, "You wrote: {}\r\n", input as char).unwrap();
    }
}
```
As well as smaller and bigger variants of the TX_BUF size and it all seems to work. You were indeed right about the write not being functional though so I added some more logic that should make sure the flags are always properly reset and we only return WouldBlock if the TX transaction has both been started *and* is not finished yet.

jonas-schievink:
Yeah, I'm not entirely sure what the intended behavior from `embedded-hal` is there, but it seems fine to just write to a buffer for now. If this is wrong, the `embedded-hal` docs should be clarified to forbid this.

This PR still needs a rebase, but other than that I think it's ready to land. Thanks!

jonas-schievink:
bors retry

PR #344, opened 2021-08-02
common/spi: add support for all nRF52 devices

lucab:
This extends the `spi` implementation to all nRF52 devices.
Some features-based conditionals are required as those devices
have some slightly different register names and a varying amount
of supported interfaces.
Manually tested locally on a nRF52840 only.


jonas-schievink (approved):
bors r+

Issue #358, opened 2021-09-23
Examples/Demos have become unmaintainable


jonas-schievink:
We currently have 24 independent demo projects in the `examples` directory, each with its own `Cargo.toml` that needs to be maintained, some being of questionable didactic value or demonstrating functionality that is already incorporated in another demo. Many haven't been updated ever since they were added, except for fixing breakage. This is not maintainable.

I'm declaring a moratorium on adding more examples until we fix this.

I propose that we:
- Convert some (or even most) demos to doc examples, especially ones that are specific to a single peripheral
- Merge all of the remaining demos into the same Cargo package (`demos`) and make them actual Cargo examples

Other solutions welcome.

therealprof:
> Convert some (or even most) demos to doc examples, especially ones that are specific to a single peripheral

I haven't really seen any convincing embedded doc examples yet. They also tend to bitrot very quickly because even if they can be compiled and do compile, people tend to put important details into the commentary because you can't really code it out and that doesn't see any testing or updates.

Also the important part is being able to transfer doc examples back into real projects and is often a big stumbling block, even if the doc example is actually good.

> Merge all of the remaining demos into the same Cargo package (demos) and make them actual Cargo examples 

My two cents: In `stm32f4xx-hal` we're thinking about going the opposite route and moving Cargo examples into their one or more Cargo packages. If you have a single set of examples, not only the HAL impl itself but also all other dependencies (and their versions) are shared so you cannot simply bump one or add a new example without upgrading all others. The same goes for features so if you'd like to make aspects of the examples configurable you'll have to do the feature disambiguation yourself and also the `Cargo.toml` does get a bit unwieldy.

Maybe there's some middle ground?

PR #363, opened 2021-11-27
add rtic serial interrupt example

elwerene:
I'm sorry I didn't get the gdb runnter to run, so I could only test a version with probe-rs (sth similiar like here: https://github.com/elwerene/nrf52833-rtic-serial-interrupt).

It took me some hours to get a working version with lots of help from people hanging around at the nRF Rust Matrix channel.

I also only could test the code on nrf52833. Should I remove the other features, or will it work on them?

### DIFF
--- a/examples/rtic-serial-interrupt/.cargo/config
+++ b/examples/rtic-serial-interrupt/.cargo/config
@@ -1,0 +1,2 @@
+[target.thumbv7em-none-eabihf]
+runner = "arm-none-eabi-gdb -tui"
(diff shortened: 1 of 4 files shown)


jonas-schievink:
Note that we're not currently merging any more examples due to https://github.com/nrf-rs/nrf-hal/issues/358

elwerene:
Oops.. well that's reasonable! Maybe it's enough to have it as an open pull request so people can find it :)

elwerene:
Well, I'm feeling stupid now as that's exactly where I could have found a working example in https://github.com/nrf-rs/nrf-hal/pull/102 :/

elwerene:
Could I somehow work this into the documentation? I really don't want anyone else to have to follow my path to UART interrupts.

PR #364, opened 2021-12-09
Add From impl for pins

ZoeyR:
I ran into this when needing to take a generic struct that could contain any gpio pin. Since all the gpio pins have a degrade function I see no reason to not implement `From`. This allows generic code to work better.

### DIFF
--- a/nrf-hal-common/src/gpio.rs
+++ b/nrf-hal-common/src/gpio.rs
@@ -541,6 +541,12 @@ macro_rules! gpio {
                     }
                 }
 
+                impl<MODE> From<$PXi<MODE>> for Pin<MODE> {
+                    fn from(value: $PXi<MODE>) -> Self {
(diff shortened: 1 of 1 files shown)


ZoeyR:
I think the build error is unrelated to my change since that's a dependency resolution issue and my code just adds a `From` impl.

jonas-schievink:
Thanks! The CI issue should have been fixed by #365.

bors r+

ZoeyR:
Awesome! Would you be able to cut a release after this? I can always pin to git but I've been trying to cut down on the amount of git dependencies in my `Cargo.toml` files.

jonas-schievink:
Sadly releases are pretty involved, so I'd like to get some more changes in before I publish one

ZoeyR:
Understandable! I can pin to git for now. Thanks for merging so fast!

PR #366, opened 2021-12-12
Add nrf5340-app-hal

sphw:
This PR adds a HAL for the nRF5340's app core. It is more or less identical to the nRF9160 HAL that already exists, with some features added or removed for the nRF5340. I have not added support for the net core yet, as I don't need it for my use case.

### DIFF
--- a/Cargo.ci.toml
+++ b/Cargo.ci.toml
@@ -7,6 +7,7 @@ members = [
   "nrf52833-hal",
   "nrf52840-hal",
   "nrf52840-hal-tests",
+  "nrf5340-app-hal",
   "nrf9160-hal",
   "examples/*",
 ]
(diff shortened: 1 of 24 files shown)


nrf5340-app-hal/log.log
```diff
@@ -0,0 +1,3527 @@
+[2021-Dec-03 05:29:48] [ info] --------------------------------------------------------------------------------
```
jonas-schievink, line 1: Looks like this was checked in by accident

sphw: Oh yup. Just removed it

jonas-schievink:
Please add the new HAL to the array in `xtask/src/lib.rs` to enable CI for it

sphw:
> Please add the new HAL to the array in `xtask/src/lib.rs` to enable CI for it

Done

jonas-schievink:
Thanks, looks pretty good!

bors r+

PR #372, opened 2022-01-07
Unified how pins are return in `free` calls

korken89:
Hi,

I unified so pins are returned the same for all peripherals.

### DIFF
--- a/nrf-hal-common/src/pwm.rs
+++ b/nrf-hal-common/src/pwm.rs
@@ -761,6 +761,10 @@ where
         let ch1 = self.pwm.psel.out[1].read();
         let ch2 = self.pwm.psel.out[2].read();
         let ch3 = self.pwm.psel.out[3].read();
+        self.pwm.psel.out[0].reset();
+        self.pwm.psel.out[1].reset();
+        self.pwm.psel.out[2].reset();
+        self.pwm.psel.out[3].reset();
         (
             self.pwm,
             if ch0.connect().bit_is_set() {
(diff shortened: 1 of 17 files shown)


nrf-hal-common/src/pwm.rs
```diff
@@ -748,8 +748,46 @@ where
...
-    pub fn free(self) -> T {
-        self.pwm
+    pub fn free(
+        self,
+    ) -> (
+        T,
+        Option<Pin<Output<PushPull>>>,
+        Option<Pin<Output<PushPull>>>,
+        Option<Pin<Output<PushPull>>>,
+        Option<Pin<Output<PushPull>>>,
```
jonas-schievink, line 758: It probably makes sense to create a `Pin` struct in every module and return that, unless the module only deals with a single pin. Otherwise these signatures get a bit unwieldy.

jonas-schievink (commented):
I wonder if it would be beneficial to add a `free` method that only returns the peripheral, for efficiency (the added volatile reads add to code size). Presumably users would almost always want the pins back though?

nrf-hal-common/src/pwm.rs
korken89: Makes sense, I'll go through and add `Pins` structs where none existed yet.

korken89:
> I wonder if it would be beneficial to add a `free` method that only returns the peripheral, for efficiency (the added volatile reads add to code size). Presumably users would almost always want the pins back though?

If there should be a free without the pins, to me it does not make sense to not get back all the hardware that was used.
The efficiency loss here are 4 register reads worst case, so the overhead does not seem like an issue to me. Would you agree?

korken89:
There, it should be ready for a re-review now @jonas-schievink

jonas-schievink (approved):
Looks good, thanks!

PR #373, opened 2022-01-08
Implement MultiwriteNorFlash for nRF52 boards that support it

ia0:
This PR implements `MultiwriteNorFlash` as suggested [here](https://github.com/nrf-rs/nrf-hal/issues/336#issuecomment-878684844).

I only listed nRF52 boards that support writing twice to the same word, because I'm not familiar with the other boards or crate features.


huntc (approved):
Thanks! Note I don’t have permission to merge.

PR #383, opened 2022-02-18
implement MultiWriteNorFlash for the nrf9160

folkertdev:
per the product specification [section 4.4](https://infocenter.nordicsemi.com/pdf/nRF9160_PS_v2.0.pdf#page=29) there can be n_write writes, where `t_write = 2` [source](https://infocenter.nordicsemi.com/pdf/nRF9160_PS_v2.0.pdf#page=35). Unless I'm missing something, that means that the contract of `MultiWriteNorFlash` is satisfied.

### DIFF
--- a/nrf-hal-common/src/nvmc.rs
+++ b/nrf-hal-common/src/nvmc.rs
@@ -190,6 +190,7 @@ where
     feature = "52811",
     feature = "52833",
     feature = "52840",
+    feature = "9160",
 ))]
 impl<T: Instance> embedded_storage::nor_flash::MultiwriteNorFlash for Nvmc<T> {}
 


folkertdev:
Using this in an application gives 

```
warning: Linking globals named 'DEVICE_PERIPHERALS': symbol multiply defined!

error: failed to load bc of "nrf9160_pac-a62a87be05e92ef9.nrf9160_pac.20f4e1dd-cgu.4.rcgu.o": 

warning: `connected-pallet-firmware` (bin "connected-pallet-firmware") generated 1 warning
The following warnings were emitted during compilation:

warning: head: 75d48ef, tag-version: b1ad42b, dirty: true, production: false
```

Not sure what this means, but I guess another feature flag needs to be added/removed somewhere?

jonas-schievink:
Indeed, this looks correct. Thanks!

bors r+

PR #385, opened 2022-03-29
Add support for partial flash erase

arturkow2000:
Partial erase is a feature that allows to reduce time CPU is halted when
using NVMC. It works by dividing page erase operation into multiple
operations, in-between each operation CPU is resumed allowing interrupt
handling.

Erasing a single page takes 85 ms on nRF52840, during write CPU can't handle interrupts,
this may result, e.g. in broken USB, especially if erase happens during enumeration.

This code has been tested on nRF52840 by zeroing page, erasing it and checking whether all bits are back to 1.


krystian-hebel (approved):

krystian-hebel (changes requested):
Tested on nRF52840, but must be made to work for other SoCs.

qwandor:
This seems like a good idea! Can you rebase it on top of the latest master? Is there any difference between different chips that we need to handle before merging this?

PR #387, opened 2022-05-19
Fixed nvms erase problem for nrf91 & nrf53

diondokter:
A byte offset was given where a word offset was required (factor 4 difference)


jamesmunns (approved):
bors r+

PR #389, opened 2022-06-01
Add free method to Saadc

hdoordt:
Add a method to disable the Saadc and regain the inner pac-level peripheral handle


jonas-schievink:
Thanks!

bors r+

jonas-schievink:
Thanks!

bors r+

PR #390, opened 2022-06-13
added method to Timer trait for resetting event flag

balderk:
Added method to reset timer event flag.


jonas-schievink (changes requested):
Please format the method like the others, with the body in its own line

PR #391, opened 2022-06-13
added impl statements for SPIM and UARTE interfaces on nrf9160

balderk:
Added impl statements for SPIM and UARTE interfaces for the nrf9160


jonas-schievink:
Thanks!

bors r+

PR #394, opened 2022-07-22
Documentation for Timer Instance

caemor:
Add documentation for pub timer instances

### DIFF
--- a/nrf-hal-common/src/timer.rs
+++ b/nrf-hal-common/src/timer.rs
@@ -353,8 +353,10 @@ pub trait Instance: sealed::Sealed {
     /// This interrupt associated with this RTC instance.
     const INTERRUPT: Interrupt;
 
+    /// Return the correct register block
     fn as_timer0(&self) -> &RegBlock0;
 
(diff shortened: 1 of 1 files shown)


nrf-hal-common/src/timer.rs
```diff
@@ -389,46 +391,56 @@ pub trait Instance: sealed::Sealed {
...
         self.as_timer0().tasks_stop.write(|w| unsafe { w.bits(1) });
         self.timer_reset_event();
     }
 
+    /// Checks if the timer is still running which means no event is yet generated for CC[0]
     fn timer_running(&self) -> bool {
         self.as_timer0().events_compare[0].read().bits() == 0
     }
 
+    ///
```
qwandor, line 410: Either add some documentation here or remove this line.

qwandor (approved):

nrf-hal-common/src/timer.rs
qwandor: Done myself.

PR #396, opened 2022-08-20
Improve examples

koalatux:
I'm not sure if the examples should be explicitly included or excluded from the workspace, but with the originally provided Cargo.example.toml the examples can't be built.

I disabled rtt in Embed.toml for the usb examples, because they don't use this crate.

The rest is just some minor cleanup.


jonas-schievink:
Thanks!

bors r+

PR #397, opened 2022-09-15
Use `is_connected` to determine whether a ctx/rtx was connected to uarte

hdoordt:
`Uarte::free` method uses `bit_is_set` to determine whether a pin was connected to `cts` and `rts`. However, a set bit represents a disconnected pin. Use `is_connected()` instead to do the inversion.


jonas-schievink:
Thanks!

bors r+

Issue #60, opened 2019-02-21
Add basic documentation to README


hannobraun:
Documentation is sorely lacking right now. It would be great to have a more useful README.

futurepaul:
I'd like to help with this. I have an NRF52840 dev kit that I have built C code for but I don't understand this crate well enough to get the blinky example to work.

If someone can walk me through the setup / build process over discord or something I'd be happy to record that conversation and write it up as documentation! I'm working on a mac but I also have linux and windows available if that's helpful.

Yatekii:
I can walk you though on IRC, no problem ;) I am there day and night. Just ping me! Mac is great even tho it has nothing to do with this crate ;)

futurepaul:
Alright thanks to Yatekii I have the blinky example working! (I wasn't converting the binary into a hex before flashing). I'll be happy to write up that process if there's a good place to put it. I'm not familiar enough with the rust embedded world to write docs beyond that, unfortunately.

hannobraun:
@futurepaul I think it's fine to dump anything that helped you get things working into the README. Not ideal, but anything's better than no documentation, so don't be afraid to propose sub-optimal docs :-)

Another good place for general documentation would be a doc comment in the root module of the respective crate (`lib.rs`).

Istar-Eldritch:
I have a nRF52-DK board (nRF52832 Microcontroller). This is how I got it working on debian, it may be used as a starting point.

# Dependencies

- arm-none-eabi-objcopy: Required to transform the `ELF` file to `HEX`
- nrfjprog: Required to install the `HEX` in the DK board.

# Dependencies for debugging
- openocb compiled with jlink support.
- gdb-multiarch

# Debugging

Start openocd 
```bash
  openocd -f interface/jlink.cfg -c "transport select swd" -f target/nrf52.cfg
```

Execute the intended example
```bash
  cargo run --example <example_name>
```

# Installing

Build the ELF file
```bash
  cargo build
```

Transform the ELF file into HEX
```bash
arm-none-eabi-objcopy \
  -O ihex \
  target/thumbv7em-none-eabihf/debug/watch \
  target/thumbv7em-none-eabihf/debug/watch.hex
```
Flash the micro
```bash
  nrfjprog --family nRF52 --eraseall && \ 
  nrfjprog --family nRF52 --program target/thumbv7em-none-eabihf/debug/watch.hex && \
  nrfjprog --family nRF52 -r
```

# Other

`.cargo/config`

[...]

PR #398, opened 2022-09-18
README: Mention USB copying of Hex files

chrysn:
Inspired by https://github.com/nrf-rs/nrf-hal/issues/60 (but not resolving it fully, probably), this adds text to the general examples README on alternatives to the `cargo embed` process described in the individual READMEs.

This might need a few more items on the list based on the methods listed in #60; I'll be asking there to state which methods are still relevant. (To me, only copying is relevant, as that's what works without going through the process of unlocking the AP lock that is engaged on the nrf42dk board).


jonas-schievink:
Thanks, this seems like a good addition!

bors r+

PR #400, opened 2022-09-20
Make SPI clock pin optional for use with WS2812

aelray:
smart-leds-rs uses SPI without clock for controlling WS2812 LEDs.


jonas-schievink:
Thanks!

bors r+

PR #401, opened 2022-09-20
Add OpenDrainIO pin state (with InputPin capability)

Finomnis:
As already reported in https://github.com/nrf-rs/nrf-hal/issues/339, there is currently no way to create an I/O pin in nrf-hal.

There are several reasons why having this would be useful. My personal usecase is the DHT22/AM2302 sensor. Controlling it requires a single-wire pulled-up open drain IO.

Sadly, this means that all libraries that can interface with it [require `InputPin + OutputPin`](https://docs.rs/dht-sensor/0.2.1/dht_sensor/trait.InputOutputPin.html).

# Solution
There is no inherent reason why nrf chips shouldn't be able to implement this. The input buffer is always available and allows reading back the pin values during every pin configuration. Although for energy saving reasons, the input buffer is currently disabled during the `OpenDrain` state.

My proposal is to add an `OpenDrainIO` state that does not disable the input buffer and implements `InputPin`.

### DIFF
--- a/nrf-hal-common/src/gpio.rs
+++ b/nrf-hal-common/src/gpio.rs
@@ -25,11 +25,6 @@ pub struct PushPull;
 /// Open drain output (type state).
 pub struct OpenDrain;
 
-/// Input/Output combined mode (type state).
-pub struct InOut<MODE> {
-    _mode: PhantomData<MODE>,
-}
-
 /// Open drain input/output (type state).
 pub struct OpenDrainIO;
 
@@ -302,7 +297,7 @@ impl<MODE> Pin<MODE> {
         self,
         config: OpenDrainConfig,
         initial_output: Level,
-    ) -> Pin<InOut<OpenDrainIO>> {
+    ) -> Pin<Output<OpenDrainIO>> {
         let mut pin = Pin {
             _mode: PhantomData,
             pin_port: self.pin_port,
@@ -353,7 +348,7 @@ impl<MODE> InputPin for Pin<Input<MODE>> {
     }
 }
 
-impl<MODE> InputPin for Pin<InOut<MODE>> {
+impl InputPin for Pin<Output<OpenDrainIO>> {
     type Error = Void;
 
     fn is_high(&self) -> Result<bool, Self::Error> {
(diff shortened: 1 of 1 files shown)


Finomnis:
Open Questions: How should `IoPin` behave? Should we switch back and forth between `OpenDrain` and `OpenDrainIO`? Or should we just stay in `OpenDrainIO`? What's the usecase where a library would use `IoPin`? Would that usecase benefit from switching the input buffer on and off, or would that rather be a performance hit?

In my case, the DHT22 sensor libraries require `InputPin + OutputPin`. When would they use `IoPin` instead? What's the benefit of `IoPin` over `InputPin + OutputPin`?

Finomnis:
Turns out other people [have doubts as well](https://github.com/rust-embedded/embedded-hal/issues/340). Therefore [IoPin was removed](https://github.com/rust-embedded/embedded-hal/issues/357) from `embedded-hal` for now. I'll remove it from this PR as well.

Finomnis:
Possible solution if we want to re-introduce `IoPin` at some point:
- Implement `IoPin<Self, Self>` for `Pin<Output<OpenDrainIO>>`
- Switch to passive output state when transitioning to `Read`
   - would require tracking if we are in `0Normal1Disabled` or `0Disabled1Normal`.

Finomnis:
Will write tests for this as well, as soon as I have an nrf52840-DK.

jonas-schievink:
I wonder if this should just be the behavior of `OpenDrain` pins in general? I agree with https://github.com/nrf-rs/nrf-hal/pull/340#issuecomment-917206892 that it's a footgun for `PushPull` pins, but for `OpenDrain`, not so much, since those are legitimate input pins.

Finomnis:
> I wonder if this should just be the behavior of `OpenDrain` pins in general?

To my knowledge this was an intentional decision, because having the input buffer connected to a floating pin can cause substantially increased power drain.

jonas-schievink:
Okay, that sounds good!

bors r+

Finomnis:
@jonas-schievink Added tests in #402

PR #402, opened 2022-09-28
Add missing OpenDrain[IO] tests

Finomnis:
- Add tests for `OpenDrainIO`
    - Added in #401
- Add missing test for `OpenDrain`


jonas-schievink:
Thanks! Note that we don't run these tests during CI, so updating them for HAL changes isn't necessary.

bors r+

Finomnis:
> Thanks! Note that we don't run these tests during CI, so updating them for HAL changes isn't necessary.
> 
> bors r+

It's good practice, though ;)

PR #412, opened 2022-11-14
Add is_done and try_wait methods to i2s::Transfer

nospam2678:
Wouldn't it be useful if being able to determine whether an I²S transfer has completed without busy looping?

The changes is this PR appears to be working for both `wait()` and `try_wait()`, but I must admit that the compiler fencing part flies a bit above my head even after refreshing myself on the topic. I'm actually preplexed of why it is needed in the first place. Is it to prevent buffer from being fetched for return prior to the loop being finished?

An implementation detail which might be questioned is whether `is_done()` is at all desired. My reasoning is that it reduces the code duplication slightly, and that adding it to the public api might be useful for someone. One could argue an extra function call adds overhead, but I am imagining the compiler would inline and optimize away such a tiny function.

### DIFF
--- a/nrf-hal-common/src/i2s.rs
+++ b/nrf-hal-common/src/i2s.rs
@@ -680,15 +680,40 @@ struct Inner<B> {
 }
 
 impl<B> Transfer<B> {
+    /// Returns `true` if the transfer is done.
+    pub fn is_done(&self) -> bool {
+        if let Some(inner) = self
+            .inner
+            .as_ref()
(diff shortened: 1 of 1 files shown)


jonas-schievink:
Thanks, this looks good. I'm a bit uneasy about the `unreachable_unchecked` calls, I wish Rust allowed moving out of fields of a `Drop` type, but I guess there's little we can do.

bors r+

PR #413, opened 2022-11-17
Correct eb3b96b with a useful function signature for try_wait

nospam2678:
This is embarrassing. It appears my quality assurance for #412, while existent, was way flawed and insufficient.

Having a `try_` method take ownership over self makes it impossible call it more than once. Clearly a mutable reference was what was indented and required to actually make the method meaningful.

It would be a good idea to either merge this PR, or revert the previous one, prior to making the next release.

### DIFF
--- a/nrf-hal-common/src/i2s.rs
+++ b/nrf-hal-common/src/i2s.rs
@@ -693,20 +693,6 @@ impl<B> Transfer<B> {
         }
     }
 
-    /// Attempts to return the buffer if the transfer is done.
-    pub fn try_wait(mut self) -> Option<(B, I2S)> {
-        if self.is_done() {
-            let inner = self
-                .inner
-                .take()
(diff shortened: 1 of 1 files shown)


jonas-schievink:
This change would be unsound (due to the `unreachable_unchecked`; but the only alternative is a panic or a signature change in the other methods, which is also not great) if the user started another transfer while the old one is still around.

I think we might want to remove `try_wait` in favor of having the user check `is_done` - if `is_done` returns `true`, the user knows that the next call to `wait` won't block.

nospam2678:
Indeed. Makes sense! Branch has been updated to remove try_wait and only leave is_done.

Maybe the corresponding is_done method should be added to also to TransferFullDuplex? I lack a hardware setup to do any I²S in the other direction, so that would be completely written blindly if I did it.

PR #421, opened 2023-06-21
Use more common syntax to pass slices to twim writes and reads in twim demo

hdoordt:


BartMassey:
I think you need to declare those arrays `mut` to be able to borrow them `& mut`?

qwandor (approved):

Issue #384, opened 2022-02-18
RTIC monotonic implementation


barafael:
There is already a working implementation of RTIC monotonic timer for TIMER{0,1,2} here: https://github.com/kalkyl/nrf-play/blob/47f4410d4e39374c18ff58dc17c25159085fb526/src/mono.rs
It would be nice for this to be part of the HAL so that the implementation does not need to be copied every time :) I'm trying to do this, but I'm not sure how.

ivario123:
I realize that this might be stale but, It would be relatively simple to add this. I made a draft [implementation](https://github.com/ivario123/nrf-hal-monotonic/blob/master/nrf-hal-common/src/monotonic.rs) making the abstraction agnostic to whether you are using the RTC or the TIMER. Using the RTC yields much better power efficiency.

Using this would look something like this:
```rust
use nrf5284_hal as hal;
use hal::monotonic::*;
use hal::rtc::RTC0;
use hal::timer::TIMER0;

#[rtic::app(device = nrf52840_hal::pac,dispatchers=[TIMER2,TIMER3])]
mod app {

    use super::*;
    #[monotonic(binds = RTC0, default = true)]
    type MyMono = MonotonicTimer<RTC0>;
    // Or this if using timer
    #[monotonic(binds = TIMER0, default = true)]
    type MyMono = MonotonicTimer<TIMER0>;
    

    #[shared]
    struct Shared {}

    #[local]
    struct Local {}

    #[init]
    fn init(cx: init::Context) -> (Shared, Local, init::Monotonics) {
        let mono = MonotonicTimer::new(cx.device.RTC0);
        // Or this if using timer
        let mono = MonotonicTimer::new(cx.device.TIMER0);


        // Return shared, local and mono
        (
            Shared {},
            Local {},
            init::Monotonics(mono),
        )
   }
}
```

Making it trivial to change clocks when writing your app

ivario123:
@barafael I think that I have finished an [implementation](https://github.com/ivario123/nrf-hal-monotonic/blob/master/nrf-hal-common/src/monotonic.rs) for this 

The API looks like this at the time of writing and it will not change much. 

```rust
#![no_main]
#![no_std]
use core::panic::PanicInfo;
use nrf52840_hal as hal;
use hal::monotonic::MonotonicTimer;
use hal::pac;
#[panic_handler]
fn panic(_panic: &PanicInfo<'_>) -> ! {
    loop {}
}
[...]

PR #427, opened 2023-10-03
RTIC Monotonic for RTC and TIMER

Kyrne:
This PR aims to add rtic monotonic implementations for the RTC and TIMER.
As #384 requested. This would allow users of the hal to use native implementations
of the Monotonic trait rather than third-party ones and therefore ensure future support for the
feature.

## Added

- monotonic.rs ( holds all of the rtic monotonic related implementations )
- 2 new optional dependencies, [fugit](https://docs.rs/fugit/latest/fugit/) and [rtic-monotonic](https://docs.rs/rtic-monotonic/latest/rtic_monotonic/)
- 1 new feature flag for all hals `--features monotonic`
- 1 new example crate ( monotonic-blinky )

## Features

- Configurable frequency ( can use all valid frequencies for the rtc and timer respectively )
- Compile time or initiation time guarantees that the frequency is valid.

## Usage

There exists an example of how to instantiate the monotonic rtc and timer in `nrf-hal/examples/monotonic-blinky`. 

### RTC

The RTC cannot at compile time guarantee that the frequency is correct.
It does however check this at initiation time, so for any invalid 
frequency it will throw an error leaving the user to correct the frequency.
By taking a reference to the clock object we can ensure that the
`lfclk` is enabled.
```rust
// RTC0 with a frequency of 32 768 Hz
type MyMono = MonotonicRtc<RTC0, 32_768>;

let clocks = hal::clocks::Clocks::new(cx.device.CLOCK);
let clocks = clocks.start_lfclk();

let mono = MyMono::new(cx.device.RTC0, &clocks).unwrap();
```

### Timer

The timer abstraction supports frequency gating since there are only 9 different prescaler values.
This ensures that the `MonotonicTimer` object is not constructible for any frequency that does
not yield a valid prescaler. The documentation for the `MonotonicTimer` lists all
valid frequencies in a table with their corresponding time until overflow.

```rust
// RTC0 with a frequency of 16 MHz
type MyMono = MonotonicTimer<TIMER0, 16_000_000>;
let mono = MyMono::new(cx.device.TIMER0).unwrap();
```

## Recommended actions
Squash and merge to remove the long commit history.

## Alternatives

There are already a few third-party implementations of these features that are less configurable
 - https://gist.github.com/korken89/fe94a475726414dd1bce031c76adc3dd
 - https://github.com/kalkyl/nrf-play/blob/47f4410d4e39374c18ff58dc17c25159085fb526/src/mono.rs
 - https://github.com/rtic-rs/rtic/tree/master/rtic-monotonics/src/nrf


ivario123:
@diondokter Not quite sure how the review process works but figured I'd mention you here since we have discussed earlier PRs

diondokter:
Hey, this looks pretty nice! However...

I'm not a maintainer here (though I think I might have some limited permissions). In fact, this crate has no active maintainers at this point...

So I don't know how to continue. Best thing would be to find a new maintainer.
Lately in the embedded rust matrix chat, the recommendation has been to go with embassy-nrf which you can use blockingly as well.

ivario123:
@qwandor Would you mind looking through this?

qwandor (approved):

qwandor:
Thanks! I've updated this to work with current master and fixed a bunch of small things. It looks good, except that it doesn't build for nrf51 or nrf52832. Is that expected?

ivario123:
> Thanks! I've updated this to work with the current master and fixed a bunch of small things. It looks good, except that it doesn't build for nrf51 or nrf52832. Is that expected?

This should not be the case, at least not for the nrf52832 chip as the docs specify the same field, I think that it is simply a matter of the field names either not existing in the pac or being different from that of the nrf52840 that we tested on. I will look into it

BartMassey:
Thanks for working on this!

I'm easily confused. Should this trait be named `rtic-monotonic`? Or is it more general than that?

ivario123:
@qwandor  it seems that we might need to modify the bits manually or re-write parts of the PAC. As this is a small feature I think that it is better to simply unsafely modify the bits and not re-write the PAC for this. 

I implemented a short fix that should be tested before including as it adds three extra bits of unchecked unsafe code.
https://github.com/Kyrne/nrf-hal/tree/fix_missing_parts_of_pac

ivario123:
> Thanks for working on this!
> 
> I'm easily confused. Should this trait be named `rtic-monotonic`? Or is it more general than that?

This is probably a good idea.

qwandor:
Thanks! We should probably fix the PAC to add the missing fields, but I'm alright with this for now.

qwandor:
> Thanks for working on this!
> 
> I'm easily confused. Should this trait be named `rtic-monotonic`? Or is it more general than that?

Do you mean the feature flag?

ivario123:
> Thanks! We should probably fix the PAC to add the missing fields, but I'm alright with this for now.

If at all possible it would be great to test it on these devices though, I only have access to 52840 devkit so I cannot test it on any other devices.

ivario123:
Excuse the force push, I resolved the conflicts with a PR instead. I also renamed the user-facing feature to rtic-monotonic as I find it describes the feature better. The feature in nrf-hal-common is still simply `monotonic` to not conflict with the name of the dependency.

BartMassey:
@qwandor Yeah, meant the feature flag.

I think we should probably have someone test any `unsafe` and get some good eyes on it before we merge that. I'm bad at Github: is there a specific commit/diff that I'm missing that does that part?

qwandor:
Unfortunately I don't have any nRF51 or nRF52832 board either. I've tested it on an nRF52833 and that works at least, and the code looks reasonable to me.

PR #431, opened 2024-01-18
Implement embedded-hal 1.0 traits

qwandor:
Now that `embedded-hal` 1.0 is released, this PR adds implementations of most of its relevant traits. For now I've added them alongside the existing `embedded-hal` 0.2 trait implementations, so that users can still use drivers based on the old traits. We can remove the old traits in a later release.

I haven't implemented the `embedded-hal-nb` and `embedded-hal-async` traits yet, and the I2C traits are also missing as the design has changed and implementing it correctly is non-trivial.

### DIFF
--- a/nrf-hal-common/src/delay.rs
+++ b/nrf-hal-common/src/delay.rs
@@ -1,8 +1,9 @@
 //! Delays.
-use cortex_m::peripheral::syst::SystClkSource;
-use cortex_m::peripheral::SYST;
 
 use crate::clocks::HFCLK_FREQ;
+use core::convert::TryInto;
+use cortex_m::peripheral::syst::SystClkSource;
+use cortex_m::peripheral::SYST;
 use embedded_hal::delay::DelayNs;
 use embedded_hal_02::blocking::delay::{DelayMs, DelayUs};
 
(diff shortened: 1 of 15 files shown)


qwandor:
Hi, could I get a review of this soon? It's blocking me from updating some other things to embedded-hal 1.0.

diondokter:
@qwandor Hi, that depends a bit on: https://github.com/nrf-rs/nrf-hal/issues/432

In essence, this crate has no active maintainers...

nrf-hal-common/src/delay.rs
```diff
@@ -27,28 +27,46 @@ impl Delay {
...
+    }
+}
+
+impl DelayNs for Delay {
+    fn delay_ns(&mut self, ns: u32) {
         // The SysTick Reload Value register supports values between 1 and 0x00FFFFFF.
         const MAX_RVR: u32 = 0x00FF_FFFF;
 
-        let mut total_rvr = us * (HFCLK_FREQ / 1_000_000);
+        let mut total_rvr = ns * (HFCLK_FREQ / 1_000_000_000);
```
mattheww, line 69: `HFCLK_FREQ` is `64_000_000`, so isn't this always going to come out as zero?

qwandor: Whoops, good point. Fixed.

mattheww: That fix looks correct to me, and I can verify that the SYST-based `DelayNs` implementation works given the fix, when it didn't before.

But I'm not a maintainer of this repository, and I haven't read through this whole PR.

BartMassey:
I'll try these out with some of my stuff over the weekend and let you know if I see anything. I have only reviewed a couple of these so far — will get to the rest shortly.

PR #434, opened 2024-03-05
Prepare for 0.16.1 release.

qwandor:


BartMassey (approved):
Thanks! This is a lot, and greatly appreciated.

I glanced at all the individual merged PRs, and everything looked fine. If you want me to do more QA, I (or one of my students) can pull this and try it on a bunch of my repos, but it will take some time. I see nothing that could reasonably be problematic here, and would suggest just going ahead.

Excited to get the 0.16.1 release out!

PR #435, opened 2024-03-06
Make embedded-hal 0.2 optional

qwandor:
This required some refactoring as some of the embedded-hal 1.0 implementations depended on the embedded-hal 0.2 implementations.

### DIFF
--- a/nrf-hal-common/src/gpio.rs
+++ b/nrf-hal-common/src/gpio.rs
@@ -86,7 +86,6 @@ use crate::pac::P1;
 use crate::pac::P1_NS as P1;
 
 use embedded_hal::digital::{ErrorType, InputPin, OutputPin, StatefulOutputPin};
-use void::Void;
 
 impl<MODE> Pin<MODE> {
     fn new(port: Port, pin: u8) -> Self {
@@ -232,9 +231,11 @@ impl<MODE> Pin<MODE> {
     }
 
     /// Convert the pin to be a push-pull output with specified drive.
-    pub fn into_push_pull_output_drive(self, initial_output: Level, drive: DriveConfig) ->
-        Pin<Output<PushPull>>
-    {
+    pub fn into_push_pull_output_drive(
+        self,
+        initial_output: Level,
+        drive: DriveConfig,
+    ) -> Pin<Output<PushPull>> {
         let mut pin = Pin {
             _mode: PhantomData,
             pin_port: self.pin_port,
@@ -404,8 +405,9 @@ impl<MODE> StatefulOutputPin for Pin<Output<MODE>> {
     }
 }
 
+#[cfg(feature = "embedded-hal-02")]
 impl<MODE> embedded_hal_02::digital::v2::InputPin for Pin<Input<MODE>> {
-    type Error = Void;
+    type Error = void::Void;
 
     fn is_high(&self) -> Result<bool, Self::Error> {
         self.is_low().map(|v| !v)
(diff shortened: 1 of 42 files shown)


nrf52811-hal/Cargo.toml
```diff
@@ -26,10 +26,7 @@ default-features = false
...
 
-[dependencies.embedded-hal]
-features = ["unproven"]
-version = "0.2.3"
-
 [features]
+embedded-hal-02 = ["nrf-hal-common/embedded-hal-02"]
 rt = ["nrf52811-pac/rt"]
-default = ["rt"]
+default = ["rt", "embedded-hal-02"]
```
ia0, line 32: I guess `embedded-hal-02` is added as a default feature to avoid breaking change. But what about users that disable default features? I'm not sure what the [Cargo Book](https://doc.rust-lang.org/cargo/reference/semver.html) recommends in those cases.

xtask/tests/ci.rs
```diff
@@ -45,6 +46,25 @@ fn main() {
...
+        // Without default features.
+        let mut cargo = Command::new("cargo");
+        let status = cargo
+            .args(&[
+                "build",
+                "--manifest-path",
+                &toml_path,
+                "--target",
+                target,
+                //"--no-default-features",
```
ia0, line 58: Is this a leftover?

ia0 (approved):
This is not a thorough review (in particular given I'm not familiar with the code) but ideally could help move forward.

nrf52811-hal/Cargo.toml
qwandor: There will be breaking changes in this release anyway, so I'm not really worried about that. I've included it by default just to make life easier for new users who might be using device drivers that depend on the embedded-hal 0.2 traits. People who care about binary size and build time can disable default features.

xtask/tests/ci.rs
qwandor: Whoops, well spotted. Uncommented.

nrf-hal-common/src/gpio.rs
```diff
@@ -232,9 +231,11 @@ impl<MODE> Pin<MODE> {
...
 
     /// Convert the pin to be a push-pull output with specified drive.
-    pub fn into_push_pull_output_drive(self, initial_output: Level, drive: DriveConfig) ->
-        Pin<Output<PushPull>>
-    {
+    pub fn into_push_pull_output_drive(
+        self,
+        initial_output: Level,
+        drive: DriveConfig,
+    ) -> Pin<Output<PushPull>> {
```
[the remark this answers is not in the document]
qwandor: Good idea, added.

nrf-hal-common/src/saadc.rs
```diff
@@ -99,6 +106,58 @@ impl Saadc {
...
+            5 => self.0.ch[0].pselp.write(|w| w.pselp().analog_input5()),
+            6 => self.0.ch[0].pselp.write(|w| w.pselp().analog_input6()),
+            7 => self.0.ch[0].pselp.write(|w| w.pselp().analog_input7()),
+            #[cfg(not(feature = "9160"))]
+            8 => self.0.ch[0].pselp.write(|w| w.pselp().vdd()),
+            #[cfg(any(feature = "52833", feature = "52840"))]
+            13 => self.0.ch[0].pselp.write(|w| w.pselp().vddhdiv5()),
+            // This can never happen the only analog pins have already been defined
+            // PAY CLOSE ATTENTION TO ANY CHANGES TO THIS IMPL OR THE `channel_mappings!` MACRO
+            _ => unsafe { unreachable_unchecked() },
```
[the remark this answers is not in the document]
qwandor: Good point. This isn't new code, but returning an error seems like a better option than causing undefined behaviour.

PR #436, opened 2024-03-06
Use embedded-hal 1.0 in tests and examples

qwandor:


examples/blinky-button-demo/src/main.rs
CodingInGreen, line 1: Why was v2 removed?

examples/i2s-peripheral-demo/src/main.rs
CodingInGreen, line 1: Can you comment on the change here?

The change from:

use embedded_hal::blocking::spi::Write;

to:

use embedded_hal::spi::SpiBus;

examples/pwm-blinky-demo/src/main.rs
CodingInGreen, line 1: Can you please comment on the change of the use of timer?

nrf52840-hal-tests/tests/gpio-input-pulled.rs
CodingInGreen, line 1: Can you please comment on the change to make pulldown_pin mutuable?

nrf52840-hal-tests/tests/gpio-output-open-drain-io.rs
CodingInGreen, line 1: Can you please comment on the change made from:

(state.input_pin.as_ref().unwrap().is_low().unwrap());

to:

(state.input_pin.as_mut().unwrap().is_low().unwrap());

examples/i2s-peripheral-demo/src/main.rs
qwandor: The SPI traits changed significantly in `embedded-hal` 1.0. You can read more about it in the `embedded-hal` documentation, but one difference is that there is now a single trait for a bus or a device, rather than separate traits for read vs. write vs. transfer.

examples/pwm-blinky-demo/src/main.rs
qwandor: What about it? `embedded-hal` 1.0 no longer includes traits for timers, so we have intrinsic methods instead. The names of some of the PWM methods have changed.

nrf52840-hal-tests/tests/gpio-input-pulled.rs
qwandor: `embedded_hal::digital::InputPin::is_low` takes `&mut self`, so it needs to be mutable for the call on line 49.

nrf52840-hal-tests/tests/gpio-output-open-drain-io.rs
qwandor: Again, `embedded_hal::digital::InputPin::is_low` takes `&mut self`, so we need a mutable reference.

CodingInGreen (approved):
Responses to comments are reasonable as are code changes. Tested on micro:bit V2, ran into some limitations with certain examples being built for other nrf hals for other chips. Overall, looks great! Approved!

PR #440, opened 2024-04-12
Implement embedded-hal 1.0 I2c trait for Twi and Twim

qwandor:
The `Twi` implementation is untested as I don't have any nrf51 devices.

Unfortunately the `Twim` implementation isn't working correctly. Adjacent operations of the same type are [specified](https://docs.rs/embedded-hal/latest/embedded_hal/i2c/trait.I2c.html#tymethod.transaction) to be carried out without an SR and SAD+R/W, which is why I avoid setting the `lastrx_stop` or `lasttx_stop` shortcut in that case, but when testing this I've found that trying to have two read operations in a row results in an overrun error for some reason.


nrf-hal-common/src/twim.rs
```diff
@@ -392,6 +397,185 @@ where
             },
         )
     }
+
+    fn write_part(&mut self, buffer: &[u8], final_operation: bool) -> Result<(), Error> {
+        unsafe { self.set_tx_buffer(buffer)? };
```
9names, line 402: No fence required here?

9names (approved):
Don't have time for a couple of days to test this out again, but logic looks good.

nrf-hal-common/src/twi.rs
```diff
@@ -250,6 +250,92 @@ where
...
+        self.0
+            .shorts
+            .write(|w| w.bb_stop().disabled().bb_suspend().disabled());
+
+        // Set Slave I2C address.
+        self.0
+            .address
+            .write(|w| unsafe { w.address().bits(address.into()) });
+
+        // `Some(true)` if the last operation was a read, `Some(false)` if it was a write.
```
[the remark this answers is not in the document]
qwandor: Yeah, I was just being lazy. Changed to an enum.

nrf-hal-common/src/twim.rs
qwandor: I'm not entirely sure. Added some more fences to be on the safe side.

qwandor:
> Don't have time for a couple of days to test this out again, but logic looks good.

I tested the `Twim` implementation on a Microbit v2 with various combinations of writes and reads (including consecutive reads), and the `Twi` implementation hasn't changed since you tested it so I think this should be fine.

PR #422, opened 2023-06-22
Support nRF52805

placrosse:

PR #477, opened 2024-05-23
Fix build after #422.

qwandor:
Merging #422 broke the build because it was out of date. This should fix it.


BartMassey (approved):

BartMassey:
Great to see support for the new chips!

PR #482, opened 2024-07-15
nrf9120 support

nerwalt:
Adding support for the nRF9120.  This is the internal part number for the nRF9161 and the nRF9151. 
Tested on an nRF9151dk.


nrf9120-hal/flash_tfm.sh
```diff
@@ -0,0 +1,2 @@
+nrfjprog --family NRF91 --recover
+nrfjprog --family NRF91 --chiperase --verify --program tfm.hex
```
qwandor, line 2: Where does `tfm.hex` come from?

nerwalt: Trusted Firmware M.  It replaces (includes) the Secure Partition Manager in newer versions of the Nordic SDK.  It is required to run in non-secure mode on the nrf91s.  It can be built with the SDK (e.g. the hello world TFM example).  It is not needed for nrf9120 support, and I didn't mean to include it in the PR.  I will remove it.  I was trying to do too many things at once yesterday when I submitted this.  Apologies for any confusion.

nrf9120-hal/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "nrf9120-hal"
+version = "0.18.0"
+description = "HAL for nRF9120 system-in-package"
+readme = "README.md"
+
+repository = "https://github.com/nrf-rs/nrf-hal"
+authors = [
+    "Jonathan Pallant (42 Technology) <jonathan.pallant@42technology.com>",
```
qwandor, line 9: Did you copy this from somewhere else? I think you can put your own details here.

nerwalt: Yes.  I started the 9160.  I need to update the README too.  Jumped the gun with this PR...

PR #489, opened 2024-10-28
removed unnecessary unsafe in pwm

BartMassey:
Uses of `addr_of_mut!()` were unnecessarily wrapped in `unsafe`. The resulting warnings stopped CI. The unnecessary `unsafe`s are now removed.

### DIFF
--- a/nrf-hal-common/src/pwm.rs
+++ b/nrf-hal-common/src/pwm.rs
@@ -1208,7 +1208,7 @@ impl Instance for crate::pac::PWM0 {
     const INTERRUPT: Interrupt = Interrupt::PWM0;
     #[inline(always)]
     fn buffer() -> *mut [u16; 4] {
-        unsafe { addr_of_mut!(BUF0) }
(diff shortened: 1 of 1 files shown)


qwandor (approved):
I guess this is because of https://blog.rust-lang.org/2024/10/17/Rust-1.82.0.html#safely-addressing-unsafe-statics, so this increases our MSRV to 1.82? I'm alright with that, but just making sure that that's what you intend.

BartMassey:
Oh, good point! I guess I can stick "allow" on these instead if we want to preserve the MSRV while still "--deny warnings". (Is "--deny warnings" in CI a good idea? I have very mixed feelings.)

I want to get the CI build working so we can bump the version of `panic-halt` to 1.0 to match `microbit` and have the checks pass.

qwandor:
I'm fine with either putting `allow` on them or bumping the MSRV, up to you.

BartMassey:
Closed in favor of #491

PR #520, opened 2025-05-05
Fix nRF51 HFCLK freq

### LINKED ISSUE #519: HFCLK_FREQ is wrong for nRF51
The code in clocks.rs does this

```
/// High Frequency Clock Frequency (in Hz).
pub const HFCLK_FREQ: u32 = 64_000_000;
```

This should be set to 16Mhz for nRF51.

kaidokert:
Fixes #519


BartMassey (approved):
Looks good to me. Have checked against NRF51822 manual. Assume other NRF51 are the same.

BartMassey:
Gemini confirms all nRF51 devices are 16MHz, whatever that is worth.

qwandor (approved):

PR #523, opened 2025-06-20
cleaned up PWM rustdoc

BartMassey:
The PWM naming and usage is pretty confusing, and the Rustdoc does not clarify it. This is a bunch of doc changes to try to triage the situation. Should file an issue to maybe change the API, although at this late date not so likely probably.


BartMassey:
Issue #524 addresses future work.

nrf-hal-common/src/pwm.rs
```diff
@@ -92,7 +93,16 @@ where
...
+    /// Sets the PWM output period from the given input
+    /// frequency. (This calculates the period by dividing
+    /// the current PWM clock rate by `freq`.)
+    ///
+    /// This calculation takes the current [Prescaler]
+    /// setting into account.
+    ///
+    /// Subsequent use of `set_duty_on_common()`,
+    /// `set_duty_on_group()`, *etc* will override this duty
+    /// cycle setting.
```
qwandor, line 105: What duty cycle setting? Isn't this setting the frequency rather than the duty cycle?

nrf-hal-common/src/pwm.rs
```diff
@@ -240,10 +257,14 @@ where
         self
     }
 
-    /// Cofigures how a sequence is read from RAM and is spread to the compare register.
+    /// Configures how a sequence is read from RAM and is
+    /// spread to the compare register. Disables
+    /// [Channel::C3] for [LoadMode::Waveform], and disables
```
qwandor, line 262: I think you mean "enabled it otherwise", though I'm not sure why.

BartMassey: Correct. I'm not sure why either — seems like just a bug, but wanted to document current behavior before thinking about it.

BartMassey:
Thanks for the review and the catches! I have a couple of other things to document as well at this point: I'll push an update RSN.

PR #541, opened 2026-02-26
Make quadrature decoder input pins generic

BartMassey:
The QDEC quarature decoder input pin struct `qdec::Pins` had the input pins hard-locked to `PUllUp`. Some decoder circuits really want `Floating`. This patch generalizes that.

~~Maybe a breaking change, but I don't think so?~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 
 (no changes)
 
+## [0.20.0]
+
+### Breaking changes
+
+- Made pin modes for `qdec` generic
+
 ## [0.19.0]
 
 ### Breaking changes
@@ -445,3 +451,4 @@ None
 [0.17.1]: https://github.com/nrf-rs/nrf-hal/releases/tag/v0.17.1
 [0.18.0]: https://github.com/nrf-rs/nrf-hal/releases/tag/v0.18.0
 [0.19.0]: https://github.com/nrf-rs/nrf-hal/releases/tag/v0.19.0
+[0.20.0]: https://github.com/nrf-rs/nrf-hal/releases/tag/v0.20.0
--- a/nrf-hal-common/src/qdec.rs
+++ b/nrf-hal-common/src/qdec.rs
@@ -4,18 +4,25 @@
 //! It is suitable for mechanical and optical sensors.
 
 use {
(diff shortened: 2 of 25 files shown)


nrf-hal-common/src/qdec.rs
```diff
@@ -4,18 +4,20 @@
...
 use {
-    crate::gpio::{Input, Pin, PullUp},
+    crate::gpio::{Input, Pin, PushPull},
     crate::pac::QDEC,
+    core::marker::PhantomData,
 };
 
 /// A safe wrapper around the `QDEC` peripheral with associated pins.
-pub struct Qdec {
+pub struct Qdec<I> {
```
qwandor: This is an API breaking change, as anyone storing this type somewhere will now have to add the parameter too.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 (no changes)
 
+## [0.19.1]
+
+- Made input pin modesfor `qdec` generic
```
qwandor: Add a `### Breaking changes` header for this.

qwandor (commented):
The new release should be 0.20.0, as this is a breaking change.

BartMassey:
Yeah, thanks for the catch. Will fix shortly.

BartMassey:
Sigh. Will fix example shortly.

qwandor (approved):

BartMassey:
Went ahead and made the LED input the same pin mode. Doesn't really matter, but it seemed weird to insist on `PullUp` for a pin that would be turned into an output immediately anyway.

BartMassey:
Is there any guarantee that the LED pin comes back with the same pin mode we sent in? I can't find much in the nRF manual about this. I can't even really tell whether the nRF puts the LED pin back to input or just sends it low between LED light-ups.

BartMassey:
I need to look a bit more into the handling of the LED pin before we merge this. The nRF documentation is not too clear on what happens with it when it is supplied to the QDEC.

BartMassey:
Confirmed that QDEC pins at free time are in the same state they were at new time: careful docs read plus experiment. Thus this patch is valid and ready to merge. I will squash it before merging, for cleaner record.

